PR #4, opened 2024-02-12
fix: Update to install process

achatur:
Removed --dry-run because it's deprecated in some versions and not working. 
Added top level dir for the openstack helm command.


cardoe (changes requested):
The point of the `--dry-run` is to create the yaml files that kubeseal consumes. If you don't use the dry-run flag then it creates those secrets live in the cluster

PR #5, opened 2024-02-13
feat: add ingress-nginx and enable argocd and nautobot to use it

cardoe:
Adds ingress-nginx to be deployed into a cluster and enables Nautobot and ArgoCD to use it


ctria (approved):
Looks good to me.

ctria (approved):

PR #6, opened 2024-02-15
feat: add component argo workflow

achatur:
Adds argo-workflow as a component.  This is not adding events or the workflows yet.  It's just setting up the workflow namespace and environment to be used.  New workflow namespace is configured as "argo".

### DIFF
--- a/components/11-argo-workflow/kustomization.yaml
+++ b/components/11-argo-workflow/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - https://github.com/argoproj/argo-workflows/releases/download/v3.5.4/namespace-install.yaml
--- a/apps/components/argo-workflow.yaml
+++ b/apps/components/argo-workflow.yaml
@@ -1,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: argo-workflow
+spec:
+  project: understack
+  source:
+    repoURL: https://github.com/rackerlabs/understack.git
+    path: components/11-argo-workflow/
+    targetRevision: HEAD
+  destination:
+    server: "https://kubernetes.default.svc"
+    namespace: argo
+  syncPolicy:
+    automated:
+      selfHeal: true
--- a/apps/components/argocd-understack-project.yaml
+++ b/apps/components/argocd-understack-project.yaml
@@ -12,6 +12,8 @@ spec:
     server: '*'
   - namespace: 'openstack'
     server: '*'
+  - namespace: 'argo'
+    server: '*'
   clusterResourceWhitelist:
   - group: '*'
     kind: '*'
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -68,3 +68,9 @@ ArgoCD should successfully get Nautobot deployed. Now come the OpenStack
 components which aren't working with GitOps methods at this time.
 
 [Install Keystone](./components/10-keystone/README.md)
+
+### Setup Argo Workflow
+
+```bash
+kubectl kustomize --enable-helm components/11-argo-workflow/ | kubectl apply --server-side -f -
```
cardoe, line 75: So here you say it needs server side apply but then the ArgoCD config isn't using server side apply. Which is it?

cardoe: You also shouldn't need to add this to the top level readme since Argo should deploy this automatically.

components/11-argo-workflow/base/templates/cluster.yaml
```diff
@@ -0,0 +1,22 @@
+---
+
+apiVersion: kind.x-k8s.io/v1alpha4
+kind: Cluster
```
cardoe, line 4: This whole config file assumes that you're running on kind and will fail in any other cluster.

components/11-argo-workflow/base/templates/configmap.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: ConfigMap
+metadata:
+  name: argo-configmap
+  namespace: "argo"
+  labels:
+    app.kubernetes.io/name: argo-workflows
+    helm.sh/chart: argo-workflows-0.40.10
+    app.kubernetes.io/instance: argo
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/version: "0.40.10"
```
cardoe, line 13: The hardcoded versions here will break in the future. But what's the point of the ConfigMap? it's empty?

components/11-argo-workflow/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+  chartHome: ../../charts/
+
+helmCharts:
+  - name: argo-workflows
+    includeCRDs: true
+    namespace: argo
+    valuesFile: values.yaml
+    releaseName: argo-workflow
+    version: 0.40.10
+    repo: https://argoproj.github.io/argo-helm
```
cardoe, line 18: Given what the project says about the charts being community maintained and not officially maintained. The fact that the kustomize native config is the officially maintained config. We also want to install this in the namespaced form and not cluster wide which the charts don't support. I think we switch.

README.md
achatur: I was wondering that. OK I'll remove this

components/11-argo-workflow/base/templates/configmap.yaml
achatur: there is no point.. it was a place holder but I can remove it

components/11-argo-workflow/base/templates/cluster.yaml
achatur: hmm.. so I am thinking overlay w/ dev vs prod.  Would that work?  This can remain a template

cardoe (approved):

PR #8, opened 2024-02-21
fix: Various improvement and fixes found in 1st test deployment

skrobul:
This PR addresses some of the issues I have found when trying to deploy this on clean cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,17 +23,35 @@ You must have the following installed:
 - `kustomize` (5.x versions)
 - `helm` (3.8 or newer)
 - `kubeseal`
+- `cmctl`
+
+Alternatively, if you don't have those dependencies you can use the dedicated
+development environment including those tools by launching `nix-shell` in the
+project directory. If you don't have `nix-shell` on your machine, it can be
+[downloaded here](https://nixos.org/download.html).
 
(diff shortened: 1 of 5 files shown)


skrobul:
One more thing I was not sure about - the bootstrap installs the `ingress-nginx` ingress controller, but the [`09-nautobot/ingress.yaml`](https://github.com/rackerlabs/understack/blob/main/components/09-nautobot/ingress.yaml) includes good old traefik's `IngressRoute`. Is this intended? If it is, it has the hardcoded `Host` frontend rule which probably needs changing.

cardoe (approved):

cardoe (approved):

cardoe:
> One more thing I was not sure about - the bootstrap installs the `ingress-nginx` ingress controller, but the [`09-nautobot/ingress.yaml`](https://github.com/rackerlabs/understack/blob/main/components/09-nautobot/ingress.yaml) includes good old traefik's `IngressRoute`. Is this intended? If it is, it has the hardcoded `Host` frontend rule which probably needs changing.

Nope. That file isn't even included in kustomize so we need to just delete it. The ingress actually comes from within the base directory.

PR #11, opened 2024-02-27
feat: add installation of OpenStack Ironic

cardoe:
Create an initial configuration to install OpenStack Ironic. The PXE service right now binds to the host's network device so this configuration won't work on all systems but it's a first step in getting this deployed.

### DIFF
--- a/components/01-secrets/README.md
+++ b/components/01-secrets/README.md
@@ -121,6 +121,44 @@ for skrt in $(find . -name "secret-keystone*.yaml" -depth 1); do
 done
 ```
 
+## Ironic
+
+Generate the necessary secrets for OpenStack Ironic.
+
+```bash
+kubectl --namespace openstack \
+        create secret generic ironic-rabbitmq-password \
+        --type Opaque \
+        --from-literal=username="ironic" \
+        --from-literal=password="$($(git rev-parse --show-toplevel)/scripts/pwgen.sh)" \
+        --dry-run=client -o yaml > secret-ironic-rabbitmq-password.yaml
+kubectl --namespace openstack \
+        create secret generic ironic-db-password \
+        --type Opaque \
+        --from-literal=password="$($(git rev-parse --show-toplevel)/scripts/pwgen.sh)" \
+        --dry-run=client -o yaml > secret-ironic-db-password.yaml
+kubectl --namespace openstack \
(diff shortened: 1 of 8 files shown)


nicholaskuechler (approved):

components/01-secrets/README.md
```diff
@@ -121,6 +121,44 @@ for skrt in $(find . -name "secret-keystone*.yaml" -depth 1); do
...
+        --type Opaque \
+        --from-literal=username="ironic" \
+        --from-literal=password="$($(git rev-parse --show-toplevel)/scripts/pwgen.sh)" \
+        --dry-run -o yaml > secret-ironic-keystone-password.yaml
+```
+
+Now let's seal them.
+
+```bash
+for skrt in $(find . -name "secret-ironic*.yaml" -depth 1); do
```
skrobul, line 151: as per the other PR - this probably should  be `-maxdepth 1` ? 
```
❯ find . -name "secret-ironic*.yaml" -depth 1
find: warning: you have specified the global option -depth after the argument -name, but global options are not positional, i.e., -depth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
find: paths must precede expression: `1'
```

the `-depth` is just [a boolean flag](https://man.cx/find#heading5)

components/13-ironic/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Deploy Ironic
+
+NOTE: The PXE service currently has the host network devices mapped into
+the container. You'll have to edit the [aio-values.yaml](./aio-values.yaml)
+file in the `network.pxe.device` field to the correct value for your
+server to have this start up successfully.
+
+Since we cannot refer to the secrets by name, we must look them up live from the cluster
+so that we can injected them into the templated configs. Upstream should really allow
```
skrobul, line 34:
```suggestion
so that we can inject them into the templated configs. Upstream should really allow
```

components/01-secrets/README.md
cardoe: You're right. I copied this before you fixed it in the file I copied it from. Good catch.

cardoe: Found where you fixed this before... 57ad0553d48

skrobul (approved):

PR #14, opened 2024-02-29
feat: DexIDP Integration for Nautobot

skrobul:
Adds Dex IDP and lets Nautobot use it to authenticate users against Keystone so the same users are used across the entire stack. Closes https://rackspace.atlassian.net/browse/PUC-167

### DIFF
--- a/components/09-nautobot/base/nautobot/templates/configmap.yaml
+++ b/components/09-nautobot/base/nautobot/templates/configmap.yaml
@@ -111,3 +111,517 @@ data:
     ; For larger installations, certain API calls (example: Relationships, GraphQL) can have a length of query parameters that go over uWSGI default limit.
     ; Setting the buffer size to larger than default (4096) can have an impact on memory utilization, but can be set as high as the header limit of 65535.
     buffer-size = 4096
+  nautobot_config.py: |
+    import os
+    import sys
+
+    from nautobot.core.settings import *  # noqa F401,F403
+    from nautobot.core.settings_funcs import is_truthy, parse_redis_connection
+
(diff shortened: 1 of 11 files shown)


components/09-nautobot/base/nautobot/templates/configmap.yaml
```diff
@@ -111,3 +111,533 @@ data:
...
+    # Ensure proper Unicode handling for MySQL
+    #
+    if DATABASES["default"]["ENGINE"].endswith("mysql"):
+        DATABASES["default"]["OPTIONS"] = {"charset": "utf8mb4"}
+
+    # This key is used for secure generation of random numbers and strings. It must never be exposed outside of this file.
+    # For optimal security, SECRET_KEY should be at least 50 characters in length and contain a mix of letters, numbers, and
+    # symbols. Nautobot will not run without this defined. For more information, see
+    # https://docs.djangoproject.com/en/stable/ref/settings/#std:setting-SECRET_KEY
+    SECRET_KEY = os.getenv("NAUTOBOT_SECRET_KEY", "at)xtt_l9+7uyp-h_@&cc_v_$vw51m*u(9un^oicnxe&_yy8cm")
```
cardoe, line 185: We should probably not have a default and require it be passed in no?

skrobul: fair point - it was copy&paste from the existing one, I'll adjust this

skrobul: fixed in the rebase

PR #16, opened 2024-02-29
feat: Adds tutorial/guide to installing UnderStack on Ubuntu and K3s

nicholaskuechler:
Adds a step by step guide to getting UnderStack up and running on a Ubuntu server with K3s.

### DIFF
--- a/components/13-ironic/aio-values.yaml
+++ b/components/13-ironic/aio-values.yaml
@@ -20,6 +20,17 @@ conf:
       automated_clean: false
     dhcp:
       dhcp_provider: none
+    oslo_messaging_rabbit:
+      rabbit_ha_queues: true
+
+endpoints:
+  oslo_messaging:
+    namespace: null
+    statefulset:
+      replicas: 3
+      name: rabbitmq-server
+    hosts:
+      default: rabbitmq-nodes
 
 network:
   api:
@@ -62,6 +73,13 @@ dependencies:
           service: oslo_db
         - endpoint: internal
           service: oslo_messaging
+    # (nicholas.kuechler) The upstream helm values have ironic-db-init
+    # as a dependency for db_sync, but we have job_db_init false so
+    # ironic-db-init never gets created, and this dependency won't pass.
+    # unset it, so that ironic-db-init is no longer listed in it.
+    # https://opendev.org/openstack/openstack-helm/src/branch/master/ironic/values.yaml#L310-L312
+    db_sync:
+      jobs:
 
 manifests:
   job_db_init: false
--- a/docs/install-understack-ubuntu-k3s.md
+++ b/docs/install-understack-ubuntu-k3s.md
@@ -1,0 +1,413 @@
+# Installing UnderStack on Ubuntu 22.04 + K3s
+
+## Get UnderStack
+
+First, let's git clone the understack repo:
+
+```bash
+git clone https://github.com/rackerlabs/understack.git
+```
+
+## Install Pre-requisites
+
+Install some packages we'll need later and some useful troubleshooting utilities.
+```bash
+apt-get -y install curl jq net-tools telnet git apt-transport-https wget
+```
+
+## Update Ubuntu
+
+Update to the latest ubuntu packages and reboot if necessary.
+```bash
+apt-get -y update
+```
+
+## Install K3s
+
+```bash
+curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--disable=traefik" sh -
+```
+
+References:
+
+* [https://docs.k3s.io/](https://docs.k3s.io/)
+
+## Install Helm
+
+The K3s installer will install kubectl, but we'll also need helm for the UnderStack install.
+
+```bash
+curl https://baltocdn.com/helm/signing.asc | gpg --dearmor | sudo tee /usr/share/keyrings/helm.gpg > /dev/null
+sudo apt-get install apt-transport-https --yes
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/helm.gpg] https://baltocdn.com/helm/stable/debian/ all main" | sudo tee /etc/apt/sources.list.d/helm-stable-debian.list
+sudo apt-get update
+sudo apt-get install helm
+```
+
+References:
+
+* [https://helm.sh/docs/intro/install/](https://helm.sh/docs/intro/install/)
+
+## Install Kustomize
+
+```bash
+curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+sudo mv kustomize /usr/bin
+```
+
+References:
+
+* [https://kubectl.docs.kubernetes.io/installation/kustomize/](https://kubectl.docs.kubernetes.io/installation/kustomize/)
+
+## Install Kubeseal
+
+```bash
+wget https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.26.0/kubeseal-0.26.0-linux-amd64.tar.gz
+tar xzf kubeseal-0.26.0-linux-amd64.tar.gz
+sudo mv kubeseal /usr/bin
+```
+
+References:
+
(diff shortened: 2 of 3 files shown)


docs/install-understack-ubuntu-k3s.md
```diff
@@ -0,0 +1,305 @@
+# Installing UnderStack on Ubuntu 22.04 + K3s
+
+## Install Pre-requisites
+
+Install some packages we'll need later and some useful troubleshooting utilities.
```
skrobul, line 5: are we installing full golang just to have the `go env GOOS` and `go env GOARCH` available?
If so, I think we can safely assume that Ubuntu is `linux` and we can get architecture with `uname -m` although I don't think we will be running ARM servers that soon.

nicholaskuechler: Yeah, I was following their docs which used it and I wasn't sure if I would need it for anything else. I've updated it to remove the go dependency. 👍

docs/install-understack-ubuntu-k3s.md
```diff
@@ -0,0 +1,305 @@
...
+## Install Helm
+
+The K3s installer will install kubectl, but we'll also need helm for the UnderStack install.
+
+```bash
+curl https://baltocdn.com/helm/signing.asc | gpg --dearmor | sudo tee /usr/share/keyrings/helm.gpg > /dev/null
+sudo apt-get install apt-transport-https --yes
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/helm.gpg] https://baltocdn.com/helm/stable/debian/ all main" | sudo tee /etc/apt/sources.list.d/helm-stable-debian.list
+sudo apt-get update
+sudo apt-get install helm
```
cardoe, line 36: Is this the official helm repo?

docs/install-understack-ubuntu-k3s.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+References:
+
+* [https://github.com/bitnami-labs/sealed-secrets?tab=readme-ov-file#installation](https://github.com/bitnami-labs/sealed-secrets?tab=readme-ov-file#installation)
+
+## Install Cert Manager cmctl
+
+```bash
+OS=$(go env GOOS); ARCH=$(go env GOARCH); curl -fsSL -o cmctl.tar.gz https://github.com/cert-manager/cert-manager/releases/latest/download/cmctl-$OS-$ARCH.tar.gz
```
cardoe: `uname` can replace these.

cardoe (commented):
I'm thinking we simplify the install of the binaries with a script.

docs/install-understack-ubuntu-k3s.md
```diff
@@ -0,0 +1,305 @@
+# Installing UnderStack on Ubuntu 22.04 + K3s
+
+## Install Pre-requisites
+
+Install some packages we'll need later and some useful troubleshooting utilities.
+```bash
+apt-get -y install curl jq yq net-tools telnet git apt-transport-https golang-go
```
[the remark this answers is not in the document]
nicholaskuechler: Already updated 👍

docs/install-understack-ubuntu-k3s.md
nicholaskuechler: Yes, I copy/pasted it from: https://helm.sh/docs/intro/install/#from-apt-debianubuntu

docs/install-understack-ubuntu-k3s.md
nicholaskuechler: Already updated 👍

nicholaskuechler:
> I'm thinking we simplify the install of the binaries with a script.

Just the CLI tools like helm, kustomize, kubeseal, etc or a script that installs everything including the k3s and apt package requirements - kind of a "quick-start.sh" that does it all for you?

skrobul:
>     I'm thinking we simplify the install of the binaries with a script.

Just a thought:

**one-time [installation](https://nixos.org/download.html) of nix**
```
sh <(curl -L https://nixos.org/nix/install) --no-daemon
```

**ongoing usage**

```
# on clean system
$ k3s 
bash: k3s: command not found...
$ helm 
bash: k3s: command not found...
$
$ nix-shell
❯ nix-shell

[nix-shell:~/devel/understack]$ k3s
NAME:
   k3s - Kubernetes, but small and simple
```


Apart from having the versioned, stable environment that always works [it is declarative](https://github.com/rackerlabs/understack/blob/main/shell.nix) and does not pollute main system. 
Produces same results on MacOS, Linux and possibly Windows ;)

cardoe:
> > ```
> > I'm thinking we simplify the install of the binaries with a script.
> > ```
> 
> Just a thought:
> 
> **one-time [installation](https://nixos.org/download.html) of nix**
> 
> ... snip ...
> 
> Apart from having the versioned, stable environment that always works [it is declarative](https://github.com/rackerlabs/understack/blob/main/shell.nix) and does not pollute main system. Produces same results on MacOS, Linux and possibly Windows ;)

Yeah I'm with you on that. Realistically virtually all of these things shouldn't have to install these on the actual host as well.

components/13-ironic/aio-values.yaml
```diff
@@ -47,12 +47,42 @@ conf:
...
+endpoints:
+  oslo_messaging:
+    namespace: null
+    auth:
+      admin:
+        username: rabbitmq
+        password: password
+        secret:
+          tls:
+            internal: rabbitmq-tls-direct
```
cardoe, line 67: This chunk shouldn't be necessary. We don't use the admin cred.

components/13-ironic/aio-values.yaml
```diff
@@ -47,12 +47,42 @@ conf:
...
+      ironic:
+        username: ironic
+        password: password
+    statefulset:
+      replicas: 3
+      name: rabbitmq-server
+    hosts:
+      default: rabbitmq-nodes
+    host_fqdn_override:
+      default: null
```
cardoe, line 77: This isn't necessary cause this is the default when you set `statefulset` and `hosts` like you did above.

components/13-ironic/aio-values.yaml
```diff
@@ -47,12 +47,42 @@ conf:
...
+    path: /ironic
+    scheme: rabbit
+    port:
+      amqp:
+        default: 5672
+      http:
+        default: 15672
```
cardoe, line 84: This chunk is the default as well.

components/13-ironic/aio-values.yaml
```diff
@@ -47,12 +47,42 @@ conf:
...
+      rabbit_ha_queues: true
   logging:
     logger_root:
       level: DEBUG
     logger_ironic:
       level: DEBUG
 
+endpoints:
+  oslo_messaging:
+    namespace: null
```
cardoe, line 60: This came from the genestack config file from keystone right? Cause that's actually wrong and been planning on deleting it.

nicholaskuechler: This came from understack keystone aio-values: https://github.com/rackerlabs/understack/blob/main/components/10-keystone/aio-values.yaml#L983-L1008

nicholaskuechler: I tried changing it all to just:
```
endpoints:
  oslo_messaging:
    hosts:
      default: rabbitmq-nodes
```
But it doesn't work...

nicholaskuechler: This one works:
```
endpoints:
  oslo_messaging:
    namespace: null
    statefulset:
      name: rabbitmq-server
    hosts:
      default: rabbitmq-nodes
```

components/13-ironic/aio-values.yaml
```diff
@@ -20,6 +20,21 @@ conf:
       automated_clean: false
     dhcp:
       dhcp_provider: none
+    oslo_messaging_rabbit:
+      rabbit_ha_queues: true
+  logging:
+    logger_root:
+      level: DEBUG
+    logger_ironic:
+      level: DEBUG
```
cardoe, line 29: So I think the logging chunk can be dropped.

cardoe (approved):

PR #17, opened 2024-03-01
feat: use a common container list for OSH

cardoe:
Rather than having to maintain and keep up with different override files for OpenStack Helm's containers, centralize it into one file that can be shared between all the components.


cardoe:
Ultimately this is a no-op and just moving the containers into 1 shared file

skrobul (approved):

PR #18, opened 2024-03-05
docs: craft a gitops focused installation

cardoe:
Craft a GitOps focused installation process which users can use to easily stand up a dev environment.


andrrax (approved):

docs/gitops-install.md
```diff
@@ -0,0 +1,144 @@
...
+
+ArgoCD will need to know where it can access your deployment config
+repo. This can be over SSH with a key or over HTTPS. Add to your
+`my-k3s.env` the following:
+
+```bash title="/path/to/understack/my-k3s.env"
+UC_DEPLOY_GIT_URL="git@github.com:my-org/uc-deploy.git"
+UC_DEPLOY_SSH_FILE="/path/to/ssh.private"
+```
+
```
skrobul, line 66: I found it easier to `source` that env file after it is completely written so that I can later on refer to `$UC_DEPLOY` instead of `/path/to/uc/deploy` - it makes sense if the instructions are updated, so the end user can just paste.

docs/gitops-install.md
```diff
@@ -0,0 +1,144 @@
...
+```bash
+kubectl kustomize --enable-helm \
+    https://github.com/rackerlabs/understack//bootstrap/argocd/ \
+    | kubectl apply -f -
+```
+
+Now configure your ArgoCD to have the credential access to your deploy repo:
+
+```bash
+kubectl -n argocd apply -f /path/to/uc-deploy/secrets/my-k3s/argocd/secret-deploy-repo.yaml
```
skrobul, line 131: The indentation in that file is broken:

```
❯ k -n argocd apply -f /home/skrobul/devel/test-uc-deploy/secrets/marek-testrun/argocd/secret-deploy-repo.yaml
error: error parsing /home/skrobul/devel/test-uc-deploy/secrets/marek-testrun/argocd/secret-deploy-repo.yaml: error converting YAML to JSON: yaml: line 10: could not find expected ':'
```

and the file:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: marek-testrun-repo
  labels:
    argocd.argoproj.io/secret-type: repo-creds
data:
  sshPrivateKey: LS0tLS1CRREDACTEDREDACTEDREDACTEDREDACTEDREDACTEDREDACTEDREDACTEDREDACTEDXXX
REDACTEDREDACTEDREDACTEDREALLYREALLYREDACTED313370x0deadb33fMzI0NzkyMgoKMi9y
REDACTEDREDACTEDREDACTEDREALLYREALLYREDACTED313370x0deadb33fMzI0NzkyMgoKMi9y
[...]
  type: Z2l0
  url: 2l0QGdpdREDACTEDREDACTED2J1bC90ZXN0LXVjLWRlcGxveS5naXQ=
```

while the `sshPrivateKey` should be just a single base64 encoded, very long line

docs/gitops-install.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+Finally run the following to have ArgoCD deploy the system:
+
+```bash
+kubectl apply -f /path/to/uc-deploy/clusters/my-k3s/app-of-apps.yaml
```
skrobul, line 137: ```Warning: metadata.finalizers: "resources-finalizer.argocd.argoproj.io": prefer a domain-qualified finalizer name to avoid accidental conflicts with other finalizer writers``` is that OK?

docs/gitops-install.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+At this point ArgoCD will work to deploy Understack.
```
skrobul, line 140: it didn't for me, after looking into the reasons, it turns out my SSH is passphrase protected

```
❯ k -n argocd describe applications app-of-apps
Name:         app-of-apps
Namespace:    argocd
[...]
    Path:             clusters/marek-testrun
    Repo URL:         git@github.com:skrobul/test-uc-deploy.git
[...]
  Conditions:
    Last Transition Time:  2024-04-03T17:15:02Z
    Message:               Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = ssh: this private key is passphrase protected
    Type:                  ComparisonError
  [...]
```

We should probably create a copy of the given key, then run `ssh-keygen -p -N "" -f /tmp/copied_ssh_key`, then remove it immediately or even better:

 `openssl rsa -in  /tmp/work_rsa | base64 | tr -d '\n'`

docs/gitops-install.md
```diff
@@ -0,0 +1,144 @@
...
+
+In this section we will use the [App of Apps][app-of-apps] pattern to define
+the deployment of all the components of UnderStack.
+
+```bash
+./scripts/gitops-deploy.sh ./my-k3s.env
+cd /path/to/uc-deploy
+git add clusters/my-k3s
+git commit -m "my-k3s: initial cluster config"
+```
```
skrobul, line 107: `git push ...` so that ArgoCD can access this data?

scripts/gitops-secrets-gen.sh
```diff
@@ -0,0 +1,96 @@
...
+mkdir -p "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/argocd"
+cat << EOF > "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/argocd/secret-deploy-repo.yaml"
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ${DEPLOY_NAME}-repo
+  labels:
+    argocd.argoproj.io/secret-type: repo-creds
+data:
+  sshPrivateKey: $(cat "${UC_DEPLOY_SSH_FILE}" | base64)
```
skrobul, line 75:
```suggestion
  sshPrivateKey: $(openssl rsa -in  ${UC_DEPLOY_SSH_FILE} | base64 | tr -d '\n')
```

skrobul (commented):
Aside from comments above, there  is a bunch of things that did not fully converge on my installation:

```
❯ k -n argocd get applications.argoproj.io                   
NAME                          SYNC STATUS   HEALTH STATUS
app-of-apps                   Synced        Healthy
argo-events                   Unknown       Healthy
argo-workflows                Synced        Degraded
cert-manager                  Synced        Healthy
dexidp                        Synced        Degraded
ironic                        OutOfSync     Degraded
keystone                      OutOfSync     Degraded
mariadb                       Synced        Healthy
mariadb-operator              Synced        Healthy
memcached                     Synced        Healthy
messaging-topology-operator   Synced        Healthy
nautobot                      Synced        Progressing
postgres-db                   Synced        Healthy
postgres-operator             Synced        Healthy
rabbitmq-cluster              Synced        Healthy
rabbitmq-operator             Synced        Healthy
```

- The ironic/keystone is most likely because of missing node labelling.
- dex is because of missing azure secret
- for `argo-events` the `sso` file is missing in `/components/12-argo-events/sso`, so kustomize fails:

```
 Message:               Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = Manifest generation error (cached): `kustomize build <path to cached source>/components/12-argo-events --enable-helm` failed exit status 1: Error: loading KV pairs: file sources: [sso]: evalsymlink failure on '<path to cached source>/components/12-argo-events/sso' : lstat <path to cached source>/components/12-argo-events/sso: no such file or directory
```

docs/gitops-install.md
```diff
@@ -0,0 +1,144 @@
...
+```bash title="/path/to/understack/my-k3s.env"
+UC_REPO="/path/to/understack"
+UC_DEPLOY="/path/to/uc-deploy"
+DEPLOY_NAME="my-k3s"
+```
+
+### Remaining pre-deployment config
+
+ArgoCD will need to know where it can access your deployment config
+repo. This can be over SSH with a key or over HTTPS. Add to your
```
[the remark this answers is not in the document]
cardoe: No doubt there's a lot of possible options here. There's also a GitHub App option for authentication. As far as the private key, I would use a GitHub Deploy Key. The real question is, long term how do we want to treat this? We're templating out a rough initial set of config that the user updates or are we having a full fledged templating system with a config file?

My goal here was to help capture a happy path deployment.

docs/gitops-install.md
```diff
@@ -0,0 +1,144 @@
...
+
+At this point we will use our configs to make the actual deployment.
+Make sure everything you've committed to your deployment repo is pushed
+to your git server so that ArgoCD can access it.
+
+If you do not have ArgoCD deployed then you can use the following:
+
+```bash
+kubectl kustomize --enable-helm \
+    https://github.com/rackerlabs/understack//bootstrap/argocd/ \
```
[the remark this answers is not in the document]
cardoe: The double slash is necessary here. It's saying git clone the repo at https://github.com/rackerlabs/understack and then use the directory bootstrap/argocd within that clone as the helm chart source. `//` would not have been the syntax I would have chosen to denote the difference.

docs/gitops-install.md
cardoe: yeah. I think upstream ArgoCD will need to change their finalizer at some point to avoid that warning. But that's the value they use and not something I'm making up.

docs/gitops-install.md
cardoe: Ah. I used a GitHub Deploy Key which defaults to being password-less. But I can call that out in the docs.

scripts/gitops-secrets-gen.sh
cardoe: So `openssl rsa -in` doesn't work on my SSH key.

docs/gitops-install.md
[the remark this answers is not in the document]
skrobul:
```
❯ kubectl kustomize --enable-helm  https://github.com/rackerlabs/understack//bootstrap/argocd/   | md5sum            
8004747d220e8ae705d3c36e765c0851  -

❯ kubectl kustomize --enable-helm https://github.com/rackerlabs/understack/bootstrap/argocd/     | md5sum
8004747d220e8ae705d3c36e765c0851  -
```

image

scripts/gitops-secrets-gen.sh
skrobul: Interesting - can you post header of your SSH key? is it text or binary?

skrobul:
> I have no idea what the argo-events sso secret is and that was one of the questions I had on the argo-events PR that was merged in. At least my system was broken in the same way on the main branch.

`sso` is a [configmap generated here](https://github.com/rackerlabs/understack/blob/c45db60c99a819299200508d38507572034a2c06/components/11-argo-workflows/kustomization.yaml#L26), sourced from [components/11-argo-workflows/sso](https://github.com/rackerlabs/understack/blob/c45db60c99a819299200508d38507572034a2c06/components/11-argo-workflows/sso#L1). This in turn uses `argo-sso` secret, which is supposed to be generated by [scripts/easy-secrets-gen.sh here](https://github.com/rackerlabs/understack/blob/c45db60c99a819299200508d38507572034a2c06/scripts/easy-secrets-gen.sh#L44C1-L56C1) but it is not for this deployment.

skrobul:
> I think dex needs to live in this repo without a hard dependency on the azure secret and instead allow us to define a temp user.

Agreed. Logged as PUC-224

docs/gitops-install.md
skrobul: Will argocd add the finalizer if we don't specify it explicitly?

docs/gitops-install.md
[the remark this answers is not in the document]
skrobul: > As far as the private key, I would use a GitHub Deploy 

May be worth linking to documentation on how to [generate a deploy key on Github](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#set-up-deploy-keys).

> There's also a GitHub App option for authentication. 

The main problem I see with deploy keys is that they allow access only to a single repository, which is fine today but long term we may need to access multiple repositories in which case the GitHub App would be more flexible but also harder to setup. So let's stick with deploy keys for now.

> We're templating out a rough initial set of config that the user updates or are we having a full fledged templating system with a config file?

When the config becomes too large, we can think about a fully automated templating system. Right now the whole `my.env` config file is 7 lines long. I think anyone can manage to edit that. My only minor nitpick here is that I would like to have a starter example that has a whole thing listed in the docs, like this:

```shell
UC_REPO="$HOME/devel/understack"
UC_DEPLOY="$HOME/devel/test-uc-deploy/"
DEPLOY_NAME="my-testrun"
UC_DEPLOY_GIT_URL="git@github.com:myuser/test-uc-deploy.git"
UC_DEPLOY_SSH_FILE="/home/myuser/.ssh/test-uc_github_deployment_key_rsa"
UC_DEPLOY_EMAIL="you@yourdomain.com"
DNS_ZONE="testrun.home.lab"
```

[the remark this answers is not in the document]
cardoe: Done. Lemme know if you think this is better.

docs/gitops-install.md
cardoe: I believe I've fixed this.

docs/gitops-install.md
cardoe: No. It's only needed if you need to cascade delete or delete in an order, which we do.

scripts/gitops-secrets-gen.sh
cardoe: `-----BEGIN OPENSSH PRIVATE KEY-----`

cardoe:
I think I've addressed all your feedback.

scripts/gitops-secrets-gen.sh
skrobul: oh okay, mine starts with:
```
-----BEGIN RSA PRIVATE KEY-----
Proc-Type: 4,ENCRYPTED
DEK-Info: AES-128-CBC,5BD
``` 
which is PKCS#1, yours seems to be the OpenSSH's standard. There are 4-5 formats out there, so looks like we probably should `ssh-keygen` for the conversion and decryption instead.

The command to remove the passphrase would be:

```
ssh-keygen -p -N "" -f /tmp/testkey
```

Unfortunately **it modifies the keyfile** in-place, so we need to operate on a copy that we have to remove from the disk after it's used.

skrobul (approved):
> I think I've addressed all your feedback.

Just did another test run on the clean cluster and everything except the SSH key decryption part worked.

scripts/easy-secrets-gen.sh
```diff
@@ -37,7 +39,7 @@ for ns in nautobot dex; do
     -o yaml \
     --type Opaque \
     --from-literal=client-secret="$NAUTOBOT_SSO_SECRET" \
-    > secret-nautobot-sso-$ns.yaml
+    > "${DEST_DIR}/secret-nautobot-sso-$ns.yaml"
```
skrobul, line 42: this is really for the section below but since it wasn't changed, I cannot make a comment on it directly:

```
    > secret-argo-sso-$ns.yaml
```

must be changed to:

```
    > ${DEST_DIR}/secret-argo-sso-$ns.yaml
```

skrobul (changes requested):
just realised I forgot to include about missing `argo-sso` secret, which causes the `secret-argo-sso-*` files being generated in the wrong folder and as a result not included in the deploy repo.


also, this does not deploy the ingress anymore - which is a requirement for the argo-workflows and Nautobot to work. Do you plan to take care of this in separate PR?

scripts/easy-secrets-gen.sh
cardoe: Yeah I think that was pulled in with PR 28. Fixed now.

skrobul (approved):
Approving with an understanding that:
- ingress controller is not being deployed with this. It results in Dex not being reachable for authentication, so the `argo-workflows` and `nautobot` cannot be used
- this will be tracked as separate story

cardoe:
> Approving with an understanding that:
> 
> * ingress controller is not being deployed with this. It results in Dex not being reachable for authentication, so the `argo-workflows` and `nautobot` cannot be used
> * this will be tracked as separate story

Absolutely agree.

PR #20, opened 2024-03-07
feat: argo-workflows and argo-events deployments (JIRA:PUC-193)

andrrax:
This creates basic argo-workflow and argo-events deployments

- deploys argo-workflows in namespaced mode
- argo-workflows controller monitors argo-events namespace
- does not configure auth for argo-workflows
- creates a single EventSource for a webhook event on POSTs to /nautobot
- creates a single Sensor with dependency on the event described above which defines a single hello-world Workflow trigger

### DIFF
--- a/components/11-argo-workflows/patch-server-deployment.yaml
+++ b/components/11-argo-workflows/patch-server-deployment.yaml
@@ -1,0 +1,10 @@
+## configure argo-server namespaced and to set to server mode to disable auth until auth can be worked out
+- op: replace
+  path: /spec/template/spec/containers/0/args
+  value:
+    - server
+    - --auth-mode=server
+    # - --auth-mode=client
+    - --namespaced
+    - --managed-namespace
+    - argo-events
--- a/components/11-argo-workflows/argo-server-cluster-role.yaml
+++ b/components/11-argo-workflows/argo-server-cluster-role.yaml
@@ -1,0 +1,26 @@
+## configure clusterrole for argo-server to provide permissions when argo-worfklows is running namespaced
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: argo-server
+rules:
+- apiGroups:
+  - argoproj.io
+  resources:
+  - clusterworkflowtemplates
+  verbs:
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: argo-server
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: argo-server
+subjects:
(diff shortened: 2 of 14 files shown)


components/12-argo-events/webhook-ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+    nginx.ingress.kubernetes.io/backend-protocol: HTTPS
+    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
+  name: nautobot-webhook
+spec:
+  ingressClassName: nginx
+  rules:
+  - host: events.local
+    http:
+      paths:
+      - path: /nautobot        # must match endpoint defined in EventSource
```
cardoe, line 16: I don't think you need the ingress here. Nautobot is running within the cluster and should just use the service that the event source created.

andrrax: Agreed. I have found the need to receive remote webhooks in development, though we can abstract ingress from this repo.

cardoe:
So just playing with workflows some I think these RBACs are big and maybe not correct?

Look at this role from the docs https://github.com/argoproj/argo-workflows/blob/main/manifests/quick-start/base/webhooks/submit-workflow-template-role.yaml which should be enough for an event to trigger a workflow. They use it in the example docs.

As far as auth goes, if we're wanting to expose the workflow view we should follow https://argo-workflows.readthedocs.io/en/latest/argo-server-sso/ and create accounts in keystone which dex will serve up.

andrrax:
> Look at this role from the docs https://github.com/argoproj/argo-workflows/blob/main/manifests/quick-start/base/webhooks/submit-workflow-template-role.yaml which should be enough for an event to trigger a workflow. They use it in the example docs.

I agree, I'll revisit the RBAC roles. That said, *some* of the RBAC role permissions were set to provide the argo-workflows UI access to view/create Workflows and WorkflowTemplates. Additionally, the because argo-workflows is configured in a namespaced scope, ClusterRoles were used to provide the same permissions to service accounts in two different namespaces (argo and argo-events). I think we'll likely want to make a decision on the workflow UI and if we want to run argo-workflows in namespaced or cluster scope so we can be precise with the RBAC perms.

components/11-argo-workflows/patch-server-deployment.yaml
```diff
@@ -0,0 +1,10 @@
+## configure argo-server namespaced and to set to server mode to disable auth until auth can be worked out
```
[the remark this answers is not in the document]
andrrax: I have a task created to follow up with auth in a separate PR.

andrrax:
I have review the RBAC configuration, and updated the RBAC roles with additional context. Ultimately I feel this configuration is sound and necessary when running both argo-workflows/events in a managed namespace setup.

PR #22, opened 2024-03-14
feat: use secrets for Nautobot SSO configuration

skrobul:
Closes PUC-209

PR #23, opened 2024-03-20
feat: Azure authentication with Dex

skrobul:
Built on top of #22 

This commit brings the possibility of authenticating through Azure SSO which should help with authenticating users.

Unfortunately, the groups functionality does not work in the current format. It boils down to the fact that `go-oidc` does not work with AADv1 for groups (as per the https://dexidp.io/docs/connectors/oidc/). We are not able to switch to AADv2 endpoint (`https://login.microsoftonline.com...``/) because the Issuer field has to match exactly and the app has been created with Azure Portal (AADv1).

I have also tried several workarounds, which all failed but I thought it may be worth recording here:

- using `providerDiscoveryOverrides` to fake out all discovered URLs except the issuer. This is not possible, because OIDC connector in dex allows only to override the token URL and authorization URL.

- setting `insecureEnableGroups:true` was the closest thing to having groups working. It makes the list of group UUIDs being put into the identity token. This list makes the token so large that it goes over the limit of the HTTP Request header value size for nginx, which is a problem when the Client attempts to do a call to https://dexidp.local/userinfo and would likely be a problem on subsequent requests too. The problem is partially because these are UUIDs (36 characters each) rather than the group names.

- use claimMapping of `roles` to `groups` similar to what we used in Nautobot, but that did not work since the AADv1 endpoint does not return them and AADv2 cannot be used due to restriction on Issuer.

One potential alternative solution that I have not tried is using Microsoft connector (https://dexidp.io/docs/connectors/microsoft/), however it also has a set of limitations that make it possibly unacceptable:
- requires Organization Admin access approval to provide access for the app to read full Directory
- may suffer from the same AADv1 vs AADv2 problem
- as per docs: "Users that reject dex’s access through Microsoft will also revoke all dex clients which authenticated them through Microsoft."


components/09-nautobot/base/nautobot/templates/configmap.yaml
```diff
@@ -581,7 +581,7 @@ data:
     SOCIAL_AUTH_OIDC_OIDC_ENDPOINT = "http://dexidp.local"
     SOCIAL_AUTH_OIDC_USERNAME_KEY = 'name'
     SOCIAL_AUTH_OIDC_KEY = 'nautobot'
-    SOCIAL_AUTH_OIDC_SECRET = 'verysecret'
+    SOCIAL_AUTH_OIDC_SECRET = open("/opt/nautobot/dex_client_secret").read()
```
cardoe, line 584: Technically this leaks a file descriptor

components/13-dexidp/values.yaml
```diff
@@ -30,17 +30,57 @@ config:
         domain: default
         keystoneUsername: demo
         keystonePassword: DEMO_PASS
+    - type: oidc
+      name: azure
+      id: azure
+      config:
+        issuer: https://login.microsoftonline.com/570057f4-73ef-41c8-bcbb-08db2fc15c2b/v2.0
```
cardoe, line 37: Can we make the tenant UUID a variable like `$AZURE_SSO_TENANT_ID` like the other bits?

cardoe (commented):
Nice work on this! Two small nits.

components/13-dexidp/values.yaml
skrobul: sure, done in [fbab834](https://github.com/rackerlabs/understack/pull/23/commits/fbab834515c629402476b954c2cff31131855f89)

components/09-nautobot/base/nautobot/templates/configmap.yaml
skrobul: cpython would normally close it during garbage collection, but that does not really make it ok - fixed in the commit below. thanks!

cardoe (approved):

PR #25, opened 2024-03-21
feat: generate memcache token secret to be constant

cardoe:
OpenStack Helm will always randomly generate this value if we don't use a constant memcache token secret in our config. This causes the secrets to always change and their annotations to change which restarts all containers from OpenStack Helm every time. This should allow us to maintain the container configs as idempotent.


ctria (approved):
LGTM

PR #27, opened 2024-04-02
feat: Workflow to synchronize Nautobot Device to Ironic Node

skrobul:
This PR brings an Argo Workflow and sensor that reacts to an 'Interface Updated' event produced by Nautobot's Event Source. Once activated, it updates (or creates) the Node representation in Ironic.


.github/workflows/build-container-images.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to ghcr.io
+        if: ${{ github.event_name != 'pull_request' }}
+        uses: docker/login-action@v3
+        with:
+          registry: "ghcr.io"
+          username: "${{ secrets.GHCR_USER }}"
+          password: "${{ secrets.GHCR_PAT }}"
```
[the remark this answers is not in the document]
skrobul: All done, but we won't be able to test this until merged as the [`GITHUB_TOKEN` does not get `write` permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) on pull requests.

argo-workflows/ironic-nautobot-sync/code/synchronize-server.py
```diff
@@ -0,0 +1,99 @@
...
+    redfish_password="REDACTED",
+    redfish_auth_type="auto",
+    redfish_verify_ca=False
+)
+
+if ironic_node.driver_info != asdict(expected_driver_info):
+    logger.info(f"{node_id} Updating driver_info")
+    patch = [{"op": "replace", "path": "/driver_info", "value": asdict(expected_driver_info)}]
+    response = client.update_node(node_id, patch)
+    logger.info(f"Updated: {response}")
```
mfencik, line 99: It would probably be good to adjust this, as anytime there will be a drac IP change in Nautobot, it will trigger this update which overrides all driver_info settings in Ironic. Main concern is the password as that will be set in a different workflow.

mfencik (changes requested):
Just one small change needed, otherwise looking good

argo-workflows/ironic-nautobot-sync/code/synchronize-server.py
skrobul: Good catch, thank you ! Addressed in 20115d8

cardoe (approved):

PR #32, opened 2024-04-08
feat: make Azure auth backend optional

skrobul:
This PR helps with standing up development/staging environments that do not require SSO authentication to Azure by deploying a generic Dex configuration with just an internal Keystone connector.

The Azure backend can be reenabled by changing the [value file](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#values-files) after the deployment.

Closes PUC-224


apps/components/dexidp.yaml
```diff
@@ -5,9 +5,19 @@ metadata:
...
+      targetRevision: 0.16.0
+      helm:
+        releaseName: dex
+        valueFiles:
+          - $values/components/13-dexidp/values-generic.yaml
+          # enable this for Azure integration (you need to create secret yourself)
+          #  - values-azure.yaml
+    - repoURL: https://github.com/rackerlabs/understack
+      path: components/13-dexidp
+      targetRevision: dex-optional-azure
```
cardoe, line 19: This branch goes away once this merges.

cardoe (changes requested):

apps/components/dexidp.yaml
skrobul: done

cardoe (approved):

PR #37, opened 2024-04-09
fix: simplify Keystone upgrade process

cardoe:
Simplify the Keystone installation process by adjusting the annotations on jobs to replace jobs as necessary so they do not conflict when changes are made. Removed the numeric prefix from the keystone component directory. Added a helm-configs override to the keystone component so that the user has some place to make local modifications.


cardoe:
Moved back to draft because I forgot I need https://review.opendev.org/c/openstack/openstack-helm/+/915336 to land first.

cardoe:
Upstream fixed landed, updated the version dependency to include that fix.

nicholaskuechler (approved):

scripts/gitops-deploy.sh
```diff
@@ -52,5 +52,6 @@ for part in operators components; do
     done
     rm -rf "${OUTPUT_DIR}/${part}/kustomization.yaml"
 done
+touch "${UC_DEPLOY}/helm-configs/${DEPLOY_NAME}/keystone.yaml"
```
skrobul, line 55: I think it may be slightly unclear what type of customisation is available - should we add some comment in that file instead? 
for example

```bash
echo "# Add your custom openstack-helm values here" > "${UC_DEPLOY}/helm-configs/${DEPLOY_NAME}/keystone.yaml"

File: components/keystone/aio-values.yaml [hunk 2]
```diff
@@ -1085,3 +1076,28 @@ manifests:
   secret_registry: true
   service_ingress_api: true
   service_api: true
+
+# we don't want to enable OpenStack Helm's
```
skrobul, line 1080: good description, consider rewrapping to more than 42 characters ;-)

skrobul (approved):
lgtm

components/keystone/aio-values.yaml
cardoe: lol yeah I dunno why I did that. This is hopefully temporary. I'm discussing with upstream how to resolve this correctly.

scripts/gitops-deploy.sh
cardoe: Sure I can do that.

cardoe: I expanded on that a bit. Lemme know if that change makes sense.

skrobul (approved):

PR #39, opened 2024-04-11
feat: package up dnsmasq as a DNS and DHCP server

skrobul:
This PR adds dnsmasq installation to the `openstack` namespace. It will:
- hand out IP addresses to hosts that are provisioned with Ironic
- cooperate with Ironic for setting per-host options
- enable a DNS server that can resolve records for endpoint URLs returned by the OpenStack Identity service catalogue

In its current form, it supports only a single DHCP range/pool. Support for multiple pools will be added as a separate story as it requires more input parameters (MACs of the relays, range names, and potentially different DNS server IPs).

Closes PUC-225

### DIFF
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -1,0 +1,36 @@
+name: container builds
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "containers/**"
+      - ".github/workflows/containers.yml"
+  pull_request:
+    paths:
+      - "containers/**"
+      - ".github/workflows/containers.yml"
+  workflow_dispatch:
+
+jobs:
+  ghcr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: setup docker buildx
+        uses: docker/setup-buildx-action@v3
+      - name: login to ghcr.io
+        if: ${{ github.event_name != 'pull_request' }}
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: build and deploy container to registry
+        uses: docker/build-push-action@v5
+        with:
+          context: "{{defaultContext}}:containers"
+          file: containers/Dockerfile.ironic
+          push: ${{ github.event_name != 'pull_request' }}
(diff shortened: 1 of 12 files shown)


cardoe:
Probably some commits can get squashed into others but overall I think it's good. Need to stop systemd-resolvd or whatever it's called if you use container hostPort 53 however. We can have it listen on the ingress if you configure the ingress to listen on UDP 53. We can also have it listen on a service IP.

nicholaskuechler (approved):

containers/Dockerfile.dnsmasq
```diff
@@ -0,0 +1,17 @@
...
+
+COPY common/helpers.sh /helpers.sh
+COPY dnsmasq/entry-point.sh /entry-point.sh
+RUN chmod +x /entry-point.sh
+COPY dnsmasq/dnsmasq.conf.j2 /etc/dnsmasq.conf.j2
+# let our entry point write out the script
+RUN ln -sf /etc/dnsmasq.d/dnsmasq.conf /etc/dnsmasq.conf
+
+EXPOSE 53/udp
+EXPOSE 67/udp
```
cardoe, line 14: Probably should drop the port 53 since this PR makes it listen on 1053.

cardoe: And we listen on port 1067 for DHCP

.github/workflows/containers.yaml
```diff
@@ -0,0 +1,71 @@
...
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/rackerlabs/openstackhelm/ironic
+          tags: |
+            type=raw,value=latest,enable={{is_default_branch}}
+            type=sha,enable={{is_default_branch}}
+            type=raw,value=2023.1-ubuntu_jammy
+          labels: |
+            org.opencontainers.image.title=ironic with ISO utils
+            org.opencontainers.image.base.name=ghcr.io/rackerlabs/openstackhelm/ironic
```
cardoe, line 42: So that's not really correct. This should be the container we're calling `FROM` in the `Dockerfile`, no?

skrobul: Good catch, thanks ! fixed in the rebase.

containers/Dockerfile.dnsmasq
skrobul: fixed in the rebase

PR #41, opened 2024-04-11
fix: makes nautobot oidc auth endpoint configurable

nicholaskuechler:
I need to be able to set django's `SOCIAL_AUTH_OIDC_OIDC_ENDPOINT` to a different url, but let's keep the existing default in place.


components/09-nautobot/base/nautobot/templates/configmap.yaml
```diff
@@ -580,7 +580,7 @@ data:
     SOCIAL_AUTH_OIDC_SSL_PROTOCOL = None
```
cardoe, line 580:
```suggestion
    if os.env("SOCIAL_AUTH_OIDC_ENDPOINT", "http://dexidp.local").startswith("http://"):
        SOCIAL_AUTH_OIDC_SSL_PROTOCOL = None
```

cardoe (commented):
Do we need to include that?

cardoe (approved):

PR #46, opened 2024-04-12
feat: adds creation of demo user and undercloud-dev nautobot access group.

nicholaskuechler:
The group `undercloud-dev` is used by nautobot and users in this group have elevated permissions. With dex integration, we can use the keystone group undercloud-dev to control access to nautobot.

### DIFF
--- a/components/keystone/aio-values.yaml
+++ b/components/keystone/aio-values.yaml
@@ -21,6 +21,12 @@ bootstrap:
           --user="${OS_USERNAME}" \
           --domain="${OS_DEFAULT_DOMAIN}" \
           "admin"
+    # create a demo user with demo password
+    openstack user create --domain="${OS_DEFAULT_DOMAIN}" --password demo demo
+    # create undercloud-dev group
+    openstack group create --or-show undercloud-dev
+    # add demo user to undercloud-dev group
+    openstack group add user undercloud-dev demo
 
 network:
   api:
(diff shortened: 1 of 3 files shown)


components/keystone/aio-values.yaml
```diff
@@ -21,6 +21,12 @@ bootstrap:
           --user="${OS_USERNAME}" \
           --domain="${OS_DEFAULT_DOMAIN}" \
           "admin"
+    # create a demo user with demo password
+    openstack user create --domain="${OS_DEFAULT_DOMAIN}" --password demo demo
+    # create undercloud-dev group
+    openstack group create --or-show undercloud-dev
+    # add demo user to undercloud-dev group
+    openstack group add user undercloud-dev demo
```
cardoe, line 29: Can the `group add user` be run multiple times? Cause the bootstrap job does get restarted periodically.

nicholaskuechler: Yes, I tested it out first:
```
root@openstack-admin-client:/# openstack group add user undercloud-dev keekz
root@openstack-admin-client:/# openstack group add user undercloud-dev keekz
root@openstack-admin-client:/# echo $?
0
```
Some subcommands have `--or-show` flag, but this particular one doesn't.

skrobul:
```bash
openstack group contains user undercloud-dev demo 2>&1| grep -q 'not in group' && openstack group add user undercloud-dev demo
```

should do the trick but guess the end result is the same so we can just leave it as-is

skrobul (approved):

PR #51, opened 2024-04-15
feat: switch back to nautobot helm chart

cardoe:
Switch back to using the nautobot helm chart now that the fixes are merged upstream. This will allow us to remain current with nautobot releases going forward.


cardoe:
So leaving this as draft because the NAUTOBOT_SECRET_KEY doesn't seem to be respected and I believe ArgoCD will always load the tests.

cardoe:
I think it's not working because of https://github.com/argoproj/argo-cd/issues/13220

cardoe:
Alright came up with a work around by using the URL to this repo. I'll try and see if I can't submit a fix upstream for ArgoCD.

cardoe:
This change will drag us forward to Nautobot 2.1.9 as well since that's what the helm charts package.

nicholaskuechler (approved):

PR #64, opened 2024-04-23
fix: dnsmasq: pin versions, switch to debian

skrobul:
Turns out the Alpine project does not follow semver and changes the versioned upstream images after the fact, so we have to pin them to a specific sha256 to keep stable builds.

Closes PUC-241. Fixes [build failures like this one](https://github.com/rackerlabs/understack/actions/runs/8789533012/job/24119504196)

### DIFF
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -57,10 +57,9 @@ jobs:
           tags: |
             type=raw,value=latest,enable={{is_default_branch}}
             type=sha,enable={{is_default_branch}}
-            type=raw,value=2023.1-ubuntu_jammy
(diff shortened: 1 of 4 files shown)


cardoe (commented):
Is pinning the images going to fix it when the issue is they bumped package versions in their repo that this is installing from?

skrobul:
> Is pinning the images going to fix it when the issue is they bumped package versions in their repo that this is installing from?

As discussed on Teams, Alpine does not offer any sensible way to pin the repository and retain old packages, so I have switched the dnsmasq container to use Debian base image.

cardoe (approved):

PR #68, opened 2024-04-26
fix: Updates README to point to the GitOps based install guide.

nicholaskuechler:
Some of the things in the old README were outdated and no longer worked. This updates the install guide to use the gitops based install: https://rackerlabs.github.io/understack/gitops-install/

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ The requirements are a Kubernetes cluster, which
 this repo will contain information about building different
 clusters and then a pile of bare metal systems that can
 be ingested into the stack to be available to be consumed
-via Openstack Ironic.
+via OpenStack Ironic.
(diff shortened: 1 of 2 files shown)


scripts/openstack-helm-sealed-secrets.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash -ex
```
cardoe, line 1: So `-e` means we have to catch every error otherwise the script fails. We're not handling the last one so this will make this script start failing.

cardoe (commented):
Not all of these scripts require bash out of the box.

cardoe (approved):

PR #73, opened 2024-04-30
fix: nautobot-interface-update-sensor

skrobul:
This sensor configuration was accidentally overwritten, most likely during rebase/cleanup. It was using the 'whalesay' only for debugging and initial development and should normally refer to `synchronize-server-to-ironic` as defined in the WorkflowTemplate.


mfencik (approved):
LGTM

PR #76, opened 2024-05-02
feat: refactor authentication to be more integrated through Dex

cardoe:
This refactors authentication to flow more through Dex and then allows operators/users to be added via Dex connectors to the system. Service accounts can be created within Keystone to have service to service authentication in the future as well.

### DIFF
--- a/components/dexidp/values-generic.yaml
+++ b/components/dexidp/values-generic.yaml
@@ -58,6 +58,11 @@ envVars:
       secretKeyRef:
         name: argo-sso
         key: client-secret
+  - name: ARGOCD_SSO_CLIENT_SECRET
+    valueFrom:
+      secretKeyRef:
+        name: argocd-sso
+        key: client-secret
 
 ingress:
   enabled: true
--- a/components/dexidp/values-generic.yaml
+++ b/components/dexidp/values-generic.yaml
@@ -27,10 +27,12 @@ config:
       config:
         # Required, without v3 suffix.
         keystoneHost: http://keystone-api.openstack.svc.cluster.local:5000
-        # Required, admin user credentials to connect to keystone.
+        # domain ID where users will authenticate against
         domain: default
+        # admin user credentials for getting refresh tokens for users without prompting them for creds
         keystoneUsername: demo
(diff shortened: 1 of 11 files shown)


cardoe:
There are at least two unresolved issues in this change:
- the domain ID for the Dex connector with Keystone for the `operator` domain
- always creating the "demo" user. We only want this created if the user set something or maybe didn't set something.

cardoe:
ArgoCD won't auth against Dex as well because it complains about the redirect URL not being correct. I don't know if its Dex or ArgoCD complaining cause there's nothing in the logs.

nicholaskuechler (approved):

scripts/easy-secrets-gen.sh
```diff
@@ -68,6 +68,23 @@ for ns in argo argo-events argocd dex; do
...
+    -o yaml \
+    --type Opaque \
+    --from-literal=issuer="https://dex.${DNS_ZONE}" \
+    --from-literal=client-secret="$ARGOCD_SSO_SECRET" \
+    --from-literal=client-id=argocd \
+    | yq '.metadata.labels |= {"app.kubernetes.io/part-of": "argocd"}' \
+    > "${DEST_DIR}/secret-argocd-sso-$ns.yaml"
+done
+unset ARGOCD_SSO_SECRET
+rm -rf "${DEST_DIR}/secret-argo-sso-argocd.yaml"
```
skrobul, line 86: should that `rm` line be moved to the paragraph above?

components/dexidp/values-generic.yaml
```diff
@@ -27,10 +27,12 @@ config:
       config:
         # Required, without v3 suffix.
         keystoneHost: http://keystone-api.openstack.svc.cluster.local:5000
-        # Required, admin user credentials to connect to keystone.
+        # domain ID where users will authenticate against
         domain: default
```
skrobul, line 31: small suggestion to make it more obvious that we expect ID, not a name here:
```suggestion
        domain: 0123456789abcd
```

skrobul (commented):
LGTM, few cosmetic issues

PR #87, opened 2024-05-10
feat: add multiple options and ranges to dnsmasq

cardoe:
Added the ability to manage multiple ranges and options in dnsmasq by defining multiple environment variables.


nicholaskuechler (approved):

skrobul:
Just a word of explanation about `#define LOPT_DHCP_AL_SVID 386` - the non-sequential number may seem odd, but it's on purpose. The `master` branch has introduced a bunch of options that have not made it to Debian's release yet. 386 was the next available number, so I kept it here to minimise the chance of conflict.

PR #94, opened 2024-05-15
feat: add argo workflow templates to argo cd application automation

andrrax:
Adding our argo-workflows and dependency manifests as an Argo CD Application.

### DIFF
--- a/apps/components/argo-workflows-templates.yaml
+++ b/apps/components/argo-workflows-templates.yaml
@@ -1,0 +1,29 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: argo-workflows-templates
+spec:
+  project: understack
+  sources:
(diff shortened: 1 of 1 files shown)


cardoe:
@skrobul I think you also mentioned some ideas here. Should we package up the workflows as a helm chart and have ArgoCD deploy that chart? Then each deployment can have its own values.yaml to override things.

skrobul:
> I am looking for thoughts on how I am handling the overlay repo here. Do we want to use the same UC_DEPOLOY overlay repo? 

If you see the use case for something like that, we can add an additional environment variable that is looked up first, if it's not present, we default to $UC_DEPLOY. Same thing for the references.

> Should we package up the workflows as a helm chart and have ArgoCD deploy that chart?

I think long term it makes sense, but for now let's just have something that deploys those workflows.

cardoe (approved):

PR #100, opened 2024-05-22
fix: adds handling for nautobot custom environment variables

nicholaskuechler:


scripts/gitops-secrets-gen.sh
```diff
@@ -111,4 +111,9 @@ if [ ! -f "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-metallb.yaml" ]; then
     echo "---" > "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-metallb.yaml"
 fi
 
+if [ ! -f "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-nautobot-env.yaml" ]; then
+    echo "Creating nautobot-env secret placeholder"
+    echo "---" > "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-nautobot-env.yaml"
```
cardoe, line 116: You'll want to create a secret with no data fields rather than an empty file. The empty file is good for a Helm values.yaml override but not for a secret that's being iterated.

cardoe (changes requested):

scripts/gitops-secrets-gen.sh
```diff
@@ -111,4 +111,19 @@ if [ ! -f "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-metallb.yaml" ]; then
...
+apiVersion: v1
+data:
+  NAUTOBOT_DEBUG: RmFsc2U=
+kind: Secret
+metadata:
+  creationTimestamp: null
+  name: nautobot-env
+  namespace: nautobot
+type: Opaque
+EOF
```
cardoe, line 126:
```suggestion
if [ ! -f "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-nautobot-env.yaml" ]; then
echo "Creating nautobot-env secret placeholder"
 kubectl --namespace nautobot \
        create secret generic nautobot-env \
        --dry-run=client \
        -o yaml \
        --type Opaque \ > "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-nautobot-env.yaml"
```

cardoe (changes requested):
I would use the command to create the secret like we do in all other places.

components/nautobot/values.yaml
```diff
@@ -47,6 +49,7 @@ celery:
   replicaCount: 1
   extraEnvVarsSecret:
     - nautobot-django
```
cardoe, line 51:
```suggestion
    - nautobot-django
```

This can be removed as well since you're not stuffing extra bits into there.

nicholaskuechler: If we remove it, then how does the secret key make it in to nautobot? I don't see any mention of "nautobot-django" in https://github.com/nautobot/helm-charts so it doesn't look like nautobot does that for us today.

cardoe: That's just the celery spot. AFAIK, celery doesn't need it?

cardoe: we can keep it for now then.

cardoe (approved):

PR #113, opened 2024-06-04
fix: applying necessary RBAC perms for workflow http templates to the…

andrrax:
… proper service account


andrrax:
These were originally applied to the wrong service account.

skrobul (approved):
needs rebase

skrobul (approved):

PR #115, opened 2024-06-05
ci: check argo workflows and workflow templates in pre-commit

cardoe:
Utilize pre-commit to validate the argo workflows and argo workflow templates against the published JSON schema.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,3 +25,12 @@ repos:
           - --line-length=120
           - --check
           - --diff
+  - repo: https://github.com/python-jsonschema/check-jsonschema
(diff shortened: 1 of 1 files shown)


nicholaskuechler (approved):

skrobul:
I like that, let's do it. 

btw, they also publish json schemas [at various locations](https://github.com/skrobul/lazyvim/blob/843960c67b53f2cd79eed965b030f9f78f8411fa/lua/plugins/lsp-config.lua#L63-L75) that can be used to verify more than just a WorkflowTemplates.

cardoe:
That's a better solution.

cardoe:
At least this adds a pre-commit rule to start.

PR #117, opened 2024-06-10
feat: incorporate sealed secrets into gen scripts

cringdahl:
- argocd Application 'sources' correctly reads 'kind: SealedSecret' and creates Secrets
- will prevent committing bare secrets to git

### DIFF
--- a/docs/gitops-install.md
+++ b/docs/gitops-install.md
@@ -114,6 +114,15 @@ your shell.
 
 TODO: some examples and documentation on how to build out a cluster
 
+### Bootstrapping ArgoCD and Sealed-Secrets
+
+If you do not have ArgoCD deployed then you can use the following:
+
+```bash
+cd ${UC_DEPLOY}
+kubectl kustomize --enable-helm https://github.com/rackerlabs/understack/bootstrap/ | kubectl apply -f -
+```
+
 ### Generating secrets
 
 Secrets in their very nature are sensitive pieces of data. The ultimate
@@ -166,14 +175,7 @@ At this point we will use our configs to make the actual deployment.
 Make sure everything you've committed to your deployment repo is pushed
 to your git server so that ArgoCD can access it.
 
-If you do not have ArgoCD deployed then you can use the following:
-
-```bash
-cd ${UC_DEPLOY}
-kubectl kustomize --enable-helm https://github.com/rackerlabs/understack/bootstrap/ | kubectl apply -f -
-```
-
-Now configure ArgoCD to be able to authenticate against Dex IdP.
+Configure ArgoCD to be able to authenticate against Dex IdP.
 
 ```bash
 kubectl -n argocd apply -f "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/secret-argocd-sso-argocd.yaml"
--- a/docs/gitops-install.md
+++ b/docs/gitops-install.md
@@ -118,12 +118,9 @@ TODO: some examples and documentation on how to build out a cluster
 
 Secrets in their very nature are sensitive pieces of data. The ultimate
 storage and injection of these in a production environment needs to be
-carefully considered. For the purposes of this document no specific
-choice has been made but tools like Vault, Sealed Secrets, SOPS, etc
-should be considered. This will only generate the necessary secrets
-using random data to successfully continue the installation.
-
-TODO: probably give at least one secure example
+carefully considered. For the purposes of this document, Sealed Secrets
+has been chosen; other tools like Vault, SOPS, etc should be considered
+for production deployments.
 
 ```bash
 # from your understack checkout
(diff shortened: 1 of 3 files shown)


docs/gitops-install.md
```diff
@@ -112,18 +112,15 @@ your shell.
 
 ### Populating the infrastructure
 
-TODO: some examples and documentation on how to build out a cluster
+For an example cluster setup, see our [documentation using Ubuntu and k3s](install-understack-ubuntu-k3s.md).
```
cardoe, line 115: This page isn't a good one to reference cause it does cluster creation and also all the stuff in this doc. It needs to go away and the parts we want to keep need to land else where.

scripts/easy-secrets-gen.sh
```diff
@@ -125,85 +139,31 @@ kubectl --namespace openstack \
...
     --from-literal=password="${IRONIC_KEYSTONE_PASSWORD}" \
-    --dry-run=client -o yaml > "${DEST_DIR}/secret-ironic-keystone-password.yaml"
+    --dry-run=client -o yaml | secret-seal-stdin "${DEST_DIR}/secret-ironic-keystone-password.yaml"
 
 if [ "x${DO_TMPL_VALUES}" = "xy" ]; then
     [ ! -f "${DEST_DIR}/secret-openstack.yaml" ] && \
     yq '(.. | select(tag == "!!str")) |= envsubst' \
         "./components/openstack-secrets.tpl.yaml" \
-        > "${DEST_DIR}/secret-openstack.yaml"
+        | secret-seal-stdin "${DEST_DIR}/secret-openstack.yaml"
```
cardoe, line 163: This file isn't a Kubernetes secret but is a Helm values.yaml with secrets. So Sealed Secrets won't work on it.

cardoe (changes requested):
So at least one of the files can't be sealed with Sealed Secrets. We should use a rebase instead of a merge as an aside. The cluster install doc isn't really one to reference.

docs/gitops-install.md
cringdahl: reverted

scripts/easy-secrets-gen.sh
cringdahl: Reverting; ArgoCD, multi-sourcing, and helm-secrets [aren't playing nice yet](https://github.com/jkroepke/helm-secrets/wiki/ArgoCD-Integration#multi-source-application-support). That's what we get for using bleeding edge technology.

scripts/easy-secrets-gen.sh
```diff
@@ -2,7 +2,22 @@
 
 cd $(git rev-parse --show-toplevel)
 
-DEST_DIR=${1:-.}
+DEST_DIR=${1:-components/01-secrets/}
```
cardoe, line 5: Why the path change here?

scripts/easy-secrets-gen.sh
```diff
@@ -2,7 +2,22 @@
...
+
+if ! type -p kubeseal kubectl > /dev/null; then
+    echo "You must have kubeseal & kubectl installed to use this script" >&2
```
cardoe, line 8: You also need `kubeseal` able to talk to your cluster as well.

scripts/easy-secrets-gen.sh
cringdahl: That's where all the docs were pointing.

scripts/easy-secrets-gen.sh
cringdahl: Do we really want to overload the script with context handling? Or is there some other thing you were thinking?

scripts/easy-secrets-gen.sh
andrrax: `DEST_DIR` looks like it was being used as a temporary location when secrets were created, and then once they're sealed they are then additionally written to `components/01-secrets/`. Since the new method is not writing them prior to being sealed, and using `DEST_DIR` only to write the sealed secrets, this looks change valid to me.

scripts/easy-secrets-gen.sh
andrrax: this should be handled with:

https://github.com/rackerlabs/understack/pull/117/files#diff-e2a91e11ec7b4c9466e228236ba28b9fa43e645cbb064eb86d3838f014d41861R12-R15

scripts/easy-secrets-gen.sh
andrrax: looking at this a bit more, I think what was there was probably preferred. if easy-secrets-gen is called without an arg, we'll assume writes to pwd. I believe 01-secrets is no longer being used.

andrrax (approved):

scripts/easy-secrets-gen.sh
cardoe: I was actually just about to follow up and say that `components/01-secrets` should be retired. I believe it's in the original doc which is why it hasn't been.

cardoe: The two secrets scripts need to get merged ultimately.

cardoe (approved):

PR #123, opened 2024-06-20
feat: utilize ApplicationSets to simplify deploying services

cardoe:
Simplify the deployment of services by utilizing ApplicationSets to perform templating instead of manual bash scripts to template our the Application yaml files.


cardoe:
Still definitely a WIP because the scripts are still there and only a handful of services are converted over. You need to create a secret in the `argocd` namespace that looks like.

```
apiVersion: v1
kind: Secret
stringData:
  config: '{"tlsClientConfig":{"insecure":false}}'
  name: "$DEPLOY_NAME"
  server: "https://kubernetes.default.svc"
metadata:
  name: $DEPLOY_NAME
  labels:
    argocd.argoproj.io/secret-type: cluster
  annotations:
    uc_repo_git_url: '$UC_REPO_GIT_URL'
    uc_repo_ref: 'appset'
    uc_deploy_git_url: '$UC_DEPLOY_GIT_URL'
    uc_deploy_ref: 'HEAD'
    dns_zone: '$DNS_ZONE'
```

cardoe:
I believe this should work. I think with a little bit more fiddling we can make cert-manager just a generic like the other operators can be.

nicholaskuechler (approved):

PR #127, opened 2024-06-24
feat: Adds OpenStack Horizon dashboard to UnderStack

nicholaskuechler:


apps/components/horizon.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
```
cardoe, line 2: This whole file needs to be deleted.

cardoe (changes requested):

cardoe (approved):

PR #135, opened 2024-07-02
fix: deployment of OpenStack Neutron and Nova

cardoe:
Attempt to fix the templating to be able to successfully deploy Neutron and Nova. Added Placement since it seemed to be required by Nova.

### DIFF
--- a/components/neutron/aio-values.yaml
+++ b/components/neutron/aio-values.yaml
@@ -12,6 +12,12 @@ endpoints:
     hosts:
       default: rabbitmq-nodes
 
+network:
+  # configure OpenStack Helm to use Undercloud's ingress
+  # instead of expecting the ingress controller provided
+  # by OpenStack Helm
+  use_external_ingress_controller: true
+
 # (nicholas.kuechler) updating the jobs list to remove the 'neutron-rabbit-init' job.
 dependencies:
   dynamic:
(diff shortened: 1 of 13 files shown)


components/openstack-secrets.tpl.yaml
```diff
@@ -76,4 +142,13 @@ secrets:
     identity:
       api:
         public: keystone-tls-public
+    network:
+      api:
```
nicholaskuechler, line 146: should be `server`

nicholaskuechler (changes requested):

cardoe:
This gets nova and neutron in a running state with some basic defaults targeting a testing environment.

components/neutron/aio-values.yaml
```diff
@@ -12,6 +12,39 @@ endpoints:
...
+    ml2_conf:
+      ml2:
+        # set the default ml2 backend as the logger for testing, actual deployment should override
+        mechanism_drivers: logger
+  neutron:
+    DEFAULT:
+      # we aren't using neutron routers
+      service_plugins: ""
+      # we don't want HA L3 routers
+      l3_ha: False
```
nicholaskuechler, line 36: yaml lint doesn't like this... should this just be `false`? Or maybe `"False"`?

cardoe: Yeah you're right. It's parsed in Python it turns out so it needed to be `"False"` like you suggested.

nicholaskuechler (approved):

PR #146, opened 2024-07-10
feat: simplify ApplicationSets and protect ArgoCD and add OpenStack pre-reqs

cardoe:
Simplified creation of ApplicationSets and added protections to ArgoCD to avoid deleting it or Applications. This adds OpenStack pre-reqs as well.


cardoe:
So after merging this you'll need to run `kubectl -n argocd apply -f apps/cluster-setup.yaml` to get things to start actually applying. The cluster-setup was excluded previously so it doesn't chain fully.

cardoe:
It won't break anything merging this and not applying. Just won't fully take effect. Follow ons will clean up some of the directories so that we don't get broken on merge.

cardoe:
This is already applied to the dev cluster.

nicholaskuechler (approved):

PR #147, opened 2024-07-10
fix: Set OpenStack container image pull policy to Always

nicholaskuechler:
When openstack updates the image in docker, such as for image docker.io/openstackhelm/neutron:2024.1-ubuntu_jammy, the default image pull policy of IfNotPresent prevents us from getting upstream improvements and bug fixes if the image is already cached in kubernetes. This change forces pulling images to ensure we're getting the latest upstream patches.

### DIFF
--- a/components/openstack-2024.1-jammy.yaml
+++ b/components/openstack-2024.1-jammy.yaml
@@ -1,5 +1,8 @@
 ---
 images:
+  # imagePullPolicy
+  pull_policy: "Always"
+
   tags:
     # these are common across all these OpenStack Helm installations
     bootstrap: "docker.io/openstackhelm/heat:2024.1-ubuntu_jammy"


skrobul (commented):
>  IfNotPresent prevents us from getting upstream improvements 

We can always just delete cached tag from the host or even pin to a specific sha256 version. This way we at least can have a predictable and reproducible environment.

With the `Always` policy, the underlying image can change and be applied when the pod restarts at say 11pm on Friday without any human interaction. 
IMO this would be acceptable if limited to a development environment, but in production I think we should be more explicit.

cardoe:
Yeah I'm all for ideas on how to do this better. Maybe using renovatebot to help?

skrobul:
I like that idea, it would at least gives us some audit trail to see when stuff changes.

cardoe:
So I agree for a production environment. This is can be for tracking current and we'll explicitly set it in releases.

cardoe (approved):

PR #154, opened 2024-07-12
fix: correct env vars for port sync workflow

andrrax:


nicholaskuechler (approved):

cardoe (approved):

skrobul:
-1 on this one. Why are we making the code less secure by leaking the token to the environment variable when we [already have the code to read secrets easily](https://github.com/rackerlabs/understack/blob/main/argo-workflows/ironic-nautobot-sync/code/ironic/secrets.py) ?

cardoe:
@skrobul thie was more to move us along. I do believe the entire secrets and connection details need an overhaul.

skrobul:
No probs, we have discussed this on the meeting with @andrrax and agreed to leave it as-is.

For future reference, https://blog.diogomonica.com//2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/ covers some of the problems with env vars.

PR #156, opened 2024-07-12
feat: adds initial neutron plugin

micahculpepper:
Initial commit of a plugin for neutron to handle provisioning calls. Proof of concept stage. Just logs context data when invoked.


components/openstack-2024.1-jammy.yaml
```diff
@@ -40,7 +40,7 @@ images:
     neutron_metadata: "docker.io/openstackhelm/neutron:2024.1-ubuntu_jammy"
     neutron_ovn_metadata: "docker.io/openstackhelm/neutron:2024.1-ubuntu_jammy"
     neutron_openvswitch_agent: "docker.io/openstackhelm/neutron:2024.1-ubuntu_jammy"
-    neutron_server: "docker.io/openstackhelm/neutron:2024.1-ubuntu_jammy"
+    neutron_server: "ghcr.io/rackerlabs/understack/neutron:af7d2c4"
```
nicholaskuechler, line 43: I noticed this one has a hash which differs from the 2023.1-jammy.yaml entry

micahculpepper: This was a workaround for kubernetes not pulling the new version of the container. I didn't update the 2023 file because we're not using it as far as I can tell.

Putting a hash in here isn't great, but I'm not sure what a better solution would be, in order to keep everything in sync. Because it's a hash, we'll have to update this file whenever a new version of the container is available, which is not ideal, but I'm not sure what to do instead.

nicholaskuechler (approved):

nicholaskuechler (approved):

.github/workflows/containers.yaml
```diff
@@ -44,8 +44,8 @@ jobs:
       - name: build and deploy container image to registry
         uses: docker/build-push-action@v5
         with:
-          context: "{{defaultContext}}:containers"
-          file: Dockerfile.${{ matrix.project }}
+          context: "{{defaultContext}}"
```
cardoe, line 47: Gotta fix up the "dnsmasq" container cause the file path changed for the `COPY` statements

micahculpepper: fixed in [0212056](https://github.com/rackerlabs/understack/pull/156/commits/021205665c4fc15f67b265becd40f49dca086925)

cardoe:
A bit of feedback though on the overall PR. Break the commits up into logical chunks that tell a story if you're going to have multiple commits. The 3 commits I would do:

1. add the Python code and the setup tools bits
2. add the build of the neutron container
3. use the neutron container in the images yaml

skrobul (approved):
+1

PR #162, opened 2024-07-16
feat: AMQP / Ironic event source

skrobul:
This EventSource provides events generated by the in-cluster Ironic instance and can be consumed to react to changes happening there.

### DIFF
--- a/components/argo-events/openstack-event-source.yaml
+++ b/components/argo-events/openstack-event-source.yaml
@@ -1,0 +1,33 @@
+kind: EventSource
+metadata:
+  name: openstack-amqp
+spec:
+  amqp:
+    openstack:
+      # amqp server url
+      url: amqp://rabbitmq-server-0.rabbitmq-nodes.openstack.svc.cluster.local:5672/ironic
+      routingKey: 'ironic_versioned_notifications.info'
+      # jsonBody specifies that all event body payload coming from this
+      # source will be JSON
+      jsonBody: true
+      # name of the exchange.
+      exchangeName: ironic
+      exchangeType: topic
+      exchangeDeclare:
+        durable: false
+      # optional consume settings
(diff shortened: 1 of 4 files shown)


nicholaskuechler (approved):

components/argo-events/openstack-event-source.yaml
```diff
@@ -0,0 +1,33 @@
+kind: EventSource
+metadata:
+  name: openstack-amqp
+spec:
+  amqp:
+    openstack:
+      # amqp server url
+      url: amqp://rabbitmq-server-0.rabbitmq-nodes.openstack.svc.cluster.local:5672/ironic
```
cardoe, line 8: Can we only support 1 node here?

skrobul: Unfortunately that seems to be the case - it did not work with comma separated, URL. I briefly looked into the code and it looks like Argo-Events simply [passes down](https://github.com/argoproj/argo-events/blob/d887e66f40dff3604f07755082758ba03c37204c/eventsources/sources/amqp/start.go#L109) the URL to the [amqp091 library's Dial function](https://pkg.go.dev/github.com/rabbitmq/amqp091-go#Dial) which does not seem to support anything but a single broker.

Additionally, the [project's non-goals](https://github.com/rabbitmq/amqp091-go?tab=readme-ov-file#non-goals) explicitly state that "Auto reconnect and re-synchronization of client and server topologies. " is not going to be supported.

cardoe: Well that stinks.

skrobul: Yeah. I didn't test this, but maybe we can create three instances of that EventSource, each with a different URL? But I don't know if there is an easy way to deduplicate events.

PR #175, opened 2024-07-22
chore: clean up image creation for OpenStack components

cardoe:
Dropped the 2023.1 images. Made the image building a bit more consistent with OpenStack in the dnsmasq section. Ensure that annotations are set on the index level due to provenance. Standardize on the "understack" namespace instead of "openstackhelm".


mfencik (approved):
+1

PR #182, opened 2024-07-26
feat: enable new ironic inspector and enable LLDP data from it

cardoe:
Enable the new ironic inspector that is in tree and enable the hooks necessary to run LLDP and populate data about our devices from that LLDP data.

see the following for more info:
https://docs.openstack.org/ironic/latest/admin/inspection.html
https://docs.openstack.org/ironic/latest/admin/inspection/managed.html
https://docs.openstack.org/ironic/latest/admin/inspection/hooks.html


khackworth (approved):
Seeing LLDP on the switch side as well now. +1

skrobul (approved):

PR #187, opened 2024-07-31
feat: neutron/undersync: change to fire & forget

skrobul:
We were not able to get Neutron into a state where it would wait for the `undersync-device` workflow to complete and react to the potential errors.

We have tried few variations:
- putting the `_move_to_network()` into a separate greenthread and periodically polling with greenthread.sleep(). This has made the workflow run in the background, but the whole Neutron instance has blocked
- similarly, replacing synchronous `time.sleep` with `greenthread.sleep` did not make any difference.
- initial Workflow trigger has been placed in the `bind_port` instead of `update_port_postcommit`. This also resulted in whole Neutron instance being blocked.

Finally we decided to simply fire the workflow and not check for the result and rely on other mechanisms to be notified about potential errors.

Related docs:

https://docs.openstack.org/neutron-lib/1.9.0/reference/modules/neutron_lib.plugins.ml2.html https://opendev.org/x/networking-arista
https://github.com/dave-tucker/odl-neutron-drivers/blob/master/odldrivers/ml2/mech_driver.py#L28 https://github.com/markmcclain/altocumulus/blob/8a82eca43ef51b9f0714baba4096a839b0b7f6e9/altocumulus/ml2/driver.py

### DIFF
--- a/python/neutron-understack/neutron_understack/neutron_understack_mech.py
+++ b/python/neutron-understack/neutron_understack/neutron_understack_mech.py
@@ -5,6 +5,8 @@ import typing
 import uuid
 
 import neutron_lib.api.definitions.portbindings as portbindings
+from neutron_lib import constants as p_const
+from neutron_lib.plugins.ml2 import api
 from neutron_lib.plugins.ml2.api import MechanismDriver
 from neutron_lib.plugins.ml2.api import NetworkContext
 from neutron_lib.plugins.ml2.api import PortContext
@@ -13,6 +15,7 @@ from neutron_understack.argo.workflows import ArgoClient
 from oslo_config import cfg
 
 LOG = logging.getLogger(__name__)
+
 
 def setup_conf():
     grp = cfg.OptGroup("ml2_type_understack")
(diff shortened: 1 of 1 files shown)


python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -202,15 +200,54 @@ def update_port_precommit(self, context):
...
+        Verify the requested segment is supported by Understack and return True or
+        False to indicate this to callers.
+        """
+        network_type = segment[api.NETWORK_TYPE]
+        return network_type in [
+            p_const.TYPE_LOCAL,
+            p_const.TYPE_GRE,
+            p_const.TYPE_VXLAN,
+            p_const.TYPE_VLAN,
+            p_const.TYPE_FLAT,
```
cardoe, line 249: I think ultimately we'll want to cull this list down.

cardoe (approved):

python/neutron-understack/neutron_understack/neutron_understack_mech.py
skrobul: Agreed, we probably can get away with just a TYPE_FLAT

PR #198, opened 2024-08-05
chore: fix type annotations

micahculpepper:
typing.Union is no longer required as of python 3.10 and set notation may now be used. This resulted in a few other formatting changes.


cardoe (commented):
You want to add some `ruff` rules to this package so that we catch and upgrade this in the future?

cardoe (approved):

PR #203, opened 2024-08-07
feat: use External Secrets to configure SSO for apps

cardoe:
Instead of having to duplicate the SSO secrets between the dex namespace and the application specific namespaces, use the External Secrets Operator to copy it and keep it in sync for us. Broke up the ArgoCD SSO configuration into its own component to not have it broken on initial deploy and have to restart. This simplifies the docs as well.

### DIFF
--- a/bootstrap/argocd/components/sso/external-secret-argocd-sso.yaml
+++ b/bootstrap/argocd/components/sso/external-secret-argocd-sso.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: argocd-sso
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: dex
+  target:
+    name: argocd-sso
+    creationPolicy: Owner
+    deletionPolicy: Delete
+    template:
+      engineVersion: v2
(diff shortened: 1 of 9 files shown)


nicholaskuechler (approved):

bootstrap/argocd/components/sso/external-secret-argocd-sso.yaml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: argocd-sso
+spec:
+  refreshInterval: 1h
```
skrobul, line 7: Can we force the sync? or does it monitor the source resources for changes?

skrobul (approved):

bootstrap/argocd/components/sso/external-secret-argocd-sso.yaml
cardoe: That's how often it checks the source for changes. I'm not sure how to force stuff without using an annotation. Which I would love to do but just don't know how yet.

skrobul: Found this:

`kubectl annotate es name-of-es force-sync=$(date +%s) --overwrite`

PR #206, opened 2024-08-08
feat: create infra AppSet and add components to it

cardoe:
Created an infra AppSet and add metallb, ingress-nginx, and cert-manager to it so that we can manage the ingress infra components together. Added the ability to disable installing metallb which is necessary when using k3s in a VM for example. This needs some more experimentation before it's formally documented as the method. Added another installation of ingress-nginx so that we can have a provisioning ingress controller as well.


apps/appsets/infra.yaml
```diff
@@ -0,0 +1,113 @@
...
+                              enabled: true
+                              default: false
+                              controllerValue: "k8s.io/provisioning-ingress-nginx"  # default: k8s.io/ingress-nginx
+                        valueFiles:
+                          - $secrets/helm-configs/{{.name}}/provisioning-ingress-nginx.yaml
+                        ignoreMissingValueFiles: true
+            selector:
+              # by setting the key in the elements 'noInstall' to 'true' it will skip installing it
+              matchExpressions:
+                - key: noInstall
```
skrobul, line 76: what do you think about renaming this to `skipInstallation` ?

skrobul (approved):
Nice work! 
Should we also provide an option to skip the installation of cert-manager and ingress-nginx if the user already has their ingress controller?

skrobul (approved):

PR #213, opened 2024-08-14
feat: convert nautobot to AppSet

cardoe:
Converted nautobot to the component AppSet. Dropped the old Application bits.


cardoe:
Now I need to sit down and go through the documentation....

nicholaskuechler (approved):

PR #215, opened 2024-08-15
feat: set interface/port names in ironic

cardoe:
Set the nautobot interface name on the corresponding ironic port on the node.


skrobul (approved):
LGTM. Do we need to cleanup the port name in any way?

PR #225, opened 2024-08-19
(WIP) feat: Adds rendering of argo workflows as mkdocs mermaid diagrams

nicholaskuechler:


skrobul (commented):
⁠so I have run out of time to properly review this today, but wanted to check if the rendered graphs are actually correct. your PR screenshot included sync-srv-redfish-intfs-to-nb which has 3 sequential steps, but they were drawn as parallel - was it just an example and the final rendering is correct or not?

If not, can this script be improved to handle this?

nicholaskuechler:
> ⁠so I have run out of time to properly review this today, but wanted to check if the rendered graphs are actually correct. your PR screenshot included sync-srv-redfish-intfs-to-nb which has 3 sequential steps, but they were drawn as parallel - was it just an example and the final rendering is correct or not?
> 
> If not, can this script be improved to handle this?

Fixed it, thanks!

skrobul (approved):

PR #226, opened 2024-08-20
feat: Map Ironic provisioning status to higher level Nautobot Status

skrobul:
Closes PUC-408

Goes together with https://github.com/RSS-Engineering/undercloud-nautobot/pull/187


argo-workflows/ironic-to-nautobot-sync/workflowtemplates/sync.yaml
```diff
@@ -12,13 +12,11 @@ spec:
       container:
         image: ghcr.io/rackerlabs/understack/ironic-nautobot-client:latest
         command:
-          - nautobot-update-cf
+          - synchronize-provision-state-to-nautobot
```
cardoe, line 15: here it's `synchronize-provision-state-to-nautobot` but....

python/understack-workflows/pyproject.toml
```diff
@@ -44,6 +44,7 @@ pytest-mock = "^3.14.0"
 synchronize-interfaces = "understack_workflows.main.synchronize_interfaces:main"
 synchronize-obm-creds = "understack_workflows.main.synchronize_obm_creds:main"
 synchronize-server = "understack_workflows.main.synchronize_server:main"
+synchronize-provision-state = "understack_workflows.main.synchronize_provision_state:main"
```
cardoe, line 47: here it's `synchronize-provision-state`

cardoe (commented):
Just a small disconnect in the workflow YAML from the executable name. It does make me wonder if we should call the binary `synchronize-ironic-to-nautobot` or something like that. Since you have the `--provision-state` as a flag. Maybe in the future there's other bits we want to push and we can add more flags?

argo-workflows/ironic-to-nautobot-sync/workflowtemplates/sync.yaml
skrobul: good catch, it's supposed to be `synchronize-provision-state`, will fix.

cardoe (approved):

PR #232, opened 2024-08-21
chore: some refactoring of the workflow code and names

cardoe:
Refactored some of the workflow code to use a common nautobot url/token parser, with tests. Trimmed the name synchronize to sync. Created a common spot for shared fixtures.


argo-workflows/sync-nb-server-to-ironic/docs/README.md
```diff
@@ -2,7 +2,7 @@
 
 # Overview
 
-The purpose of the workflow templates in this folder is to synchronize
+The purpose of the workflow templates in this folder is to sync
```
skrobul, line 5: not sure if we need to shorten it here too

skrobul (approved):

argo-workflows/sync-nb-server-to-ironic/docs/README.md
cardoe: reverted.

PR #246, opened 2024-08-29
feat: adds demo for the ingest server workflow

nicholaskuechler:


argo-workflows/ingest-server/docs/dumps.json
```diff
@@ -0,0 +1,106 @@
+root@openstack-admin-client:/# cat node-12345678-legacy.json
```
stevekeay, line 1: Errm?

PR #247, opened 2024-08-30
feat: Rook / Ceph deployment for our storage needs

skrobul:
This PR adds basic deployment of Rook so that we can use PVCs in the development cluster.

It has been tested on the new uc-iad3-dev and has deployed successfully:

```
❯   kubectl --namespace rook-ceph get cephcluster
NAME        DATADIRHOSTPATH   MONCOUNT   AGE     PHASE   MESSAGE                        HEALTH        EXTERNAL   FSID
rook-ceph   /var/lib/rook     2          9m12s   Ready   Cluster created successfully   HEALTH_WARN              8ea3e27d-f56a-4477-88de-e499ab5f1795
```

The Health Warning is related to an insufficient number of OSDs deployed because we only have two nodes but a minimum of three are required.

Partially delivers PUC-495

### DIFF
--- a/operators/rook/values-cluster.yaml
+++ b/operators/rook/values-cluster.yaml
@@ -1,0 +1,187 @@
+# Defaults:
+# https://github.com/rook/rook/blob/release-1.15/deploy/charts/rook-ceph-cluster/values.yaml
+
+# TODO: adjust or remove this section when we have 3rd node
+cephClusterSpec:
+  mon:
+    # Set the number of mons to be started. Generally recommended to be 3.
+    # For highest availability, an odd number of mons should be specified.
+    count: 2
+    # The mons should be on unique nodes. For production, at least 3 nodes are recommended for this reason.
+    # Mons should only be allowed on the same node for test environments where data loss is acceptable.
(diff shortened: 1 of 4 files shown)


operators/rook/values-cluster.yaml
```diff
@@ -0,0 +1,187 @@
+# Defaults:
+# https://github.com/rook/rook/blob/release-1.15/deploy/charts/rook-ceph-cluster/values.yaml
+
+# TODO: adjust or remove this section when we have 3rd node
+cephClusterSpec:
+  mon:
+    # Set the number of mons to be started. Generally recommended to be 3.
+    # For highest availability, an odd number of mons should be specified.
+    count: 2
```
cardoe, line 9: If we leave this at 3, can we then make a kustomize patch for each deployment to adjust it since it's being customized?

cardoe (approved):

operators/rook/values-cluster.yaml
skrobul: The intention is to test it on the new cluster, then keep it to the default (3) so not sure if we need it long term

PR #248, opened 2024-08-30
fix: improve argo-events and the openstack event source

cardoe:
Switched argo-events to run cluster wide since each namespace has its own resources. This makes it easier for us to not have to copy secrets around so removed that case. Updated the openstack event source to be narrower in access scope and use the generated credentials by the rabbitmq operator removing the need for a generated secret at all. Added a Pod Disruption Budget to the event bus so we do not lose events.

### DIFF
--- a/apps/understack-workflows/eventsource-openstack/argo-rabbitmq.yaml
+++ b/apps/understack-workflows/eventsource-openstack/argo-rabbitmq.yaml
@@ -3,30 +3,23 @@ apiVersion: rabbitmq.com/v1beta1
 kind: User
 metadata:
   name: argo
-  namespace: openstack
 spec:
-  tags:
-  - management  # available tags are 'management', 'policymaker', 'monitoring' and 'administrator'
-  - policymaker
   rabbitmqClusterReference:
     name: rabbitmq  # rabbitmqCluster must exist in the same namespace as this resource
     namespace: openstack
-  importCredentialsSecret:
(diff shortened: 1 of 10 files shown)


apps/understack-workflows/eventsource-openstack/argo-rabbitmq.yaml
```diff
@@ -3,30 +3,23 @@ apiVersion: rabbitmq.com/v1beta1
...
 spec:
   vhost: "ironic"
   userReference:
     name: "argo"  # name of a user.rabbitmq.com in the same namespace; must specify either spec.userReference or spec.user
   permissions:
-    write: ".*"
-    configure: ".*"
-    read: ".*"
+    write: "^$"
+    configure: "^$"
```
skrobul, line 21: have you tested this? IIRC removing configure permission resulted in EventSource failure, but cannot remember what was it exactly

cardoe: So it should only result in an EventSource failure if we're changing anything from the config that creates that. It's currently deployed and not reporting any errors.

```
Status:
  Conditions:
    Last Transition Time:  2024-08-30T21:03:17Z
    Status:                True
    Type:                  Deployed
    Last Transition Time:  2024-08-30T21:03:17Z
    Status:                True
    Type:                  SourcesProvided
```

PR #249, opened 2024-08-30
feat: workflow updates Nautobot Interface.status instead of device.custom_field.conected_to_network

stevekeay:
We used to flip a bit on the nautobot Device object for “connected_to_network”
when changing a device from/to “provisioning” and “tenant” settings.

We need this to be interface specific.  Because of Ironic's knowlege of the
interface that is being used for PXE boot, we are called with a server interface
UUID.

We update the nautobot status field of connected switch interface.

This tells undersync which mode the port should be in - provisioning or
tenant (normal).

This would be released in tandem with https://github.com/RSS-Engineering/undersync/pull/14 fixes PUC-425

### DIFF
--- a/python/neutron-understack/neutron_understack/neutron_understack_mech.py
+++ b/python/neutron-understack/neutron_understack/neutron_understack_mech.py
@@ -263,14 +263,16 @@ class UnderstackDriver(MechanismDriver):
             return "tenant"
 
     def _move_to_network(self, context):
+        interface_uuid = context.current["id"]
         device_uuid = context.current["binding:host_id"]
         network_name = self.__network_name(context.current["network_id"])
-        LOG.debug(f"Selected {network_name=} for {device_uuid=}")
+        LOG.debug(f"Selected {network_name=} for {device_uuid=} {interface_id=}")
 
         result = argo_client.submit(
             template_name="undersync-device",
             entrypoint="trigger-undersync",
             parameters={
+                "interface_uuid": interface_uuid,
                 "device_uuid": device_uuid,
                 "network_name": network_name,
                 "dry_run": cfg.CONF.ml2_type_understack.argo_dry_run,
--- a/python/neutron-understack/neutron_understack/neutron_understack_mech.py
+++ b/python/neutron-understack/neutron_understack/neutron_understack_mech.py
@@ -263,16 +263,16 @@ class UnderstackDriver(MechanismDriver):
             return "tenant"
 
     def _move_to_network(self, context):
-        interface_uuid = context.current["id"]
+        interface_mac = context.current["mac_address"]
         device_uuid = context.current["binding:host_id"]
         network_name = self.__network_name(context.current["network_id"])
-        LOG.debug(f"Selected {network_name=} for {device_uuid=} {interface_id=}")
+        LOG.debug(f"Selected {network_name=} for {device_uuid=} {interface_mac=}")
 
         result = argo_client.submit(
             template_name="undersync-device",
             entrypoint="trigger-undersync",
             parameters={
-                "interface_uuid": interface_uuid,
+                "interface_mac": interface_mac,
                 "device_uuid": device_uuid,
                 "network_name": network_name,
                 "dry_run": cfg.CONF.ml2_type_understack.argo_dry_run,
--- a/python/neutron-understack/neutron_understack/tests/test_neutron_understack_mech.py
+++ b/python/neutron-understack/neutron_understack/tests/test_neutron_understack_mech.py
@@ -1,0 +1,37 @@
+from dataclasses import dataclass
+from unittest.mock import patch, MagicMock
+import pytest
+from neutron_understack.argo.workflows import ArgoClient
+
+from neutron_understack.neutron_understack_mech import UnderstackDriver
+
+# TODO: I am not sure how we run tests in this project.
+
+@dataclass
+class ContextDouble:
+  current: dict
+
+def mock_context_data(file):
+    ref = pathlib.Path(__file__).joinpath(filename)
+    with ref.open("r") as f:
+        return ContextDouble(json.load(f))
+
+@patch('neutron_understack.argo.workflows.ArgoClient')
+def test_update_port_postcommit(mock_argo_client):
+    context_data = mock_context_data(
+        "fixtures/neutron_update_port_postcommit.json"
+    )
+
+    UnderstackDriver.update_port_postcommit(context_data)
+
+    mock_argo_client.assert_called_once_with(
+        template_name="undersync-device",
+        entrypoint="trigger-undersync",
+        parameters={
+            "interface_uuid": "e5d5cd73-ca9a-4b74-9d52-43188d0cdcaa",
+            "device_uuid": "41d18c6a-5548-4ee9-926f-4e3ebf43153f",
+            "network_name": "provisioning",
+            "dry_run": false,
+            "force": false,
+        },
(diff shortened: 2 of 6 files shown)


python/understack-workflows/understack_workflows/nautobot.py
```diff
@@ -120,6 +127,22 @@ def update_cf(self, device_id: UUID, field_name: str, field_value: str):
         self.logger.info(f"save result: {response}")
         return response
 
+    def update_switch_interface_status(self, server_interface_id: UUID, new_status: str) -> NautobotInterface:
+        server_interface = self.interface_by_id(server_interface_id)
+        connected_endpoint = server_interface.connected_endpoint
+        if not connected_endpoint:
+            raise Exception("Interface {server_interface_id} not connected in Nautobot")
```
skrobul, line 134: missing f-prefix without which the string interpolation would not work

```suggestion
            raise Exception(f"Interface {server_interface_id} not connected in Nautobot")
```

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -263,14 +263,16 @@ def __network_name(self, network_id: str):
             return "tenant"
 
     def _move_to_network(self, context):
+        interface_uuid = context.current["id"]
```
skrobul, line 266: I am afraid this won't be as simple, but I am not 100% sure on that.

The way I understand it: 
- the ml2 driver calls [`update_port_postcommit`](https://docs.openstack.org/neutron-lib/latest/reference/modules/neutron_lib.plugins.ml2.api.html#neutron_lib.plugins.ml2.api.MechanismDriver.update_port_postcommit) with a relevant context that is of [PortContext](https://docs.openstack.org/neutron-lib/latest/reference/modules/neutron_lib.plugins.ml2.api.html#neutron_lib.plugins.ml2.api.PortContext) type
- this gets passed to `_move_to_network` without modifications

Now, the problem I see here is that `id` fields corresponds to the ID of a port as given by Neutron. Unless I'm wrong, these are completely different to what Nautobot expects to identify an Interface.

These IDs are automatically assigned when Ironic [creates](https://opendev.org/openstack/ironic/src/branch/master/ironic/common/neutron.py#L359) Neutron ports for each of its [baremetal Ports](https://docs.openstack.org/ironic/latest/contributor/api/ironic.objects.port.html#ironic.objects.port.Port).

At the moment, both baremetal ports and neutron port IDs will not match as they are both randomly allocated - on baremetal side during the node inspection, and on the neutron side when the code linked above is executed.

I think the easiest way around that situation may be to match using the MAC address. Alternatively, we will have to pre-create the neutron ports with specific IDs which may be tricky.

skrobul: One clarification on this - even though Ironic initially [copies the baremetal port attributes](https://opendev.org/openstack/ironic/src/commit/3f22f8c5f00e6a1af9f221df9b22698a02baa245/ironic/common/neutron.py#L305) presumably including the baremetal port id, it won't be preserved because Python Openstack SDK (which is used to interface with Neutron), [does not pass on](https://docs.openstack.org/openstacksdk/latest/user/connection.html#openstack.connection.Connection.create_port) that attribute.

Even if it did, [Neutron API](https://docs.openstack.org/api-ref/network/v2/#create-port) wouldn't accept it

```
root@openstack-admin-client:/tmp# cat create_port_with_id.json
{
    "port": {
        "id": "bc1ab041-13af-464c-81c6-87db1e15aaaa",
        "admin_state_up": true,
        "name": "private-port",
        "network_id": "a0c267bf-33af-48b9-b540-daa6d41bff3f",
        "port_security_enabled": false,
        "allowed_address_pairs": []
    }
}

root@openstack-admin-client:/tmp# curl -X POST -H "Content-Type: application/json" -H "X-Auth-Token: $TOKEN" http://neutron-server.openstack.svc.cluster.local:9696/v2.0/ports -d @create_port_with_id.json
{"NeutronError": {"type": "HTTPBadRequest", "message": "Attribute 'id' not allowed in POST", "detail": ""}}root@openstack-admin-client:/tmp#
root@openstack-admin-client:/tmp#
```

stevekeay: I thought it was too good to be true when someone said "that is the ID you need".  I reworked this in 35d1295 to find the interface by MAC address, we can revisit this in the future if we ever sync up the UUID values.

python/neutron-understack/neutron_understack/tests/test_neutron_understack_mech.py
```diff
@@ -0,0 +1,37 @@
...
+# TODO: I am not sure how we run tests in this project.
+
+@dataclass
+class ContextDouble:
+  current: dict
+
+def mock_context_data(file):
+    ref = pathlib.Path(__file__).joinpath(filename)
+    with ref.open("r") as f:
+        return ContextDouble(json.load(f))
```
cardoe, line 17: I'd probably add in some more magic with fixtures.
```suggestion
        return ContextDouble(json.load(f))
        
@pytest.fixture
def neutron_update_port_postcommit() -> ContextDouble:
    return mock_context_data("fixtures/neutron_update_port_postcommit.json")
```

python/neutron-understack/neutron_understack/tests/test_neutron_understack_mech.py
```diff
@@ -0,0 +1,37 @@
...
+
+@patch('neutron_understack.argo.workflows.ArgoClient')
+def test_update_port_postcommit(mock_argo_client):
+    context_data = mock_context_data(
+        "fixtures/neutron_update_port_postcommit.json"
+    )
+
+    UnderstackDriver.update_port_postcommit(context_data)
```
cardoe, line 25:
```suggestion
def test_update_port_postcommit(mock_argo_client, neutron_update_port_postcommit):

    UnderstackDriver.update_port_postcommit(neutron_update_port_postcommit)
```

python/neutron-understack/neutron_understack/tests/test_neutron_understack_mech.py
```diff
@@ -0,0 +1,37 @@
...
+
+    mock_argo_client.assert_called_once_with(
+        template_name="undersync-device",
+        entrypoint="trigger-undersync",
+        parameters={
+            "interface_uuid": "e5d5cd73-ca9a-4b74-9d52-43188d0cdcaa",
+            "device_uuid": "41d18c6a-5548-4ee9-926f-4e3ebf43153f",
```
cardoe, line 32:
```suggestion
            "interface_uuid": neutron_update_port_postcommit.current["id"],
            "device_uuid": neutron_update_port_postcommit.current["device_id"],
```

cardoe (commented):
So won't block this being merged but I think this would be an improvement for the test.

python/neutron-understack/neutron_understack/neutron_understack_mech.py
cardoe: @skrobul We can go directly to python-neutronclient instead of using the OpenStack SDK. I've noticed the OpenStack SDK is a bit too high level in a number of places. It aims to be a higher level simple interface and I guess that makes sense.

cardoe: My comments doesn't matter because like you said, neutron rejects that.

cardoe (approved):

PR #250, opened 2024-09-03
feat: update sync server workflow to return node state

mudakavip:
Changes in this PR are:
1. Python code update to sync_server.py to return state of the ironic node
2. Argo workflow update to sync_server_to_ironic.yaml file to return output results for next steps to utilize.

fixes PUC-494

### DIFF
--- a/argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
+++ b/argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
@@ -12,27 +12,26 @@ spec:
         value: "{}"
   templates:
     - name: main
-      input:
-        parameters:
-          - name: interface_update_event
-      output:
-        parameters:
-          - name: server_state
-            valueFrom:
-              expression: "{{ steps.sync-server.outputs.parameters.result }}"
       steps:
         - - name: sync-server
+            input:
+              parameters:
+                - name: interface_update_event
+            output:
+              parameters:
+                - name: server_state
+                  value: "{{ steps.sync-server.outputs.parameters.result.server_state }}"
             container:
-            image: ghcr.io/rackerlabs/understack/ironic-nautobot-client:latest
-            command:
-              - sync-server
-            args:
-              - "{{input.parameters.interface_update_event}}"
-            volumeMounts:
-              - mountPath: /etc/ironic-secrets/
-                name: ironic-secrets
-                readOnly: true
+              image: ghcr.io/rackerlabs/understack/ironic-nautobot-client:latest
+              command:
+                - sync-server
+              args:
+                - "{{input.parameters.interface_update_event}}"
+              volumeMounts:
+                - mountPath: /etc/ironic-secrets/
+                  name: ironic-secrets
+                  readOnly: true
             volumes:
               - name: ironic-secrets
                 secret:
-                secretName: production-ironic-for-argo-creds
+                  secretName: production-ironic-for-argo-creds
--- a/argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
+++ b/argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
@@ -11,18 +11,28 @@ spec:
       - name: interface_update_event
         value: "{}"
   templates:
-    - name: sync-server
-      container:
-        image: ghcr.io/rackerlabs/understack/ironic-nautobot-client:latest
-        command:
-          - sync-server
-        args:
-          - "{{workflow.parameters.interface_update_event}}"
-        volumeMounts:
-          - mountPath: /etc/ironic-secrets/
-            name: ironic-secrets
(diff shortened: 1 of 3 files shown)


argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
```diff
@@ -11,18 +11,28 @@ spec:
...
+      input:
+        parameters:
+          - name: interface_update_event
+      output:
+        parameters:
+          - name: server_state
+            valueFrom:
+              expression: "{{ steps.sync-server.outputs.parameters.result }}"
+      steps:
+        - - name: sync-server
```
cardoe, line 24: Why did you make this to be a nested workflow instead of just putting the input and output parameters on the existing step? I'm looking at https://argo-workflows.readthedocs.io/en/latest/walk-through/output-parameters/

python/understack-workflows/understack_workflows/main/sync_server.py
```diff
@@ -77,3 +77,8 @@ def main():
     response = client.update_node(node.uuid, patches)
     logger.info(f"Patching: {patches}")
     logger.info(f"Updated: {response}")
+    return ironic_node.provision_state
+
+
+if __name__ == "__main__":
+    main()
```
cardoe, line 84: Since the value is returned by the main function, how is it being captured by the container runner? Seems like you'd either have to grab it from stdout or from a file?

cardoe (commented):
Some questions

python/understack-workflows/understack_workflows/main/sync_server.py
mudakavip: Made changes to send the output to stdout.

argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
mudakavip: I was just trying to follow how other workflows were setup but made changes accordingly please review and let me know if any other changes are needed

cardoe: So this converts this from a single step workflow (a single drop in function) into a chain of steps workflow with just 1 step. This means we'll get the overhead of a workflow runner instead of just the one step. I would revert this back.

In fact this change will make it that the output parameters won't work. Because you're just returning the output parameter into the chain of steps and not out of the actual workflow itself.

cardoe (changes requested):

python/understack-workflows/understack_workflows/main/sync_server.py
```diff
@@ -77,3 +77,8 @@ def main():
     response = client.update_node(node.uuid, patches)
     logger.info(f"Patching: {patches}")
     logger.info(f"Updated: {response}")
+    return ironic_node.provision_state
+
+
+if __name__ == "__main__":
+    sys.stdout.write(main())
```
cardoe, line 84: This also won't work with how Python scripts are installed with a Python package. The `__main__` part isn't called and instead the function is called directly, which shouldn't return anything.

cardoe (changes requested):

argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
cardoe: https://github.com/argoproj/argo-workflows/blob/main/examples/output-parameter.yaml so that's from their docs for example.

mudakavip: ok got it, will revert it back to the nested workflow that I had

mudakavip: I have used the text file generated by the script and returning the output from the workflow in single step.

argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
```diff
@@ -11,7 +11,12 @@ spec:
       - name: interface_update_event
         value: "{}"
   templates:
-    - name: sync-server
+    - name: main
```
cardoe, line 14:
```suggestion
    - name: sync-server
```

The parent workflow has a ref to `sync-server` so it will break that. We need to ultimately add some tests to check references.

cardoe (changes requested):

argo-workflows/sync-nb-server-to-ironic/workflowtemplates/sync-server-to-ironic.yaml
mudakavip: oops, I missed that when I was changing things. Fixed it now

cardoe (approved):

PR #255, opened 2024-09-05
ci: build and clean up containers on each PR

cardoe:
Make it possible to test our containers by building and pushing the containers for the life of the PR. Once the PR is closed we'll delete and clean up the containers.

### DIFF
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -85,8 +85,6 @@ jobs:
             type=raw,value=latest,enable={{is_default_branch}}
             type=raw,value=latest,enable=${{ github.event_name == 'workflow_dispatch' }}
             type=sha,enable={{is_default_branch}}
-          labels: |
-            org.opencontainers.image.title=dnsmasq for Understack's Ironic
         env:
           # Create the annotations at the index as well since this
           # defaults to manifest only and we have to manually merge
@@ -111,7 +109,6 @@ jobs:
       matrix:
         container:
           - name: ironic-nautobot-client
-            title: Understack Ironic Nautobot Clients
 
     steps:
       - name: setup docker buildx
@@ -134,8 +131,6 @@ jobs:
             type=raw,value=latest,enable=${{ github.event_name == 'workflow_dispatch' }}
             type=ref,event=tag
             type=ref,event=pr
-          labels: |
-            org.opencontainers.image.title=${{ matrix.container.title }}
         env:
           # Create the annotations at the index as well since this
           # defaults to manifest only and we have to manually merge
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -9,6 +9,7 @@ on:
       - ".github/workflows/containers.yaml"
       - "python/**"
   pull_request:
+    types: [opened, synchronize, reopened, closed]
     paths:
       - "containers/**"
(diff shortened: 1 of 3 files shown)


.github/workflows/containers.yaml
```diff
@@ -149,9 +144,59 @@ jobs:
...
+
+    steps:
+      - name: clean up PR container
+        uses: actions/github-script@v7
+        env:
+          CONTAINER_NAME: '${{ matrix.container }}'
+        with:
+          script: |
+            const container_name = `understack%2F${process.env.CONTAINER_NAME}`;
+            const response = github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
```
skrobul, line 177: The `getAllPackageVersionsForPackageOwnedByOrg` seems to [be an async](https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/ec6e43c0d80cfbb1fa7768b7dc43368abaa63ad2/src/generated/method-types.ts#L7817) function so should be awaited
   
```suggestion
            const response = await github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
```

.github/workflows/containers.yaml
```diff
@@ -149,9 +144,59 @@ jobs:
...
+              package_type: "container",
+              package_name: process.env.CONTAINER_NAME,
+              org: context.repo.owner,
+            });
+
+            const target_tag = `pr-${context.payload.pull_request.number}`;
+            console.log(`Looking for tag ${target_tag}` for container ${container_name}`);
```
skrobul, line 184:
```suggestion
            console.log(`Looking for tag ${target_tag} for container ${container_name}`);
```

.github/workflows/containers.yaml
```diff
@@ -149,9 +144,59 @@ jobs:
...
+              version.metadata.container.tags.includes(target_tag)
+            );
+
+            if (matchingVersion) {
+              console.log(`Found tag to delete "${target_tag}":`, matchingVersion);
+              await github.rest.packages.deletePackageVersionForOrg({
+                package_type: "container",
+                package_name: container_name,
+                org: context.repo.owner,
+                package_version_id: pkg_id,
```
skrobul, line 197: should that `pkg_id` be replaced with `matchingVersion` ?

.github/workflows/containers.yaml
```diff
@@ -149,9 +144,59 @@ jobs:
...
+            const response = github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
+              package_type: "container",
+              package_name: process.env.CONTAINER_NAME,
+              org: context.repo.owner,
+            });
+
+            const target_tag = `pr-${context.payload.pull_request.number}`;
+            console.log(`Looking for tag ${target_tag}` for container ${container_name}`);
+
+            const versions = response.data;
```
skrobul, line 186:
```suggestion
            const versions = response.data || [];
```

.github/workflows/containers.yaml
cardoe: I bungled this PR. Thank you for catching all this.

cardoe:
https://github.com/rackerlabs/understack/pkgs/container/understack%2Fironic-nautobot-client/270640358?tag=pr-255 published by this PR

PR #257, opened 2024-09-06
feat: development shells with Devbox

skrobul:
This PR 
- ~converts `neutron_understack` to use Poetry instead of setuptools~
- adds [Devbox](https://www.jetify.com/devbox/) integration greatly simplifies setting up the local development environment and makes it consistent. It can even be used in Github's CI.

    Three separate environments were created with (hopefully) appropriate tools inside them:
    - `ironic_understack`
    - `neutron_understack`
    - `understack-workflows`

    In future, we may consider consolidating all of these, but I was not sure if there won't be any conflicts so kept them separately for now.

- ~fixes `neutron_understack` tests~


stevekeay (approved):
LGTM

PR #258, opened 2024-09-09
feat: keystone authentication against dex

cardoe:


components/dex/values.yaml
```diff
@@ -49,6 +49,11 @@ config:
       name: "Undercloud ArgoCD"
       redirectURIs:
         - "https://argocd.$DNS_ZONE/auth/callback"
+    - id: keystone
+      secretEnv: KEYSTONE_SSO_CLIENT_SECRET
+      name: "OpenStack Keystone"
+      redirectURIs:
+        - "https://keystone-api.$DNS_ZONE/redirect_uri"
```
skrobul, line 56: `keystone-api` service does not listen on default port 443, so this should either be

`https://keystone-api.$DNS_ZONE:5000/redirect_uri`
or
`https://keystone.$DNS_ZONE/redirect_uri`

judging by the Apache configs below it should be the latter

skrobul:
Depends on https://github.com/RSS-Engineering/undercloud-deploy/pull/140

Closes PUC-498

components/dex/values.yaml
cardoe: Yes. you're right.

components/keystone/aio-values.yaml
```diff
@@ -33,6 +33,48 @@ bootstrap:
...
+                    "type": "REMOTE_USER"
+                },
+                {
+                    "type": "HTTP_OIDC_GROUPS"
+                }
+            ]
+        }
+    ]
+    EOF
+    openstack identity provider show sso 2>/dev/null || openstack identity provider create --remote-id https://dex.dev.undercloud.rackspace.net sso
```
cardoe, line 65: TODO: fix this.

cardoe (commented):
I can't +1 this cause I started it.

skrobul (approved):

PR #260, opened 2024-09-09
feat: operators: allow customization and add rook

skrobul:


apps/appsets/operators.yaml
```diff
@@ -14,26 +14,104 @@ spec:
...
+                      chart: rook-ceph-cluster
+                      targetRevision: v1.15.0
+                      helm:
+                        releaseName: rook-ceph-cluster
+                        valueFiles:
+                          - $understack/operators/rook/values-cluster.yaml
+                          - $deploy/helm-configs/{{.name}}/rook-cluster.yaml
+                        ignoreMissingValueFiles: true
+                - component: cnpg-system
+                  skipComponent: '{{has "rook" ((default "[]" (index .metadata.annotations "uc_skip_components") | fromJson))}}'
```
cardoe, line 49:
```suggestion
                  skipComponent: '{{has "cnpg-system" ((default "[]" (index .metadata.annotations "uc_skip_components") | fromJson))}}'
```

cardoe (changes requested):

apps/appsets/operators.yaml
skrobul: good catch, fixed in rebase thx

cardoe (approved):

PR #261, opened 2024-09-09
ci: fix bad container name from 5521afbd6

cardoe:


skrobul:
I don't  think this is going to work - passing the `container_name` value directly will result in double escaping because the octokit already escapes request parameters during JSON serialization:

https://github.com/octokit/request.js/blob/bec87c90d08449815ed8cac4f86eb3d082ed5f9c/src/fetch-wrapper.ts#L24

it should probably be just    

    const container_name = `understack/${this.containerName}`;`

PR #303, opened 2024-09-16
docs: make the install paths easier to use

cardoe:
Attempt to make the install paths easier to use by sticking to the deploy repo checkout.

### DIFF
--- a/docs/deploy-guide/gitops-install.md
+++ b/docs/deploy-guide/gitops-install.md
@@ -129,12 +129,9 @@ has been chosen; other tools like Vault, SOPS, etc should be considered
 for production deployments.
 
 ```bash
-# from your understack checkout
-./scripts/gitops-secrets-gen.sh ${UC_DEPLOY}/my-k3s.env
-pushd "${UC_DEPLOY}"
-git add my-k3s
-git commit -m "my-k3s: secrets generation"
-popd
+./scripts/gitops-secrets-gen.sh ./my-k3s.env
+git -C "${UC_DEPLOY}" add secrets/my-k3s
+git -C "${UC_DEPLOY}" commit -m "my-k3s: secrets generation"
 ```
 
 ### Defining the app deployment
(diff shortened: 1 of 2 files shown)


cardoe:
@skrobul any suggestions other than prefixing with that `cd` command each time?

skrobul (commented):
> @skrobul any suggestions other than prefixing with that cd command each time?

I think the previous version with pushd/popd was actually fine, it did not work during our deploy because `Saving the Environment Config` section used `cd` and did not have instruction to go back to the main understack repository.

When it other suggestions, the only other one which might be viable is something like this:

instead of saying:
```bash
./scripts/gitops-secrets-gen.sh ${UC_DEPLOY}/my-k3s.env
pushd "${UC_DEPLOY}"
git add secrets/my-k3s
git commit -m "my-k3s: secrets generation"
popd
```

say:
```bash
./scripts/gitops-secrets-gen.sh ${UC_DEPLOY}/my-k3s.env
(
  cd "${UC_DEPLOY}"
  git add secrets/my-k3s
  git commit -m "my-k3s: secrets generation"
)
```

In other words, `cd` inside the `()` subshell is scoped only to that subshell.

cardoe:
How about that change @skrobul ?

docs/deploy-guide/gitops-install.md
```diff
@@ -175,13 +168,15 @@ to your git server so that ArgoCD can access it.
...
-kubectl -n argocd apply -f "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/argocd/secret-*-cluster.yaml"
+cd /path/to/uc-deploy
+kubectl -n argocd apply -f "secrets/${DEPLOY_NAME}/argocd/secret-*-cluster.yaml"
 ```
 
 Now configure your ArgoCD to have the credential access to your deploy repo:
 
 ```bash
-kubectl -n argocd apply -f "${UC_DEPLOY}/secrets/${DEPLOY_NAME}/argocd/secret-deploy-repo.yaml"
+cd /path/to/uc-deploy
```
skrobul, line 178: not necessary, user has changed the directory already in a previous step

docs/deploy-guide/gitops-install.md
```diff
@@ -193,8 +188,7 @@ kubectl label node $(kubectl get nodes -o 'jsonpath={.items[*].metadata.name}')
 Finally run the following to have ArgoCD deploy the system:
 
 ```bash
-# from your understack checkout
-kubectl apply -f apps/aio-app-of-apps.yaml
+kubectl -n argocd apply -f https://raw.githubusercontent.com/rackerlabs/understack/main/apps/aio-app-of-apps.yaml
```
skrobul, line 191: I am conflicted about this one. It's good that it does not require user to pop back into understack's directory, but it's not so good because we cannott offer any guarantees that version at `main` branch is compatible with the version that user has checked out locally.

Do we have to change to the $UC_DEPLOY directory at all now that you solved the `git -C` thing?

skrobul (approved):

docs/deploy-guide/gitops-install.md
cardoe: That's a good point. I can update this as well.

skrobul (approved):

PR #309, opened 2024-09-17
feat: Neutron ML2 workflow calls nautobot job to set tenant networks

stevekeay:
We now have two separate operations:

1) to set to provisioning mode, the regular nautobot API is used to set the port status
2) to set to tenant mode, a special API call runs Milan's Nautobot job.  This will set the port status as well as creating the relevant VLAN(s).

### DIFF
--- a/python/understack-workflows/understack_workflows/main/undersync_device.py
+++ b/python/understack-workflows/understack_workflows/main/undersync_device.py
@@ -65,6 +65,9 @@ def argument_parser():
         "--device-id", type=UUID, required=False, help="Nautobot device UUID"
     )
     parser.add_argument("--network-name", required=True)
+    parser.add_argument(
+        "--network-id", type=UUID, required=True, help="Nautobot network UUID"
+    )
     parser = parser_nautobot_args(parser)
     parser.add_argument(
         "--force",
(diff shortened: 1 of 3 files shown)


python/understack-workflows/understack_workflows/main/undersync_device.py
```diff
@@ -23,20 +22,73 @@ def update_nautobot(args) -> UUID:
...
+    The nautobot job will assign vlans as required and set the interface
+    into the correct mode for "normal" tenant operation.
+
+    The switch ID is returned.
+    """
+
+    # Making this http request directly because it was not clear how to get
+    # the pynautobot api client to call an arbitrary endpoint:
+
+    uri = f"{nb_url}api/plugins/undercloud-vni/ucvni_jobs/prep_switch_interface"
```
mfencik, line 71: The default nb url does NOT contain ending "/":
https://github.com/rackerlabs/understack/blob/f024a0e7ae9b0e39b081f6878cbe43aa045b74aa/python/understack-workflows/understack_workflows/helpers.py#L44

mfencik (changes requested):

python/understack-workflows/understack_workflows/main/undersync_device.py
stevekeay: good catch :)

mfencik (approved):

PR #323, opened 2024-09-19
fix: Moving default token filename outside of class

mudakavip:


python/neutron-understack/neutron_understack/argo/workflows.py
```diff
@@ -6,25 +6,26 @@
...
 
-class ArgoClient:
-    DEFAULT_TOKEN_FILENAME = "/run/secrets/kubernetes.io/serviceaccount/token"
+DEFAULT_TOKEN_FILENAME = "/run/secrets/kubernetes.io/serviceaccount/token"
+
 
+class ArgoClient:
     def __init__(
         self,
         token: str = None,
```
cardoe, line 15:
```suggestion
        token: str | None = None,
```

cardoe (approved):

PR #325, opened 2024-09-20
feat: uc-iad3-dev - redeployment for bravo iteration

skrobul:
This PR includes various changes required to move us to a multi-node, full K8s cluster that replaced K3s.

Related: https://github.com/RSS-Engineering/undercloud-deploy/pull/148

Closes PUC-518

### DIFF
--- a/apps/appsets/components.yaml
+++ b/apps/appsets/components.yaml
@@ -163,16 +163,14 @@ spec:
                               value: '{{.name }}-cluster-issuer'
                 - component: argo-events
                   skipComponent: '{{has "argo-events" ((default "[]" (index .metadata.annotations "uc_skip_components") | fromJson))}}'
-                  source:
-                    repoURL: '{{index .metadata.annotations "uc_repo_git_url"}}'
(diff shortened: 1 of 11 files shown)


apps/appsets/components.yaml
```diff
@@ -163,16 +163,14 @@ spec:
...
-                    - group: ''
-                      kind: ServiceAccount
-                      name: workflow
-                      jsonPointers:
-                        - /imagePullSecrets
+                  sources:
+                    - repoURL: '{{index .metadata.annotations "uc_repo_git_url"}}'
+                      targetRevision: '{{index .metadata.annotations "uc_repo_ref"}}'
+                      path: 'components/argo-events'
+                    - repoURL: '{{index .metadata.annotations "uc_deploy_git_url"}}'
```
cardoe, line 170: Only concern on this one is that this path doesn't exist by default.

cardoe (approved):

apps/appsets/components.yaml
cardoe: That's the `secrets/$DEPLOY_NAME/argo-events` PATH.

PR #328, opened 2024-09-20
chore: migrate nodes to live in baremetal project

cardoe:
Move the project that our servers live in and get controlled in do a new domain called 'infra' with a project called 'baremetal'

### DIFF
--- a/components/keystone/aio-values.yaml
+++ b/components/keystone/aio-values.yaml
@@ -10,17 +10,17 @@ bootstrap:
           --user="${OS_USERNAME}" \
           --domain="${OS_DEFAULT_DOMAIN}" \
           "admin"
-    # create 'argoworkflow' user
-    # credentials for ironic-nautobot-sync and other argo workflows
-    openstack project create undercloud --or-show
-    openstack user create --project undercloud --password demo argoworkflow --or-show
-    openstack role add --user argoworkflow --project undercloud member
-    openstack role add --user argoworkflow --project undercloud admin
-    # allow ironic user to see servers in undercloud project
-    openstack role add --project undercloud --user ironic --user-domain service  member
+    # create 'infra' domain
+    openstack domain create --or-show infra
+    # create 'baremetal' project for our ironic nodes to live in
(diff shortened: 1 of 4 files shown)


components/keystone/aio-values.yaml
```diff
@@ -10,17 +10,17 @@ bootstrap:
...
-    # allow ironic user to see servers in undercloud project
-    openstack role add --project undercloud --user ironic --user-domain service  member
+    # create 'infra' domain
+    openstack domain create --or-show infra
+    # create 'baremetal' project for our ironic nodes to live in
+    openstack project create --or-show --domain infra baremetal
+    # create 'argoworkflow' user for automation
+    openstack user create --or-show --domain infra --password demo argoworkflow
+    # give 'argoworkflow' 'admin' over the 'baremetal' project
+    openstack role add --user-domain infra --project-domain infra --user argoworkflow --project baremetal admin
```
nicholaskuechler, line 20: I think we still need member role for horizon ui access

components/keystone/aio-values.yaml
```diff
@@ -10,17 +10,17 @@ bootstrap:
...
+
+    # this is too early because ironic won't exist
     openstack role add --project service --user ironic --user-domain service service
-    # add 'demo' user to have 'member' role, needed for horizon dashboard use
-    openstack role add --user demo --project undercloud member
```
nicholaskuechler, line 23: I think we still need member role for horizon ui access

components/keystone/aio-values.yaml
cardoe: Will the `argoworkflow` user be logging into the horizon UI? People should be logging in as themselves and not the service account.

components/keystone/aio-values.yaml
cardoe: Okay here.

cardoe: I've made `demo` be part of `ucuser`.

skrobul (approved):
let's see what breaks

PR #329, opened 2024-09-20
chore: bump mariadb operator to 0.26.0

cardoe:
The mariadb-operator had a breaking API change. This upgrades to the 0.26.0 CRDs so that we can move to the newer version. This has to be done in increments to handle all the upgrades.


nicholaskuechler (approved):

cardoe:
Nick's right. We'll not merge this until it's fully up to date.

PR #332, opened 2024-09-23
fix: back port allowing newer boot methods on iDRAC

cardoe:
With OpenStack Ironic 2024.2 we'll be able to use HTTP based boot methods. This back ports this to our 2024.1


cardoe:
Switched to 2024.2 so this is unblocked.

PR #342, opened 2024-09-24
fix: argo_python: revert change of owner_name lookup

skrobul:
The owner_name variable refers to a pod name, so that Secret can be automatically cleaned up by kubernetes when the Pod is removed.


stevekeay (approved):

cardoe (approved):
It's a revert to make it work but ultimately we want to improve on this. But this fixes what's broken so merge.

PR #343, opened 2024-09-25
fix: sync_interfaces: make the port names unique

skrobul:
Ironic's database schema requires the baremetal port names to be unique so we prepend the Node ID.


stevekeay (approved):

cardoe (approved):
Well that was unfortunate. Good solution.

PR #356, opened 2024-10-02
feat: adds openstack-exporter for prometheus

nicholaskuechler:


apps/appsets/components.yaml
```diff
@@ -188,6 +188,27 @@ spec:
...
+                      targetRevision: '{{index .metadata.annotations "uc_deploy_ref"}}'
+                      ref: secrets
+                      path: 'secrets/{{.name}}/openstack-exporter'
+                    - repoURL: registry.scs.community/openstack-exporter
+                      chart: prometheus-openstack-exporter
+                      targetRevision: 0.4.5
+                      helm:
+                        releaseName: prometheus-openstack-exporter
+                        valueFiles:
+                          - $understack/components/openstack-exporter/values.yaml
```
cardoe, line 209: If we don't need this file then just drop it.

apps/appsets/components.yaml
```diff
@@ -188,6 +188,27 @@ spec:
...
                       targetRevision: '{{index .metadata.annotations "uc_repo_ref"}}'
                       path: 'components/chrony'
+                - component: openstack-exporter
+                  componentNamespace: monitoring
+                  skipComponent: '{{has "openstack-exporter" ((default "[]" (index .metadata.annotations "uc_skip_components") | fromJson))}}'
+                  sources:
+                    - repoURL: '{{index .metadata.annotations "uc_repo_git_url"}}'
+                      targetRevision: '{{index .metadata.annotations "uc_repo_ref"}}'
+                      path: 'components/openstack-exporter'
+                      ref: understack
```
cardoe, line 198: Since we're not using anything from here just drop this section.

apps/appsets/components.yaml
```diff
@@ -188,6 +188,27 @@ spec:
...
+                    - repoURL: '{{index .metadata.annotations "uc_deploy_git_url"}}'
+                      targetRevision: '{{index .metadata.annotations "uc_deploy_ref"}}'
+                      ref: secrets
+                      path: 'secrets/{{.name}}/openstack-exporter'
```
cardoe, line 202: Is there actually any secrets needing to be applied? Or this is just the ref for the helm-config? Cause if its just the ref then drop the path.

cardoe (changes requested):

apps/appsets/components.yaml
nicholaskuechler: We'll need it in the future, but I've removed it for now.

apps/appsets/components.yaml
nicholaskuechler: We'll need it in the future, but I've removed it for now.

apps/appsets/components.yaml
nicholaskuechler: We'll need it in the future, but I've removed it for now.

cardoe (approved):

PR #371, opened 2024-10-09
feat: Password standard

stevekeay:
Historically we obtain DRAC passwords from CORE which has issues:

- in the future we don't want baremetal nodes in CORE/DCX
- requires manual deployment steps to assign locations in DCX and configure the physical DRAC
- prevents automation of the process because we have no way of knowing which arbitrary CORE device number belongs to each auto-discovered server

The changes here make DRAC passwords that are unique on each server, but without having to store each individual password in vault, CORE, etc.

The passwords are generated based on a cryptographic hash function of the IP address and a static master key.

Plan for Transition to new passwords:

1. back up the old passwords in a safe place
2. change the password on each existing baremetal node
3. update the passwords in CORE to reflect the new standard

Plan to release this code:

5. Make the master secret available to understack workflows https://github.com/RSS-Engineering/undercloud-deploy/pull/183
6. merge this PR and check the workflows are still functional

### DIFF
--- a/python/understack-workflows/understack_workflows/bmc_password_standard.py
+++ b/python/understack-workflows/understack_workflows/bmc_password_standard.py
@@ -1,0 +1,28 @@
+import ipaddress
+from base64 import b64encode
+from hashlib import sha256
+
+PASSWORD_LENGTH = 24
+
+
+def standard_password(ip_addr: str, master_key: str) -> str:
+    """Return a password string for the given IP address.
+
+    >>> standard_password("10.3.2.30", "ultra-secret string")
+    'DdushPf7IUTs6oya3ADS0OM3'
+    """
+    try:
+        ipaddress.ip_address(ip_addr)
+    except ValueError:
+        raise ValueError(f"Need an IPv4 address, not '{ip_addr}'") from None
+
+    if not master_key:
+        raise ValueError("Missing/empty master_key!")
+
+    long_password = _sha256_string(master_key + ip_addr)
+
+    return long_password[:PASSWORD_LENGTH]
+
+
+def _sha256_string(text: str) -> str:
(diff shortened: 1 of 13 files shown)


python/understack-workflows/understack_workflows/bmc_password_standard.py
```diff
@@ -0,0 +1,32 @@
...
+IPV4_REGEXP = r"^\d+\.\d+\.\d+\.\d+$"
+
+def standard_password(ip_addr: str, master_key: str) -> str:
+    """Return a password string for the given IP address
+
+    >>> standard_password("10.3.2.30", "ultra-secret string")
+    'DdushPf7IUTs6oya3ADS0OM3'
+    """
+    if not re.match(IPV4_REGEXP, ip_addr):
+        raise ValueError(f"Needed an IPv4 address, not '{ip_addr}'")
```
skrobul, line 15: python has built-in `ipaddress` module standard library since 3.4

```python
import ipaddress
try:
    ip = ipaddress.ip_address(ip_addr)
except ValueError:
    raise ValueError(f"Needed an IPv4 address, not '{ip_addr}'")
```

Might be better than regexp validation

skrobul (commented):
also `python/understack-workflows/tests/test_sync_bmc_creds.py` needs to be updated to mock the credential() call for master key

python/understack-workflows/understack_workflows/bmc_password_standard.py
stevekeay: Fixed in rebase.  I also corrected the tests.

stevekeay:
Having done all of this, I now feel like it would be less invasive to make this change on the undercloud-deploy side, so I am closing this in favour of https://github.com/RSS-Engineering/undercloud-deploy/pull/183.

It is possible that we will move forward with these changes at a later date once the dust settles.

PR #375, opened 2024-10-15
fix: skippable provisioning-ingress-nginx

skrobul:
These were the changes needed to get the staging environment running.

### DIFF
--- a/apps/appsets/infra.yaml
+++ b/apps/appsets/infra.yaml
@@ -64,6 +64,9 @@ spec:
                         valueFiles:
                           - $deploy/{{.name}}/helm-configs/ingress-nginx.yaml
                         ignoreMissingValueFiles: true
+                - component: provisioning-ingress-nginx
+                  skipComponent: '{{has "provisioning-ingress-nginx" ((default "[]" (index .metadata.annotations "uc_skip_components") | fromJson))}}'
+                  sources:
                     - repoURL: https://kubernetes.github.io/ingress-nginx
                       chart: ingress-nginx
                       targetRevision: 4.9.1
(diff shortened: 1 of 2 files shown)


apps/appsets/infra.yaml
```diff
@@ -64,6 +64,9 @@ spec:
                         valueFiles:
                           - $secrets/helm-configs/{{.name}}/ingress-nginx.yaml
                         ignoreMissingValueFiles: true
+                - component: provisioning-ingress-nginx
```
cardoe, line 67:
```suggestion
                - component: provisioning-ingress-nginx
                  componentNamespace: ingress-nginx
```

apps/appsets/infra.yaml
```diff
@@ -64,6 +64,9 @@ spec:
...
+                  skipComponent: '{{has "provisioning-ingress-nginx" ((default "[]" (index .metadata.annotations "uc_skip_components") | fromJson))}}'
+                  sources:
                     - repoURL: https://kubernetes.github.io/ingress-nginx
```
cardoe, line 70:
```suggestion
                    - repoURL: '{{index .metadata.annotations "uc_deploy_git_url"}}'
                      targetRevision: '{{index .metadata.annotations "uc_deploy_ref"}}'
                      ref: secrets
                    - repoURL: https://kubernetes.github.io/ingress-nginx
```

cardoe (changes requested):
So they'll fight over the namespace ownership so maybe they need to be in separate namespaces. But you'll need the references. In #377 I changed the reference from `secrets` to `deploy`.

cardoe (approved):
I'm still not sure what's going to happen with the namespace and if we need to figure out how to split the namespaces.

PR #385, opened 2024-10-17
feat: argocd: configure oauth2 client id for CLI usage

skrobul:
This is a public client so has to be separate from the web one.

Related:
https://github.com/RSS-Engineering/undercloud-deploy/pull/195


skrobul:
Note to the reviewer(s): I have tried `$argocd-sso:cli-client-id` but the secret interpolation does not work in this field

cardoe:
They've now added a whole config file interpolation with gomplate which I think we should ultimately switch to.

cardoe (approved):

PR #388, opened 2024-10-18
feat: adds snmp-exporter for prometheus

dsgnr:


apps/appsets/components.yaml
```diff
@@ -205,6 +205,24 @@ spec:
...
+                    - repoURL: '{{index .metadata.annotations "uc_deploy_git_url"}}'
+                      targetRevision: '{{index .metadata.annotations "uc_deploy_ref"}}'
+                      ref: deploy
+                    - repoURL: https://prometheus-community.github.io/helm-charts
+                      chart: prometheus-snmp-exporter
+                      targetRevision: 5.6.0
+                      helm:
+                        releaseName: prometheus-snmp-exporter
+                        valueFiles:
+                          - $deploy/{{.name}}/helm-configs/prometheus-snmp-exporter.yaml
```
cardoe, line 224: I'd ask that you flip this to the top under the `uc_repo` and `uc_deploy` as that's what we're using as our standard.

cardoe: https://rackerlabs.github.io/understack/deploy-guide/add-remove-app/

PR #389, opened 2024-10-18
feat: Add building ubuntu image

transt:
Add the directory keekz-ubuntu-extras under ironic-images.
Update .github/workflows/build-ironic-images.yaml to build ubuntu image.

Test run on a fork repo:
Screenshot 2024-10-18 at 10 03 49 AM


.github/workflows/build-ironic-images.yaml
```diff
@@ -15,35 +15,61 @@ on:
...
           ELEMENTS_PATH: ${{ env.pythonLocation }}/share/ironic-python-agent-builder/dib:${{ github.workspace }}/ironic-images/custom_elements
-      - name: Dynamically set timestamp environment variable
-        run: echo "TIMESTAMP=$(date +"%Y%m%d%H%M%S")" >> $GITHUB_ENV
-      - name: Publish IPA Release
+
+      - name: Get commit hash
+        run: echo "COMMIT_HASH=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+
+      - name: Set timestamp environment variable
+        run: echo "TIMESTAMP=$(git show --no-patch --no-notes --pretty='%cd' --date=format:'%Y%m%d%H%M%S' ${{ env.COMMIT_HASH }})" >>  $GITHUB_ENV
```
cardoe, line 62: You can combine these into 1 run cause you're not using the COMMIT_HASH variable anywhere else. Additionally you shouldn't have to get the revision this way since it's available from the `github` context as `github.sha` see: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs

If you're wanting to output information then you can look at: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions

.github/workflows/build-ironic-images.yaml
```diff
@@ -15,35 +15,61 @@ on:
...
+
+      - name: Publish Release
```
cardoe, line 64: Now that we're doing a matrix I think we'll have to split up the release into its own step. Because this will cause two release tags to be published. So I think each build will need to use the https://github.com/actions/upload-artifact to upload the built bits. Then the create release step will need to download those artifacts and make the release from that so that there's one release with all the files.

cardoe (changes requested):
This should just be generically called "ubuntu-$release" or some such. Definitely no keekz in the name and extras was there just cause that's the package pulled in.

.github/workflows/build-ironic-images.yaml
transt: I used timestamp of the commit sha so that it would be one release. The previous version generated a timestamp when the job run so with two jobs, you get two timestamp and 2 releases. The image that I added to the PR shows the one release and the three built files.

.github/workflows/build-ironic-images.yaml
transt: Thanks, I used the github.sha suggestion.

.github/workflows/build-ironic-images.yaml
cardoe: So the GitHub Actions matrix runner will run each one separately to completion. You are creating the release in the matrix so it'll make 2 releases on each run instead of one combined release.

cardoe: https://github.com/argoproj/argo-workflows/blob/main/.github/workflows/release.yaml that's an example of a matrix then coming back together to one release.

cardoe (approved):

PR #391, opened 2024-10-21
feat: add Crossplane and disable flavor bootstrap script (PUC-598)

cardoe:
Add Crossplane to be deployed and configured with the OpenStack provider so that we can use that to create OpenStack resources from Crossplane. Configured Crossplane with the Keystone admin user. Added some information on how to define a flavor in nova using Crossplane and disabled the nova bootstrap script creating flavors.


cardoe:
Eventually the flavors here could be crafted by the data in #396

PR #401, opened 2024-10-24
chore: switch openstack public endpoints to https

skrobul:


components/nova/aio-values.yaml
```diff
@@ -18,6 +18,12 @@ endpoints:
       name: rabbitmq-server
     hosts:
       default: rabbitmq-nodes
+  compute:
+    port:
+      api:
```
cardoe, line 23:
```suggestion
      osapi:
```

I think nova uses that.

components/neutron/aio-values.yaml
```diff
@@ -11,6 +11,13 @@ endpoints:
       name: rabbitmq-server
     hosts:
       default: rabbitmq-nodes
+  network:
+    port:
+      api:
```
cardoe, line 16:
```suggestion
      server:
```

I think neutron uses this

cardoe (commented):
I think you need these two.

components/neutron/aio-values.yaml
skrobul: hmm where did you see that?
[The openstack-helm values.yaml for neutron](https://opendev.org/openstack/openstack-helm/src/commit/8e5ad85fd09983e07ddc87de76e480b4ceca9120/neutron/values.yaml#L2535) seems to have it as 'api'

components/nova/aio-values.yaml
skrobul: https://opendev.org/openstack/openstack-helm/src/commit/8e5ad85fd09983e07ddc87de76e480b4ceca9120/nova/values.yaml#L1880

cardoe (approved):

PR #414, opened 2024-10-25
chore: renovate rule to watch OpenStack Helm

cardoe:
Added a renovate bot rule to watch OpenStack Helm

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -10,5 +10,24 @@
     ":prNotPending",
     ":rebaseStalePrs",
     ":automergeDigest"
+  ],
+  "customDatasources": {
+    "openstackhelm": {
+      "defaultRegistryUrlTemplate": "https://tarballs.opendev.org/openstack/openstack-helm/",
+      "format": "html",
+      "transformTemplates": [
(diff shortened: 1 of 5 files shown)


.github/renovate.json
```diff
@@ -10,5 +10,24 @@
     ":prNotPending",
     ":rebaseStalePrs",
     ":automergeDigest"
+  ],
+  "customDatasources": {
+    "openstackhelm": {
+      "defaultRegistryUrlTemplate": "https://tarballs.opendev.org/openstack/openstack-helm/",
+      "format": "html",
+      "transformTemplates": [
+        "{\"releases\": $map($filter($.releases, function($v) {$match($v.version, /^{{packageName}}-/)}), function($v) { {\"version\": $match($v.version, /^{{packageName}}-([0-9.]+).tar.gz$/).groups[0]} } ) }"
```
skrobul, line 19:
```suggestion
        "{\"releases\": $map($filter($.releases, function($v) {$match($v.version, /^{{packageName}}-/)}), function($v) { {\"version\": $match($v.version, /^{{packageName}}-([0-9.]+)\\.tar\\.gz$/).groups[0]} } ) }"
 ```

We may want to escape the literal dots here, but I bet the original version still works so nbd

skrobul (approved):

.github/renovate.json
cardoe: Yeah I dunno. I used the recommended website to craft the rule and then escaped it until the checker was happy. Not sure if it'll do the right thing yet.

cardoe: What should we do with the PR?

PR #370, opened 2024-10-09
feat: Ingest server

stevekeay:
This is replacement server ingestion code.  The idea is to "enroll" a server via DHCP, so it is the DHCP server that will trigger this workflow whenever a new device appears.

This will replace some existing workflows as it takes care of the ingestion piece in one go.

The device that gets created in Nautobot looks like the below - it has all the info we are accustomed to, except for "U" height and CORE number.  The hostname format also looks different.

image

image

Sample logs from running this thing:
```
2024-10-09 22:20:10,198 - __main__ - INFO - /home/steve.keay/p/understack/python/understack-workflows/understack_workflows/main/ingest_or_refresh_baremetal_node.py starting for bmc_ip_address='10.46.96.156'
2024-10-09 22:20:10,200 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): nautobot.dev.undercloud.rackspace.net:443
2024-10-09 22:20:10,512 - urllib3.connectionpool - DEBUG - https://nautobot.dev.undercloud.rackspace.net:443 "GET /api/ HTTP/1.1" 200 839
2024-10-09 22:20:10,520 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 10.46.96.156:443
2024-10-09 22:20:11,648 - urllib3.connectionpool - DEBUG - https://10.46.96.156:443 "GET /redfish/v1/Systems/System.Embedded.1/ HTTP/1.1" 200 2757
2024-10-09 22:20:11,659 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 10.46.96.156:443
2024-10-09 22:20:12,364 - urllib3.connectionpool - DEBUG - https://10.46.96.156:443 "GET /redfish/v1/Managers/iDRAC.Embedded.1/EthernetInterfaces/NIC.1 HTTP/1.1" 200 625
2024-10-09 22:20:12,370 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 10.46.96.156:443
2024-10-09 22:20:13,081 - urllib3.connectionpool - DEBUG - https://10.46.96.156:443 "GET /redfish/v1/Systems/System.Embedded.1/EthernetInterfaces/ HTTP/1.1" 200 253
2024-10-09 22:20:13,089 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 10.46.96.156:443
[...]

PR #415, opened 2024-10-25
feat(workflows): enroll-server moves it to available state

cardoe:
Make the enroll-server script return the device UUID so we can use it for future steps. Then add the process to move the node from enroll to manage and then finally to available.


cardoe:
So I know #370 is new server ingestion. I kind of envisioned using the OpenStack CLI as much as possible in steps rather than writing code to call each API endpoint so using this as an example of that.

stevekeay (approved):

PR #421, opened 2024-10-29
feat(openstack-admin-client): adds useful troubleshooting tools

nicholaskuechler:
Adds a couple useful troubleshooting tools to the openstack-admin-client.


cardoe (commented):
So I'm a -1 on this as I don't think the "admin-client" container should even be running in the cluster nor should it just be some generic environment to login to. This is for scripted operations.

PR #424, opened 2024-10-30
chore: Add mypy policing of types to CI

stevekeay:
We have two options - mypy or pyright.  This PR explores the mypy option.  #429 explores the other.


stevekeay:
closing this out because doug chose pyright

PR #430, opened 2024-11-01
chore: Refactor workflow python code

stevekeay:
Some of this helps to satisfy the type checker, but mostly it is aesthetic.


python/understack-workflows/tests/test_topology.py
```diff
@@ -1,29 +1,58 @@
...
-
-def read_json_samples(file_path):
-    here = pathlib.Path(__file__).parent
-    ref = here.joinpath(file_path)
-    with ref.open("r") as f:
-        return json.loads(f.read())
-
-
-NAUTOBOT_DATA = read_json_samples("json_samples/nautobot_device_data.json")
+NAUTOBOT_DATA = NautobotDevice(
```
cardoe, line 6: I see the same device above. Can we not share this and put it in a fixture so that it's not open coded?

stevekeay: Done in 1f3cc8b

cardoe (changes requested):
Instead of:

```
from fixture_nautobot_device import FIXTURE_DELL_NAUTOBOT_DEVICE
```

in each of the files. import that into `tests/conftest.py` and make a function like:

```
@pytest.fixture
def dell_nautobot_device() -> NautobotDevice:
    return FIXTURE_DELL_NAUTOBOT_DEVICE
```

Then for `def test_switch_connections():` change it to `def test_switch_connections(dell_nautobot_device: NautobotDevice):` and use that.

cardoe (approved):

PR #434, opened 2024-11-04
feat: Create vni from neutron

mfencik:


components/neutron/aio-values.yaml
```diff
@@ -48,7 +48,8 @@ conf:
       ml2:
         # set the default ml2 backend to our plugin, neutron_understack
         mechanism_drivers: understack
-        tenant_network_types: local
+        tenant_network_types: "vxlan,local"
+        type_drivers: "flat,vlan,local,understack_vxlan"
```
cardoe, line 52: Don't want flat here.

PR #443, opened 2024-11-05
fix: NIC matching on server enroll

skrobul:
Before:

    >>> interface_is_relevant("NIC.Slot.2-1-1")
    False
    >>> interface_is_relevant("NIC.Slot.2-1")
    True

After:

    >>> interface_is_relevant("NIC.Slot.2-1-1")
    True
    >>> interface_is_relevant("NIC.Slot.2-1")
    True
    >>>

Closes PUC-658

### DIFF
--- a/python/understack-workflows/understack_workflows/bmc_chassis_info.py
+++ b/python/understack-workflows/understack_workflows/bmc_chassis_info.py
@@ -211,7 +211,9 @@ def parse_lldp_port(port_data: dict[str, str]) -> dict:
 
 
 def interface_is_relevant(url: str) -> bool:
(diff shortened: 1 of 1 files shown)


cardoe:
The issue here is these machines have the broken firmware on them. These interfaces will change once the iDRAC version is bumped.

cardoe (commented):
The issue here is you're ingesting servers with iDRAC firmware that we would want to blacklist. It's broken on these NICs. And shouldn't be used. The naming will change to what's expected once we update them.

skrobul:
Fair enough, thanks for clarification - closing

PR #450, opened 2024-11-07
feat: adds github workflow to create an ironic esp image

nicholaskuechler:
https://docs.openstack.org/ironic/latest/install/configure-esp.html
https://docs.openstack.org/ironic/latest/install/configure-pxe.html#uefi-pxe-grub

Screenshot 2024-11-07 at 3 13 24 PM

### DIFF
--- a/.github/workflows/build-ironic-images.yaml
+++ b/.github/workflows/build-ironic-images.yaml
@@ -14,6 +14,10 @@ on:
   merge_group:
     types: [checks_requested]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   discover:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


cardoe (changes requested):
Add this to the regular images workflow so it's all released together.

.github/workflows/build-ironic-images.yaml
```diff
@@ -86,3 +137,13 @@ jobs:
...
           files: ironic-images/upload/*
         if: ${{ github.ref == 'refs/heads/main' }}
+
+  cleanup:
+    runs-on: ubuntu-latest
+    needs: build-tenant-images
+
+    steps:
+      - name: Clean up artifacts
+        uses: geekyeggo/delete-artifact@v5
```
cardoe, line 147: Is this just so we don't have a ton of artifacts?

nicholaskuechler: Yep exactly. I think the default artifact lifetime is 90 days.

cardoe: We can set the retention time to 1 day but I won't block this going in.

cardoe (approved):
So this proves what I was saying on the original change here that the release needs to be in a final step as one upload.

PR #452, opened 2024-11-07
ci: split up code coverage reporting to support forks

cardoe:
Make it possible to support forks by splitting the code coverage reporting into two steps. Forks cannot make changes against the main repo so they cannot update the ticket so it must be split into two parts. Simplify the reporting as well by merging it together into one coverage file.


mfencik (approved):
LGTM

PR #454, opened 2024-11-08
feat: PUC-598: create or set baremetal flavor

haseebsyed12:
`argo template create baremetal-flavor.yaml`
`argo submit -n argo-events --from workflowtemplate/baremetal-flavor -p flavor_name="gp4.small" -p flavor_properties="--property resources:CUSTOM_BAREMETAL_GP4SMALL=1 --property resources:VCPU=2 --property resources:MEMORY_MB=4096 --property resources:DISK_GB=20"`

[workflows-executed](https://workflows.dev.undercloud.rackspace.net/workflows/argo-events/baremetal-flavor-sdzmm?tab=workflow&nodeId=baremetal-flavor-sdzmm-838540840&uid=065328a3-8452-43c1-82bc-ef5d5d5cf06d)

### DIFF
--- a/workflows/argo-events/workflowtemplates/baremetal-flavor.yaml
+++ b/workflows/argo-events/workflowtemplates/baremetal-flavor.yaml
@@ -1,0 +1,68 @@
+---
+apiVersion: argoproj.io/v1alpha1
+metadata:
+  name: baremetal-flavor
+  annotations:
+    workflows.argoproj.io/title: Set or Create flavors for use with the Bare Metal service
+    workflows.argoproj.io/description: |
+      Defined in `workflows/argo-events/workflowtemplates/baremetal-flavor.yaml`
+      Set or create a flavor with specified properties if it does not already exist.
+kind: WorkflowTemplate
+spec:
+  serviceAccountName: workflow
+  entrypoint: main
+  arguments:
+    parameters:
+      - name: flavors
+        value:
+          '[{"name": "gp2.small", "properties": "--property resources:CUSTOM_BAREMETAL_GP2SMALL=1 --property resources:VCPU=0 --property resources:MEMORY_MB=0 --property resources:DISK_GB=0"},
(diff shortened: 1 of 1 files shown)


workflows/argo-events/workflowtemplates/baremetal-flavor.yaml
```diff
@@ -0,0 +1,133 @@
+---
+apiVersion: argoproj.io/v1alpha1
+metadata:
+  name: baremetal-flavor
+  namespace: argo-events
```
cardoe, line 5:
```suggestion
```

remove the explicit namespace to allow us to test in other namespaces.

workflows/argo-events/workflowtemplates/baremetal-flavor.yaml
```diff
@@ -0,0 +1,133 @@
...
+      inputs:
+        parameters:
+          - name: flavor_properties
+      container:
+        image: ghcr.io/rackerlabs/understack/openstack-client:2024.2-ubuntu_jammy
+        command:
+          - sh
+          - -c
+          - |
+            eval "openstack flavor create '{{workflow.parameters.flavor_name}}' {{inputs.parameters.flavor_properties}}"
```
cardoe, line 98: Not understanding why we need the eval?  Cause the parameters are passed as 1 blob?

https://argo-workflows.readthedocs.io/en/latest/walk-through/scripts-and-results/ might be worth using the script syntax instead of the command syntax then.

cardoe (commented):
See https://argo-workflows.readthedocs.io/en/latest/walk-through/scripts-and-results/ and consider if the `script:` syntax would work better?

cardoe (commented):
So this is a good first step. Ultimately need to walk over all the flavors in a defined repo and execute this.

cardoe (changes requested):
Again. It's a step but no where addresses the needs asked. There's going to be flavors and we need to parse them and ensure they are applied. This workflow assumes someone will manually execute this.

PR #457, opened 2024-11-08
chore: update renovate configs for reuse and enable some auto merge

cardoe:
Split up the renovate configs so we can reuse them in some other repos. Enabled auto merging of GitHub Actions minor and patch updates.


mfencik (approved):
LGTM

PR #461, opened 2024-11-11
feat: Have Enroll process change the idrac hostname to match nautobot

stevekeay:


python/understack-workflows/understack_workflows/bmc_hostname.py
```diff
@@ -0,0 +1,20 @@
+from understack_workflows.bmc import Bmc
+from understack_workflows.helpers import setup_logger
+
+REDFISH_PATH = "/redfish/v1/Managers/iDRAC.Embedded.1/EthernetInterfaces/NIC.1"
```
cardoe, line 4: So overall this is hardcoded to Dell's. I really believe we need to use the sushy library here to be more generic.

cardoe: This isn't an ask to rewrite at the moment.

cardoe (approved):

PR #466, opened 2024-11-12
chore: Change the value that Enroll workflow sets in Ironic local_link_info

stevekeay:
Having the correct data to hand will simplify later nautobot jobs.


skrobul (approved):
LGTM.

In relation to our call earlier - I checked and Ironic has some limitations as to what we can put into local link information, but [it's not very restrictive](https://opendev.org/openstack/ironic/src/commit/feb774000923a9ae90dee7167ff6b816ce34c570/ironic/api/controllers/v1/utils.py#L98) and this change should work just fine

PR #472, opened 2024-11-13
feat: stop waiting for all interfaces to be online in IPA

skrobul:
This changes behavior of `systemd-networkd-wait-online` so that it does not wait the default 120s for all of interfaces to become online, which in reality never comes effectively delaying boot for 2 minutes.

Instead, we are satisfied with just one interface reaching the IPv4 routable state.

Closes PUC-580

### DIFF
--- a/ironic-images/ipa-debian-bookworm.yaml
+++ b/ironic-images/ipa-debian-bookworm.yaml
@@ -6,3 +6,4 @@
   - journal-to-console
   - package-installs
   - undercloud-ipa
+  - install-static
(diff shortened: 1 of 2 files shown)


cardoe (approved):

skrobul:
for the record - this has not worked:

```
# ls -la /etc/systemd/system/systemd-networkd-wait-online.service.d/
ls: cannot access '/etc/systemd/system/systemd-networkd-wait-online.service.d/': No such file or directory
root@debian:/etc/systemd/system# 
```

PR #474, opened 2024-11-13
feat: set raid config during ironic baremetal node enroll

nicholaskuechler:


stevekeay (approved):
LGTM, super long names

PR #478, opened 2024-11-13
fix: updates argo to mermaid script for new workflow directory structure and layout

nicholaskuechler:
Screenshot 2024-11-13 at 4 07 11 PM

### DIFF
--- a/scripts/argo-workflows-to-mkdocs.py
+++ b/scripts/argo-workflows-to-mkdocs.py
@@ -101,6 +101,7 @@ def main_orig():
 def main():
     if len(sys.argv) < 3:
         print("Usage: python script.py argo-workflows-input-dir mkdocs-output-dir")
+        print("The script will scan for 'workflowtemplates' dir in subdirs")
         sys.exit(1)
 
     input_dir = sys.argv[1]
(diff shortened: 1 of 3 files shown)


scripts/argo-workflows-to-mkdocs.py
```diff
@@ -616,7 +641,14 @@ def generate_mermaid(workflows, nodes, output_name, output_file, workflow_readme
     with open(output_file, "w") as f:
         f.write(f"# {output_name}\n")
         f.write("\n")
-        f.write("# Workflow Diagram\n")
+
+        if workflow_readme:
+            f.write(workflow_readme)
+        else:
+            f.write("This workflow does not have a `docs/README.md` file! :(\n")
```
cardoe, line 648: Where does this path come from?

nicholaskuechler: I updated the script to give a little more information when you run it: https://github.com/rackerlabs/understack/pull/478/files#diff-4c8cbfaab0abae3c33c48c1101c1a2f38f080540424e761c7756fbecccebea9fR103-R104 but you run the script giving it the workflows directory, then it scans subdirs for `workflowtemplates` directory and parses the workflowtemplates. 

What you might be interested in is this bit: https://github.com/rackerlabs/understack/pull/478/files#diff-4c8cbfaab0abae3c33c48c1101c1a2f38f080540424e761c7756fbecccebea9fR137-R150 - it's grabbing the workflow's title and description from the yaml but setting a message if they're missing. Here's an example of what the keystone workflows shows with the 'title not set' message:

Screenshot 2024-11-13 at 5 42 45 PM

cardoe (approved):

PR #484, opened 2024-11-15
feat: lint workflow templates

transt:
Add git workflow to argo lint understack workflow templates.

### DIFF
--- a/.github/workflows/lint-workflow-templates.yaml
+++ b/.github/workflows/lint-workflow-templates.yaml
@@ -1,0 +1,58 @@
+name: Lint Workflow Templates
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  lint-workflow-templates:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout latest changes from the PR branch
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+
+      - name: Set up Kind cluster with Helm
+        uses: helm/kind-action@v1.10.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-workflow-templates.yaml
```diff
@@ -0,0 +1,42 @@
...
+          chmod +x argo-linux-amd64
+          sudo mv argo-linux-amd64 /usr/local/bin/argo
+          argo version
+
+      - name: Argo Lint Workflow Template
+        run: |
+          for file in workflows/argo-events/workflowtemplates/*.yaml; do
+              echo "Linting workflow template: $file"
+              argo lint $file
+          done
```
cardoe, line 42:
```suggestion
        run: |
          ret=0
          for file in workflows/argo-events/workflowtemplates/*.yaml; do
              echo "Linting workflow template: $file"
              argo lint $file
              if [ $? -ne 0 ]; then
                  echo "::error file=$file,title=argo lint failed"
                  ret=1
              fi
          done
          exit $ret
```

Does `argo lint` return a non-zero exit code? Should we bubble up some more info?

see this doc for context: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message

cardoe (commented):
Just wondering how we'll make the workflow fail and report the error.

.github/workflows/lint-workflow-templates.yaml
transt: Updated

cardoe: Nice!

cardoe (approved):

PR #487, opened 2024-11-15
fix: fixed up ironic state workflow and document it and keystone workflow

cardoe:
Fixed up the ironic sensor and workflow to actually run on events while documenting it better. Added some documentation to the keystone one as well and sync its configuration.


cardoe:
image

skrobul (approved):

PR #491, opened 2024-11-18
fix: fixes casing for UnderStack in docs

nicholaskuechler:
I also noticed we were pinning the mkdocs github actions to very old versions from March 2024. I've unpinned them.

### DIFF
--- a/docs/component-argo-workflows.md
+++ b/docs/component-argo-workflows.md
@@ -1,7 +1,7 @@
 # Argo Workflows
 
 [Argo Workflows][argo-workflows] was chosen as our workflow / orchestration service. It runs natively inside a
-Kubernetes environment and easily integrates with [Argo Events][argo-events]. Understack will be deployed with the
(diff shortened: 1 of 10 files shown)


skrobul (approved):

cardoe (changes requested):
@skrobul I enabled the container pinning after you had asked for that. If you want to move the version up, you need to switch to like `v4.3.0` instead of `v4`. The bot will pin stuff right back.

nicholaskuechler:
> @skrobul I enabled the container pinning after you had asked for that. If you want to move the version up, you need to switch to like `v4.3.0` instead of `v4`. The bot will pin stuff right back.

I've already removed unpinning them.

skrobul:
> The bot will pin stuff right back.

I was hoping it would  pin stuff back to whatever commit the `v4` points to today, is that not how it works?

cardoe (changes requested):
I'll remove my block.

cardoe (approved):
Apparently I can only approve to remove the blocker.

PR #494, opened 2024-11-19
PUC-648: adding new custom node fields to nautobot

haseebsyed12:


cardoe (commented):
Overall this is spot on. The only feedback I would say is change the `lessee` from a custom field to updating the `Tenant` on the `Device` object.

python/understack-workflows/understack_workflows/nautobot.py
```diff
@@ -48,13 +48,19 @@ def non_lag_interface_by_mac(
             )
         return interface  # type: ignore
 
-    def update_cf(self, device_id: UUID, field_name: str, field_value: str):
+    def update_cf(
+        self, device_id: UUID, tenant_id: UUID = None, fields: dict[str, str] = None
```
cardoe, line 52:
```suggestion
        self, device_id: UUID, tenant_id: UUID | None = None, fields: dict[str, str]
```

cardoe (changes requested):
Interesting that pyright didn't catch this issue because the type wasn't correct.

python/understack-workflows/understack_workflows/nautobot.py
```diff
@@ -48,13 +48,19 @@ def non_lag_interface_by_mac(
...
         if not device:
             raise Exception(f"No such device {device_id}")
         if not device.custom_fields:
             raise Exception(f"Device {device_id} has no custom fields")
-        device.custom_fields[field_name] = field_value
+
+        for field_name, field_value in fields.items():
+            device.custom_fields[field_name] = field_value
+
+        device.tenant = tenant_id
```
cardoe, line 63:
```suggestion
        device.tenant = tenant_id  # type: ignore[attr-defined]
```

cardoe (changes requested):

PR #502, opened 2024-11-20
fix: Switch from MAC address stored in Nautobot asset_tag to custom field

stevekeay:
The custom fields have been added to ansible, and to the dev and staging environments, and in those environments the data has been copied from the asset_tag field to the new custom field.

We had to change how we do the graphql filter - you could search in Nautobot's built-in string attributes by passing an array of possible values to find, but that doesn't work for custom fields.

### DIFF
--- a/python/understack-workflows/understack_workflows/nautobot_device.py
+++ b/python/understack-workflows/understack_workflows/nautobot_device.py
@@ -1,4 +1,3 @@
-import json
 import re
 from dataclasses import dataclass
 from ipaddress import IPv4Interface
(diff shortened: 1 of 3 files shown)


python/understack-workflows/understack_workflows/nautobot_device.py
```diff
@@ -139,26 +138,26 @@ def switches_for(nautobot, chassis_info: ChassisInfo) -> dict:
...
+        devices(cf_chassis_mac_address__re: "(%s)"){
             id name
-            mac: asset_tag
+            mac: cf_chassis_mac_address
             location { id name }
             rack { id name }
         }
     }"""
-        % formatted_list
+        % pattern
```
skrobul, line 160: string interpolation creates the GraphQL injection vulnerability. For example, one could say

```python
mac_addresses='{"malicious_input")} { devices { id name } } #"}'
```

Consider replacing this with query variables (untested):
```python
query = """
query($pattern: String!) {
    devices(cf_chassis_mac_address__re: $pattern) {
        id name
        mac: cf_chassis_mac_address
        location { id name }
        rack { id name }
    }
}
"""

variables = {"pattern": "|".join(mac_addresses)}
result = nautobot.graphql.query(query, variables=variables)
```

skrobul (approved):
minor suggestion

python/understack-workflows/understack_workflows/nautobot_device.py
```diff
@@ -139,26 +138,26 @@ def switches_for(nautobot, chassis_info: ChassisInfo) -> dict:
...
-
     returns a dict[mac_address] -> dict switch information indexed by mac
     """
-    formatted_list = json.dumps(list(mac_addresses))
+    pattern = "|".join(mac_addresses)
 
     query = (
         """{
-        devices(asset_tag: %s){
+        devices(cf_chassis_mac_address__re: "(%s)"){
```
cardoe, line 153: Do we have to do a regex? isn't that going to be a slow query?

stevekeay: We had to change how we do the graphql filter - you could search in Nautobot's built-in string attributes by passing an array of possible values to find, but that doesn't work for custom fields.   Regexp does work, however.

I assumed that this is somehow translated to SQL and executed by the database but I didn't prove that.  The performance is good on our staging system, and certainly much faster than performing multiple individual queries.  Not sure how else this can be done.

stevekeay: Well, I did have the idea of moving a lot of this code to the Nautobot plugin, but that doesn't quite resonate with me for some reason.

cardoe: ah okay. I was just asking about the regex. I'm a +1 on less plugin. I'm hoping we can get to a clean install at some point.

stevekeay:
I updated all the queries to avoid suspicion due to string interpolation.

PR #506, opened 2024-11-20
fix: ironic-conductor fails to start with agent inspect interface is enabled

nicholaskuechler:
Today I found one of our environment had an ironic-conductor failing to start. I'm not sure why, but the agent inspection interface caused ironic to crash without a stack trace. After troubleshooting, I ended up getting ironic to work by removing the agent interface from the list of enabled inspection interfaces.


skrobul (changes requested):
Hold off on merging that one - I think I might have broken it. The Ironic hooks are configured as:
`hooks: "$default_hooks,resource-class"`

but the `ironic_understack` module is not installed in the image anymore:

```bash
root@1327175-hp3:/# pip list 2>/dev/null | grep understack
understack-flavor-matcher 0.0.0
root@1327175-hp3:/#
```

Working on fix now

PR #509, opened 2024-11-21
fix: ironic image builds without the ironic_understack module

skrobul:
The original intention was to copy both modules into /tmp/ironic-understack/ but the trailing / was missed. Instead, make it more explicit to ensure that both are installed.

Closes #506


mfencik (approved):
LGTM

PR #511, opened 2024-11-21
fix: fix flavor to ironic resource class conversion

nicholaskuechler:
Resource class is coming in to ironic as:
```
| resource_class         | baremetal.gp2.small                                                                                                                                                                 |
```
But it should be `baremetal.gp2small`


mfencik (approved):
LGTM

PR #519, opened 2024-11-25
feat: add logging for original port context

mfencik:


python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -230,6 +232,23 @@ def delete_port_precommit(self, context):
     def delete_port_postcommit(self, context):
         log_call("delete_port_postcommit", context)
 
+        network_id = context.current["network_id"]
+
+        if network_id == cfg.CONF.ml2_type_understack.provisioning_network:
+            connected_interface_uuid = self.fetch_connected_interface_uuid(
+                context.current
+            )
+            port_status = "Active"
```
skrobul, line 241: this variable can be inlined

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -230,6 +232,23 @@ def delete_port_precommit(self, context):
...
+            configure_port_status_data = self.nb.configure_port_status(
+                connected_interface_uuid, port_status
+            )
+            switch_uuid = configure_port_status_data.get("device", {}).get("id")
+            nb_vlan_group_id = UUID(self.nb.fetch_vlan_group_uuid(switch_uuid))
+            self.undersync.sync_devices(
+                vlan_group_uuids=str(nb_vlan_group_id),
+                dry_run=cfg.CONF.ml2_understack.undersync_dry_run,
+            )
```
skrobul, line 250: do we want some debug logs after this?

skrobul (approved):

python/neutron-understack/neutron_understack/neutron_understack_mech.py
mfencik: true, but in case this will change for something longer, i like to have it like that

python/neutron-understack/neutron_understack/neutron_understack_mech.py
mfencik: as discussed, we need to review the logging in general

PR #523, opened 2024-11-26
feat: Detect and Avoid Duplicate DRAC IP address on server Enroll

stevekeay:
This improves diagnostic messages during various enroll failure scenarios, to provide a more detailed error message about what went wrong.

It also detects two extra failure cases when DRAC IP is already in use on another device, and when DRAC interface already has an IP address.  This is designed to prevent any "duplicate" IP assignments in Nautobot.

closes PUC-690


skrobul (approved):
LGTM. Please fixup [fc8368e](https://github.com/rackerlabs/understack/pull/523/commits/fc8368e3acc6d9c2cd6f468d4c9f451f9bbb8c19) if possible

PR #531, opened 2024-12-03
feat: Helm for DB config, MariaDB configurable storage

skrobul:
> [!CAUTION]
> Do note merge without reading the section below!

### Description
This PR changes how the MariaDB component for OpenStack is configured.

By moving to a dedicated Helm chart, we allow better customisation of options delivered to the [mariadb-operator](https://github.com/mariadb-operator/mariadb-operator).
The main reason behind this change was to have the ability to change the `storageClass` used to back the database PVCs, as we previously relied on a default storage class being configured correctly.

Please review, but **DO NOT MERGE** because ArgoCD's autosync **may wipe** the database and other resources if the changes are not applied in correct order.
I am planning to integrate this change as outlined in the plan below (please provide feedback on it too if you have any):

Depends on https://github.com/RSS-Engineering/undercloud-deploy/pull/239
Closes PUC-521

---

### Migration Plan to switchover clusters to UnderstackDB

- merge the PR for deploy repo
- make sure that both dev and cluster follow `HEAD`
- create a new branch (`migration-X`)that only disables `autosync` for `operators` and `openstack` components.
- set **BOTH** clusters to follow the `migration-X` branch
- verify that autosync is disabled for both `openstack` and `operators` appsets through CLI or UI, refresh as necessary
- merge the understack PR into the `migration-X` branch
  - refresh the `bravo-uc-iad3-understackdb` app, review the changes. This should mostly be annotation and name changes like these:
    ```
    ❯ argocd app diff argocd/bravo-uc-iad3-dev-understackdb 
    
    ===== k8s.mariadb.com/Database openstack/glance ======
    8c8
    <     argocd.argoproj.io/instance: bravo-uc-iad3-dev-glance
    ---
    >     argocd.argoproj.io/instance: bravo-uc-iad3-dev-understackdb
    53a54
    >   name: glance
    
    ===== k8s.mariadb.com/Database openstack/horizon ======
    8c8
    <     argocd.argoproj.io/instance: bravo-uc-iad3-dev-horizon
    ---
    >     argocd.argoproj.io/instance: bravo-uc-iad3-dev-understackdb
    53a54
    >   name: horizon
    
      [...]
    ```
- Sync the `bravo-uc-iad3-understackdb` if you are happy with those changes.
- Switchover to the openstack appset and verify that the `User`,`Database`,`Grant` resources have moved away from their relevant apps.
- Repeat the same for `charlie/staging` cluster.
- Verify that everything is in sync and there is no outstanding diffs
- merge the PR to `main` branch.
- Switchover the ArgoCD instances to follow `HEAD/HEAD` or their equivalent refs again.


cardoe:
So I don't understand why the Database, User and Grant objects are being moved. Cause certain pieces will be deployed in global, regional, and fabric levels. With this setup you're bringing everything to one scope which works for AIO but not in a large scale.

I thought we would put the pieces in `components/openstack` into a helm chart and that's it.

skrobul:
> So I don't understand why the Database, User and Grant objects are being moved. 

Few reasons:
1. Mostly because of [this comment](https://rackspace.atlassian.net/browse/PUC-521?focusedCommentId=3234939), if that's not what you meant, can you please elaborate on what is the actual requirement here?

2. This removes clearly repetitive, error prone, hardcoded yaml which can now be generated by providing only what matters (the metadata).
3. Logically it makes sense to me keep the database configuration [in a single place](https://github.com/rackerlabs/understack/blob/mariadb-configurable-storage/helm-charts/understackdb/values.yaml) rather than spread over several files. It is also more consistent and now takes total of 51 lines instead of 376 lines. 

>  Cause certain pieces will be deployed in global, regional, and fabric levels. 

We would still deploy those releases on global, regional and fabric levels -  just with different values. Happy to discuss this further if it's not premature optimization. All with auditable history, versioning and other benefits of using charts.

> I thought we would put the pieces in components/openstack into a helm chart and that's it.

Ok, now I don't understand why this [was said](https://rackspace.atlassian.net/browse/PUC-521?focusedCommentId=3234939) in the JIRA about "configuration of storageclass for MariaDB" which did not mention openstack anywhere...


Now, ignoring that for a minute - let's look into the contents of this folder - we can certainly work towards moving it, but in it's current form it may not be the best idea to just throw it into a chart. It's looking like dumping ground for random pieces - it has:
- part of the mariadb cluster config (this PR takes care of that part)
- value file for memcached chart
- YAML manifest for rabbitmq cluster
- 4 manifests for External Secrets
- service account for argo workflows
- the README file which states that these are pre-req components for the OpenStack.

This README is not accurate, as you cannot deploy OpenStack without having `Databases`, `Users` and `Grants` deployed.

PR #536, opened 2024-12-05
feat: Namespace create

mfencik:
This will create a Nautobot namespace with openstack network ID as a name when create a network in openstack. This is needed when subnets will be created as Nautobot requires namespace for subnets.

### DIFF
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -3,14 +3,19 @@ import pathlib
 from urllib.parse import urljoin
 
 import requests
+from neutron_lib import exceptions as exc
 from oslo_log import log
 from requests.models import HTTPError
 
 LOG = log.getLogger(__name__)
 
 
-class NautobotError(Exception):
-    pass
+class NautobotError(exc.NeutronException):
+    message = "Nautobot error"
+
+
+class NautobotNotFoundError(NautobotError):
+    message = "%(obj)s not found in Nautobot. ref=%(ref)s"
 
 
 class Nautobot:
@@ -84,6 +89,24 @@ class Nautobot:
 
     def ucvni_delete(self, network_id):
         url = f"/api/plugins/undercloud-vni/ucvnis/{network_id}/"
+        return self.make_api_request(url, "delete")
+
+    def fetch_namespace_by_name(self, name: str) -> str:
+        url = f"/api/ipam/namespaces/?name={name}&depth=1"
+        resp_data = self.make_api_request(url, "get")
+        try:
+            return resp_data["results"][0]["id"]
+        except (IndexError, KeyError) as error:
+            LOG.error("Nautobot error: %(error)s", {"error": error})
+            raise NautobotNotFoundError(obj="namespace", ref=name) from error
+
+    def namespace_create(self, name: str) -> dict:
+        url = "/api/ipam/namespaces/"
+        payload = {"name": name}
+        return self.make_api_request(url, "post", payload)
+
+    def namespace_delete(self, namespace_uuid: str) -> dict:
+        url = f"/api/ipam/namespaces/{namespace_uuid}/"
         return self.make_api_request(url, "delete")
 
     def prep_switch_interface(
--- a/python/neutron-understack/neutron_understack/neutron_understack_mech.py
+++ b/python/neutron-understack/neutron_understack/neutron_understack_mech.py
@@ -5,7 +5,6 @@ from uuid import UUID
 
 import neutron_lib.api.definitions.portbindings as portbindings
 from neutron_lib import constants as p_const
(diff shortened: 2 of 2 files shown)


python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -130,21 +129,17 @@ def create_network_postcommit(self, context):
...
-                self.nb.ucvni_create(
-                    network_id, ucvni_group, network_name, segmentation_id
-                )
-            except Exception as e:
-                LOG.exception(
-                    "unable to create network %(net_id)s", {"net_id": network_id}
-                )
-                raise exc.NetworkNotFound(net_id=network_id) from e
-
+            self.nb.ucvni_create(network_id, ucvni_group, network_name, segmentation_id)
```
cardoe, line 132: Does this not need to be wrapped in a neutron_lib exception to work correctly (e.g. throw back a useful error instead of a 500)

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -166,19 +161,13 @@ def delete_network_postcommit(self, context):
...
             ucvni_group = conf.ucvni_group
-            try:
-                self.nb.ucvni_delete(network_id)
-            except Exception as e:
-                LOG.exception(
-                    "unable to delete network %(net_id)s", {"net_id": network_id}
-                )
-                raise exc.NetworkNotFound(net_id=network_id) from e
-
+            self.nb.ucvni_delete(network_id)
```
cardoe, line 164: Does this not need to be wrapped in a neutron_lib exception to work correctly (e.g. throw back a useful error instead of a 500)

python/neutron-understack/neutron_understack/neutron_understack_mech.py
mfencik: fixed in nautobot client by subclassing the NeutronException

python/neutron-understack/neutron_understack/neutron_understack_mech.py
mfencik: fixed in nautobot client by subclassing the NeutronException

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -86,6 +87,20 @@ def ucvni_delete(self, network_id):
         url = f"/api/plugins/undercloud-vni/ucvnis/{network_id}/"
         return self.make_api_request(url, "delete")
 
+    def fetch_namespace_by_name(self, name: str) -> str:
+        url = f"/api/ipam/namespaces/?name={name}&depth=1"
+        resp_data = self.make_api_request(url, "get")
+        return resp_data["results"][0]["id"]
```
cardoe, line 93: What if it's not found? Does this return an empty list which means we throw an IndexError exception here?

mfencik: fixed, we will raise Nautobot error

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -3,13 +3,14 @@
...
 import requests
+from neutron_lib import exceptions as exc
 from oslo_log import log
 from requests.models import HTTPError
 
 LOG = log.getLogger(__name__)
 
 
-class NautobotError(Exception):
+class NautobotError(exc.NeutronException):
```
cardoe, line 13:
```
    message = "Nautobot error"
```


Instead of pass.

We should then have:

```
class NautobotNotFoundError(NautobotError):
    message = "%(obj)s not found in Nautobot. ref=%(ref)s"
```

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -86,6 +87,24 @@ def ucvni_delete(self, network_id):
...
         return self.make_api_request(url, "delete")
 
+    def fetch_namespace_by_name(self, name: str) -> str:
+        url = f"/api/ipam/namespaces/?name={name}&depth=1"
+        resp_data = self.make_api_request(url, "get")
+        try:
+            return resp_data["results"][0]["id"]
+        except (IndexError, KeyError) as error:
+            LOG.error("Nautobot error: %(error)s", {"error": error})
+            raise NautobotError() from error
```
cardoe, line 97:
```suggestion
            raise NautobotNotFoundError(obj="namespace", ref=name) from error
```

python/neutron-understack/neutron_understack/nautobot.py
mfencik: implemented

python/neutron-understack/neutron_understack/nautobot.py
mfencik: implemented

cardoe (approved):

PR #540, opened 2024-12-09
fix: flavors: temporary fix to load the specs

skrobul:
This sets the `FLAVORS_ENV` to 'nonprod', so that the enroll server workflow can recognize which flavors to use.

In future versions, this needs to be changed so that this environment variable can be modified for each of the environments.


skrobul:
superseded by #541

PR #541, opened 2024-12-09
fix: we don't want the env name in the flavor

cardoe:
We don't want to be prefixing the flavor names with an environment name. Just put the environment data in the correct place.

### DIFF
--- a/operators/nova-flavors/nova_flavors/flavor_synchronizer.py
+++ b/operators/nova-flavors/nova_flavors/flavor_synchronizer.py
@@ -44,15 +44,13 @@ class FlavorSynchronizer:
         existing_flavors = self._nova.flavors.list()
         for flavor in desired_flavors:
             nova_flavor = next(
-                (flv for flv in existing_flavors if flv.name == flavor.stripped_name),
+                (flv for flv in existing_flavors if flv.name == flavor.name),
                 None,
             )
 
             update_needed = False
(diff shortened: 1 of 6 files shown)


nicholaskuechler (approved):

skrobul (changes requested):
ok so codewise the matching should work now, but where is the configuration that decides which flavor types to be used?

I assume we want to just mount the `prod` or `nonprod` into `/etc/understack_flavors/`, but that does not seem to be implemented yet:
- [the `enroll-server` still mounts whole](https://github.com/rackerlabs/understack/blob/c3547041fd5e99919dc06f9f45380540e5d6daec/workflows/argo-events/workflowtemplates/enroll-server.yaml#L76) `flavors` directory which contains both `prod` and `nonprod`. This must be adjusted before merging.

- [nova-flavors synchronizer](https://github.com/RSS-Engineering/undercloud-deploy/pull/224/files)  that was [recently merged]((https://github.com/rackerlabs/understack/pull/499) will need to have that configuration too. There is no `Deployment` for it yet, but when creating one we have to remember to mount only sub path.

skrobul (approved):
https://github.com/RSS-Engineering/undercloud-deploy/pull/245  adds the missing piece of setting the `FLAVORS_DIR` for both enroll-server and nova-flavors. With this, I think that PR makes more sense.

PR #550, opened 2024-12-11
fix: fix router plugin

mfencik:
The core port_forwarding plugin is hard coded to use the default "router" plugin so we can't use it with our UnderStackL3ServicePlugin. ATM we don't use the port_forwarding so we're removing it for now, if needed in the future, we can create our own by just subclassing the core port_forwarding plugin and point it to use our UnderStackL3ServicePlugin


mfencik:
looks like @skrobul beat me to it - https://github.com/rackerlabs/understack/pull/549/files

PR #558, opened 2024-12-17
feat: create/delete subnet in nautobot when openstack subnet create/delete is used

mfencik:


cardoe (approved):
+1 only side item is that I don't think we'll want to wrap every call with our own piece and will want to just use pynautobot in the end.

PR #575, opened 2025-01-06
fix: fixes dockerfile logic to use builder stage as input for prod build stage

nicholaskuechler:


cardoe (approved):

skrobul:
-1 on this change. Using builder stage as a source for production defeats the purpose  of using multi stage builds in the first place as all of the build dependencies, headers and other build assets are unnecessarily carried over.

I assume this was temporary workaround to get the build going, so I have proposed more permanent solution in #578

PR #579, opened 2025-01-07
feat(workflows): switch to nautobot native idempotent IDs

cardoe:
Nautobot has integrated the idempotent IDs feature so switch to that instead of using our custom piece.


cardoe:
I validated this by pushing a change to make the workflow use the container tag "pr-579" and creating a project via the openstack command line. I then checked that project ID in Nautobot and it matched the same ID. This is unlike before the idempotent IDs change landed in Nautobot.

stevekeay (approved):
LGTM, do we need a follow-up change to walk back our customisation of nautobot?

cardoe:
> LGTM, do we need a follow-up change to walk back our customisation of nautobot?

Yes we will.

PR #581, opened 2025-01-07
feat: adds github workflow to run a periodic build test

nicholaskuechler:


cardoe:
So I think you had a good idea about using the matrix strategy so we can test against multiple environments. You could set the OS_CLOUD variable based on the matrix.

.github/workflows/terraform-build-test.yaml
```diff
@@ -0,0 +1,47 @@
+---
+name: terraform-build-test
+
+on:
+  schedule:
+    - cron: '0 0,12 * * *'
```
skrobul, line 6: Consider changing the build hours to something that will happen when more people across the team are online

```suggestion
    - cron: '0 7,16 * * *'
```

stevekeay (approved):

PR #596, opened 2025-01-14
feat: adjust understack mech driver for VLAN network type

mfencik:


python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -126,17 +126,29 @@ def create_network_postcommit(self, context):
...
+        vlan_group_id_and_vlan_tag = self.nb.prep_switch_interface(
+            connected_interface_id=conf.network_node_switchport_uuid,
+            ucvni_uuid=network_id,
+            modify_native_vlan=False,
+            vlan_tag=int(segmentation_id),
+        )
+        self.undersync.sync_devices(
+            vlan_group_uuids=str(vlan_group_id_and_vlan_tag["vlan_group_id"]),
+            dry_run=cfg.CONF.ml2_understack.undersync_dry_run,
+        )
```
skrobul, line 151: I think there is a good refactoring opportunity here as this method already does four things:
- creates a VNI
- creates a Namespace
- adds VLAN to a network node's trunk 
- calls undersync

What do you think about extracting those to separate methods?

skrobul (approved):

python/neutron-understack/neutron_understack/neutron_understack_mech.py
mfencik: since this is temporary, i don't wanna spend more time than needed. We can always refactor later on.

PR #598, opened 2025-01-15
feat: PUC-529 : create provisioning network on successful deployment of neutron-server 

haseebsyed12:


workflows/openstack/sensors/sensor-neutron-deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+              apiVersion: batch/v1
+              kind: Job
+              metadata:
+                generateName: create-provision-network
+              spec:
+                template:
+                  spec:
+                    containers:
+                      - name: create-network
+                        image: docker.io/openstackhelm/heat:2024.1-ubuntu_jammy
```
cardoe, line 30:
```suggestion
                        image: docker.io/openstackhelm/openstack-client:2024.2
```

workflows/openstack/sensors/sensor-neutron-deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+                              secretKeyRef:
+                                name: neutron-keystone-user
+                                key: OS_USERNAME
+                          - name: SERVICE_OS_PASSWORD
+                            valueFrom:
+                              secretKeyRef:
+                                name: neutron-keystone-user
+                                key: OS_PASSWORD
+                          - name: SERVICE_OS_ROLES
+                            value: admin,service
```
cardoe, line 126:
```suggestion
```

this shouldn't be needed.

workflows/openstack/eventsources/eventsource-openstack-neutron.yaml
```diff
@@ -0,0 +1,22 @@
...
+  template:
+    serviceAccountName: openstack-events
+  resource:
+    neutron-deployment:
+      namespace: openstack
+      resource: deployments
+      group: apps
+      version: v1
+      eventTypes:
+        - ADD
```
cardoe, line 16: So is this only when the deployment is created? Can we add some descriptions / comments in these to tell the user this is how it works? We could also make this happen on update as well if we check first if the provisioning network exists?

workflows/openstack/sensors/sensor-neutron-deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+      name: openstack-neutron-server-deployment
+  triggers:
+    - template:
+        name: create-provisioning-network
+        k8s:
+          operation: create
+          source:
+            resource:
+              apiVersion: batch/v1
+              kind: Job
```
cardoe, line 22: How do we ensure only 1 job happens since it happens only on add? If we do it on update that's fine. Is there a max number of jobs that should exist? Is there a cleanup rule?

What about running this as an Argo Workflow?

cardoe: The Argo Workflow is just a question and not an ask to change things.

haseebsyed12: I tried on UPDATE event type also .. 4 jobs are executing simultaneously on UPDATE event

workflows/openstack/eventsources/eventsource-openstack-neutron.yaml
haseebsyed12: Added comments, sure I will add logic for UPDATE too. what I noticed was 4 jobs are executing simultaneously on UPDATE event

cardoe (approved):

PR #603, opened 2025-01-15
fix: update keystone memcached cache settings

nicholaskuechler:
I believe we've hit an old issue described here: https://bugs.launchpad.net/keystone/+bug/1901124


skrobul (approved):
+1 

note to self for future reference: https://dogpilecache.sqlalchemy.org/en/latest/api.html#dogpile.cache.backends.memcached.PyMemcacheBackend.params.memcached_expire_time

cardoe (approved):

PR #608, opened 2025-01-16
feat: add OVN for Neutron L3 routing

cardoe:
Added a deployment of OVS and OVN so that we can utilize OVN for Neutron L3 routing.

### DIFF
--- a/components/openstack-2024.2-jammy.yaml
+++ b/components/openstack-2024.2-jammy.yaml
@@ -68,4 +68,13 @@ images:
     placement: "docker.io/openstackhelm/placement:2024.2-ubuntu_jammy"
     placement_db_sync: "docker.io/openstackhelm/placement:2024.2-ubuntu_jammy"
 
+    # openvswitch
+    openvswitch_db_server: "docker.io/openstackhelm/openvswitch:ubuntu_jammy-20250111"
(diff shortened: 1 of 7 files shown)


components/openstack-2024.2-jammy.yaml
```diff
@@ -68,4 +68,13 @@ images:
...
 
+    # openvswitch
+    openvswitch_db_server: "docker.io/openstackhelm/openvswitch:latest-ubuntu_jammy"
+    openvswitch_vswitchd: "docker.io/openstackhelm/openvswitch:latest-ubuntu_jammy"
+
+    # ovn
+    ovn_ovsdb_nb: "docker.io/openstackhelm/ovn:ubuntu_jammy"
+    ovn_ovsdb_sb: "docker.io/openstackhelm/ovn:ubuntu_jammy"
+    ovn_northd: "docker.io/openstackhelm/ovn:ubuntu_jammy"
+    ovn_controller: "docker.io/openstackhelm/ovn:ubuntu_jammy"
```
skrobul, line 79: this image seems to be updated frequently - what do you think about pinning it to `ubuntu_jammy-20250111` for example ?

cardoe: Yeah I'd be fine with that. I really would like to build our own images of all these components to avoid the docker.io rate limiting issue. We've got the quay.io account that would also give us scanning feedback.

skrobul (approved):

PR #611, opened 2025-01-17
fix: rename workflow parameters

stevekeay:
A long time ago this workflow used to take a switch ID argument.  This was changed to a Vlan Group ID argument, but I was lazy and failed to update identifiers and parameter strings to reflect the change in behaviour.  This PR cleans up the confusment.

closes PUC-732

### DIFF
--- a/python/understack-workflows/understack_workflows/main/undersync_device.py
+++ b/python/understack-workflows/understack_workflows/main/undersync_device.py
@@ -117,7 +117,7 @@ def call_undersync(args, vlan_group_id: UUID):
 
     try:
         return undersync.sync_devices(
-            [str(vlan_group_id)], dry_run=args.dry_run, force=args.force
+            str(vlan_group_id), dry_run=args.dry_run, force=args.force
         )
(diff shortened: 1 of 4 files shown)


python/understack-workflows/understack_workflows/main/undersync_device.py
```diff
@@ -117,7 +117,7 @@ def call_undersync(args, vlan_group_id: UUID):
 
     try:
         return undersync.sync_devices(
-            [str(vlan_group_id)], dry_run=args.dry_run, force=args.force
+            str(vlan_group_id), dry_run=args.dry_run, force=args.force
```
nicholaskuechler, line 120: Seems OK to me, but I did notice in the other files it's `vlan_group_uuid` but in this one it's just `vlan_group_id`. Not sure if these were also meant to be changed to `vlan_group_uuid`.

skrobul (approved):
lgtm, although @nicholaskuechler  has good point about standardizing on one version

PR #625, opened 2025-01-23
feat: Neutron: consistent vlan assignments 

skrobul:
This PR modifies the Understack mechanism driver by hooking `create_network_precommit` event to verify if the selected VLAN ID is available in particular fabric before assigning it.

That feature heavily relies on the Nautobot data being correct. The concept of availability is abstracted away through custom API plugin in Nautobot which returns simple 'true/false'.

For the input, we take the `interface_id` which is configured statically in Neutron's configuration as `network_node_switchport_uuid`. This refers to an interface of a "network node" which presumably is always the first interface to be hooked into the network.

The second input is a `vlan_tag` as chosen by the Neutron.

If Nautobot considers the vlan to be available, the hook returns and the allocation proceeds as normal. If it happens to be taken, the Neutron is asked to allocate new segment (and thus new VLAN ID) and the check is repeated.

In addition to this, for extra safety, the same check is also executed during prep_switch_interface but that's out of scope for this commit because it's implemented in the Nautobot plugin change.

Goes together with https://github.com/RSS-Engineering/undercloud-rackspace/pull/338

Related: https://rackspace.atlassian.net/browse/PUC-736

### DIFF
--- a/python/neutron-understack/neutron_understack/config.py
+++ b/python/neutron-understack/neutron_understack/config.py
@@ -50,6 +50,16 @@ mech_understack_opts = [
             "is used to trunk all vlans used by a neutron router."
         ),
     ),
+    cfg.BoolOpt(
+        "enforce_unique_vlans_in_fabric",
+        default=True,
+        help=(
+            "When enabled, Neutron performs an extra check during the creation of a"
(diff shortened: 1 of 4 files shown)


mfencik (approved):
LGTM

PR #626, opened 2025-01-23
fix: re-enable full erasing of devices during cleaning

nicholaskuechler:
The existing settings skip the full erasing during cleaning. This re-enables the full erasing.


cardoe (commented):
Does this rebuild the RAID on deploy as well?

PR #629, opened 2025-01-24
fix: update enroll workflow to put node in manageable status before cleaning it

nicholaskuechler:
The node needs to be in `manageable` status before we can perform the raid cleaning.


mfencik (approved):
LGTM

PR #638, opened 2025-01-31
feat: understack mech: configurable VXLAN/VLAN operation

skrobul:
This allows the ML2 to operate both in VLAN and VXLAN environments.

related: https://github.com/RSS-Engineering/undercloud-rackspace/pull/344

### DIFF
--- a/components/nautobot/nautobot_config.py
+++ b/components/nautobot/nautobot_config.py
@@ -425,8 +425,9 @@ INSTALLATION_METRICS_ENABLED = is_truthy(
 # }
 PLUGINS_CONFIG = {
     "vni_custom_model": {
-        "FORCE_UNIQUE_VLANS": os.getenv("VNI_CUSTOM_MODEL_FORCE_UNIQUE_VLANS", None)
-        == "true"
+        "FORCE_UNIQUE_VLANS": is_truthy(
+            os.getenv("VNI_CUSTOM_MODEL_FORCE_UNIQUE_VLANS", "false")
+        )
     }
 }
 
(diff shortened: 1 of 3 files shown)


components/nautobot/nautobot_config.py
```diff
@@ -423,6 +423,12 @@
 #         'buzz': 'bazz'
 #     }
 # }
+PLUGINS_CONFIG = {
```
stevekeay, line 426: It seems like this duplicates the section added in https://github.com/RSS-Engineering/undercloud-rackspace/pull/344/files - I assume there is a reason why it needs to be in two places.

stevekeay (approved):

components/nautobot/nautobot_config.py
skrobul: It is indeed a duplicate, the linked section is for rackspace-specific setup that also has bunch of other settings. I am adding it here, because we have modified the generic Understack mech driver and it expects that configuration to exist.

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -143,16 +143,18 @@ def create_network_postcommit(self, context):
         )
         self._create_nautobot_namespace(network_id, external)
 
+        vlan_tag = int(segmentation_id) if provider_type == p_const.TYPE_VLAN else None
```
mfencik, line 146: As discussed, in this section, we should only trigger prep_switch interface when it's vlan type network.

mfencik (changes requested):
image

mfencik (approved):
LGTM

PR #649, opened 2025-02-04
feat: Create SVIs on leaf switches using a Network with a router-external Subnet 

stevekeay:
This PR has the required ml2 driver changes.  Note that this
represents a change to the meaning of the Nautotbot data.

- [x] release https://github.com/RSS-Engineering/undersync/pull/30 to understand this new data.
- [x] We should change the one-to-one relationship for
      UCVNI->Prefic to a one-to-many relationship
- [x] extend existing roles from Interface to VNI
- [x] extend existing cf_dhcp_relay_ipv4_address to VNI
- [x] restart nautobot so that custom field changes take effect
- [x] cleanup required for existing data, remove unnecessary relations
- [x] convert existing virtual Interfaces to new format

We currently represent SVIs explicitly in Nautobot:
- we have an IpAddress object with the SVI address (.1)
- for each leaf Device, we have an Interface named "Vlan{vlan_id}"
- we associate the IpAddress with each Interface
- Interface.role is used to select the SVI configuration template

We simplify this by leveraging the direct relationship between VNI and Prefix.

Using the existing Prefix->VNI mapping is a natural way to model the SVI.

- For VLANs which have a prefix, via Vlan->UCVNI->Prefix
- SVI is implicitly created wherever we create the VLAN
- bake in assumption that the SVI has .1 IP Address
- specify the template to use with VNI.role
- optionally specify DHCP relay address with custom field

This has advantages:
- data is better normalized
- fewer API or DB operations to create/delete an SVI
- keeps the "-1" and "-2" switches firmly in sync (important in a VPC pair)
- when the vlan is deleted the "SVI" will go away automatically
- easy to automatically create the SVI wherever the VLAN is used

When the subnet is created in openstack

We already add the prefix to nautobot today.

However, if the subnet's "router:external" attribute is set to true, we now
perform the additional step of associating the subnet with the VNI.

When the subnet is deleted, that relationship will implicitly
cascade-delete.

We don't handle updates - it is currently assumed that no updates will
occur except for create and delete.

Note that we will not update the switches on subnet create or delete.
It is assumed that the subnet will be created before any ports are
associated with that network.

It is further assumed that is a subnet is deleted, the network will be
deleted shortly thereafter and this will perform the required cleanup.

### DIFF
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -122,6 +122,22 @@ class Nautobot:
         }
         return self.make_api_request(url, "post", payload)
 
+    def associate_subnet_with_network(
+        self, network_uuid: str, subnet_uuid: str, role: str
+    ):
+        url = f"/api/plugins/undercloud-vni/ucvnis/{network_uuid}"
+        payload = {
+            "role": {"name": role},
+            "relationships": {
+                "ucvni_prefix": {
+                    "destination": {
+                        "objects": [subnet_uuid],
+                    },
+                },
+            },
+        }
+        self.make_api_request(url, "put", payload)
+
     def subnet_delete(self, subnet_uuid: str) -> dict:
         url = f"/api/ipam/prefixes/{subnet_uuid}/"
         return self.make_api_request(url, "delete")
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -38,7 +38,7 @@ class Nautobot:
         try:
             response.raise_for_status()
         except HTTPError as error:
-            LOG.error("Nautobot error: %(error)s", {"error": error})
+            LOG.error("Nautobot error: %s %s", error, response.content)
             raise NautobotError() from error
 
     def make_api_request(
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -53,7 +53,12 @@ class Nautobot:
             {"payload": payload, "caller_function": caller_function},
         )
         resp = self.s.request(
-            http_method, endpoint_url, timeout=10, json=payload, params=params
+            http_method,
+            endpoint_url,
+            timeout=10,
+            json=payload,
+            params=params,
+            allow_redirects=False,
         )
 
         if resp.content:
@@ -125,7 +130,7 @@ class Nautobot:
     def associate_subnet_with_network(
         self, network_uuid: str, subnet_uuid: str, role: str
     ):
-        url = f"/api/plugins/undercloud-vni/ucvnis/{network_uuid}"
+        url = f"/api/plugins/undercloud-vni/ucvnis/{network_uuid}/"
         payload = {
             "role": {"name": role},
             "relationships": {
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -134,7 +134,7 @@ class Nautobot:
         payload = {
             "role": {"name": role},
             "relationships": {
-                "ucvni_prefix": {
+                "ucvni_prefixes": {
                     "destination": {
                         "objects": [subnet_uuid],
                     },
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -136,7 +136,7 @@ class Nautobot:
                 },
             },
         }
-        self.make_api_request(url, "put", payload)
+        self.make_api_request(url, "patch", payload)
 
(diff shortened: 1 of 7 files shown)


python/understack-workflows/devbox.json
```diff
@@ -10,7 +10,8 @@
     "kubeseal@latest",
     "kustomize@latest",
     "telepresence2@latest",
-    "jq@latest"
+    "jq@latest",
+    "poetry@latest"
```
skrobul, line 14: we may want a specific version (currently 1.8.3) here to avoid unnecessary diffs in [`poetry.lock`](https://github.com/rackerlabs/understack/blob/5e1bb3f33f7598a37a2df2f8dc4c56e54450954c/python/neutron-understack/poetry.lock#L1) when new version comes out.

skrobul (approved):

python/understack-workflows/devbox.json
stevekeay: good idea, done in rebase

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -122,6 +122,22 @@ def subnet_create(self, subnet_uuid: str, prefix: str, namespace_name: str) -> d
...
+            "role": {"name": role},
+            "relationships": {
+                "ucvni_prefix": {
+                    "destination": {
+                        "objects": [subnet_uuid],
+                    },
+                },
+            },
+        }
+        self.make_api_request(url, "put", payload)
```
skrobul, line 139: since this updates the already existing VNI we probably want to `PATCH` it rather than updating wholesale with `PUT`.
When testing this with @mfencik  we tried changing it to `PATCH`, but that has presented new problem - Nautobot requires the `ucvni_id` (note that it's not an UUID but simply a VNI number) in the request body.
This can be replicated with:

```
curl --request PATCH \
  --url http://localhost:8000/api/plugins/undercloud-vni/ucvnis/bc1679a9-5219-4f5e-bb7d-2d423102f27c/ \
  --header 'Authorization: Token {{NAUTOBOT_TOKEN}}' \
  --header 'Content-Type: application/json' \
  --header 'content-type: application/json' \
  --data '{
  "role": {
    "name": "svi_vxlan_anycast_gateway"
  },
  "relationships": {
    "ucvni_prefix": {
      "destination": {
        "objects": [
          "c5efc724-9d1d-4c24-8e33-0f7fd47c5149"
        ]
      }
    }
  }
}'

```


adding the `ucvni_id` by hand worked but obviously there is some extra work needed

skrobul (changes requested):
@stevekeay we tried to run this in dev today, unfortunately it did not work. Few things found along the way:

- [`shared_nautobot_namespace` neutron setting was missing](https://github.com/RSS-Engineering/undercloud-deploy/pull/275) fixed but not merged as we hijacked that branch to run custom image for neutron.
- Nautobot validator for `prep_switch_interface` did not really allow null values for `vlan_tag`, fixed in https://github.com/RSS-Engineering/undercloud-rackspace/pull/359
- Github won't build the new version of the Neutron container if there is a merge conflict of the PR so if you want to test further changes, make sure to rebase against `origin/main`.

Commands used for test:

```
openstack network create --external --provider-network-type vxlan test-ext1  
openstack subnet create --no-dhcp --network test-ext1 --subnet-range 146.20.241.224/28 test-ext1-subnet
openstack server create   --image 'Ubuntu-24.04' --flavor gp2.small --nic net-id=test-ext1 --use-config-drive --key-name team sviprov.marek.test
```

I am working only half day today so can pick this up further tomorrow but thought I'd let you know in case you want to play around with it in the meantime.

python/neutron-understack/neutron_understack/nautobot.py
mfencik: the problem is here https://github.com/RSS-Engineering/undercloud-rackspace/blob/main/vni-custom-model/vni_custom_model/api/serializers.py#L11 as it appears that that validation runs on PATCH as well and when ucvni_id is not present and there's no ucvni_group, it will return None. We need to adjust it to account for PATCH

stevekeay:
I Rebased onto main and fixed the merge conflicts

stevekeay:
rebased once again onto main

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -38,7 +38,7 @@ def _log_and_raise_for_status(self, response):
         try:
             response.raise_for_status()
         except HTTPError as error:
-            LOG.error("Nautobot error: %(error)s", {"error": error})
+            LOG.error("Nautobot error: %s %s", error, response.content)
             raise NautobotError() from error
```
cardoe, line 42: Honestly some kind of detailed exception is what should be getting thrown here rather than the default which has a message body still of "Nautobot error"

skrobul (approved):
tested and ready to merge when https://github.com/RSS-Engineering/undercloud-rackspace/pull/360 is merged.

PR #650, opened 2025-02-05
feat: PUC-752: reading OIDCCryptoPassphrase from a file

haseebsyed12:
Problem: 
Unable to authenticate when the user receives the initial state, it is encrypted with a key from pod A, then user is redirected to the identity provided and finally when they get back through redirect, they may land on pod B which is no longer able to decrypt the state.

Solution:

OIDCCryptoPassphrase is a password used for encryption of the state cookie and cache entries and it must be the same on all replicas.

Previously we were setting the OIDCCryptoPassphrase to a random value that is generated on pod startup and multiple pods have different value.

### DIFF
--- a/scripts/gitops-secrets-gen.sh
+++ b/scripts/gitops-secrets-gen.sh
@@ -261,10 +261,22 @@ convert_to_var_name() {
 convert_to_secret_name() {
     echo "$1" | tr '[:upper:]' '[:lower:]' | tr '_' '-'
 }
+# Default password generator using pwgen.sh
+# shellcheck disable=SC2317
+default_pwgen() {
+    "${SCRIPTS_DIR}/pwgen.sh" 2>/dev/null
+}
+
+# Custom password generator with only alphabets
+# shellcheck disable=SC2317
+alpha_only_pwgen() {
+    head /dev/urandom | tr -dc A-Za-z | head -c 32
+}
 
 load_or_gen_os_secret() {
     local data_var=$1
     local secret_var=$2
+    local gen_func=${3:-default_pwgen}
 
     if kubectl -n openstack get secret "${secret_var}" &>/dev/null; then
         data="$(kubectl -n openstack get secret "${secret_var}" -o jsonpath='{.data.password}' | base64 -d)"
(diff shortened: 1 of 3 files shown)


cardoe (commented):
Can you write a description of why the change is needed to leave us some bread crumbs for future logic. I've recently found myself needing to refer to our commits in upstream OpenStack projects and our commits have been lacking in explanation which would be helpful.

scripts/gitops-secrets-gen.sh
```diff
@@ -261,10 +261,20 @@ convert_to_var_name() {
 convert_to_secret_name() {
     echo "$1" | tr '[:upper:]' '[:lower:]' | tr '_' '-'
 }
+# Default password generator using pwgen.sh
```
skrobul, line 264:
```suggestion
# Default password generator using pwgen.sh
# shellcheck disable=SC2317
```

skrobul (approved):

scripts/gitops-secrets-gen.sh
```diff
@@ -261,10 +261,22 @@ convert_to_var_name() {
...
+# shellcheck disable=SC2317
+default_pwgen() {
+    "${SCRIPTS_DIR}/pwgen.sh" 2>/dev/null
+}
+
+# Custom password generator with only alphabets
+# shellcheck disable=SC2317
+alpha_only_pwgen() {
+    head /dev/urandom | tr -dc A-Za-z | head -c 32
+}
```
cardoe, line 274:
```suggestion
default_pwgen() {
    "${SCRIPTS_DIR}/pwgen.sh" 2>/dev/null
}

# Custom password generator with only alphabets
# shellcheck disable=SC2317
alpha_only_pwgen() {
    head /dev/urandom | tr -dc A-Za-z | head -c 32
}
```

So we're not setting LC_ALL to C like the `pwgen.sh` script does. Let's instead add another parameter to `pwgen.sh` and let you specify the range.

So in that script do `tr -dc ${2:-_A-Z-a-z-0-9}`

scripts/gitops-secrets-gen.sh
```diff
@@ -274,7 +286,7 @@ load_or_gen_os_secret() {
         return 1
     else
         echo "Generating ${secret_var}"
-        data="$("${SCRIPTS_DIR}/pwgen.sh" 2>/dev/null)"
+        data="$(${gen_func})"
```
cardoe, line 289:
```suggestion
        data="$("${SCRIPTS_DIR}/pwgen.sh" 32 "A-Za-z" 2>/dev/null)"
```

cardoe (changes requested):

cardoe (approved):

PR #663, opened 2025-02-10
feat(neutron): make our VXLAN type driver unique

cardoe:
Instead of standing in for the VXLAN type driver directly, have it be its own thing that will be ignored by mechanisms that consume it. The reason is that the existing VXLAN type driver is treated and consumed as a L3VNI while we need L2VNI.


cardoe:
Do not merge until the `drop!` commit is dropped.

nicholaskuechler (approved):

PR #665, opened 2025-02-10
chore(neutron): drop unused net types and use new config

cardoe:
We won't ever support gre and flat networks so drop those out right. local is really for testing as a no-op so remove that as well but it might find its way back in for test purposes later. Fixed incorrect usage of pass. Handle old and new config file location.


mfencik (approved):
LGTM

PR #668, opened 2025-02-12
feat: understack trunk

mfencik:
This PR will enable us to set "tenant_vlan_id" on UCVNI in Nautobot when first subport is added to neutron trunk and user specifies the "segmentation_id" when adding the subport. The idea is that we will use the "tenant_vlan_id" as mapped VLAN that is local to the switchport during server provisioning. 
This PR covers only the trunk driver part and there will be another one where we will adjust the understack mechanism driver, so when server will go through provisioning, we will detect the trunk_details on the port attached and do the needful to configure our network.

### DIFF
--- a/python/neutron-understack/neutron_understack/trunk.py
+++ b/python/neutron-understack/neutron_understack/trunk.py
@@ -1,14 +1,56 @@
+from neutron.objects.trunk import SubPort
 from neutron.services.trunk.drivers import base as trunk_base
+from neutron_lib import exceptions as exc
 from neutron_lib.api.definitions import portbindings
+from neutron_lib.callbacks import events
+from neutron_lib.callbacks import registry
+from neutron_lib.callbacks import resources
 from neutron_lib.services.trunk import constants as trunk_consts
 from oslo_config import cfg
+from oslo_log import log
+
+from neutron_understack import config
+from neutron_understack import utils
+from neutron_understack.nautobot import Nautobot
+
+config.register_ml2_understack_opts(cfg.CONF)
+
+LOG = log.getLogger(__name__)
 
 SUPPORTED_INTERFACES = (portbindings.VIF_TYPE_OTHER,)
 
 SUPPORTED_SEGMENTATION_TYPES = (trunk_consts.SEGMENTATION_TYPE_VLAN,)
 
 
+class SubportSegmentationIDError(exc.NeutronException):
+    message = (
+        "Segmentation ID: %(seg_id)s cannot be set to the Subport: "
+        "%(subport_id)s as there is already another Segmentation ID: "
+        "%(nb_seg_id)s in use by the Network: %(net_id)s that is "
+        "attached to the Subport. Please use %(nb_seg_id)s as "
+        "segmentation_id for this subport."
+    )
+
+
 class UnderStackTrunkDriver(trunk_base.DriverBase):
+    def __init__(
+        self,
+        name,
+        interfaces,
+        segmentation_types,
+        agent_type=None,
+        can_trunk_bound_port=False,
+    ):
+        super().__init__(
+            name,
+            interfaces,
+            segmentation_types,
+            agent_type=agent_type,
+            can_trunk_bound_port=can_trunk_bound_port,
+        )
+        conf = cfg.CONF.ml2_understack
+        self.nb = Nautobot(conf.nb_url, conf.nb_token)
+
     @property
     def is_loaded(self):
         try:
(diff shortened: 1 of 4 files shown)


python/neutron-understack/neutron_understack/trunk.py
```diff
@@ -1,14 +1,56 @@
...
+from neutron_lib.callbacks import resources
 from neutron_lib.services.trunk import constants as trunk_consts
 from oslo_config import cfg
+from oslo_log import log
+
+from neutron_understack import config
+from neutron_understack import utils
+from neutron_understack.nautobot import Nautobot
+
+config.register_ml2_understack_opts(cfg.CONF)
```
skrobul, line 16: I am not sure if that should that really be part of the ML2 plugin?

image


Now that this config (and as far as I can see it's only for Nautobot credentials) is used in both ML2 and service plugins, should we move that to something more generic?

python/neutron-understack/neutron_understack/trunk.py
```diff
@@ -26,3 +68,63 @@ def create(cls, plugin_driver):
...
+                network_id=subport_network_id
+            )
+
+            self._configure_tenant_vlan_id(
+                tenant_vlan_id=ucvni_tenant_vlan_id,
+                ucvni_uuid=subport_network_id,
+                subport=subport,
+            )
+
+    def subports_added(self, resource, event, trunk_plugin, payload):
```
skrobul, line 122:
```suggestion
    def on_subports_add(self, resource, event, trunk_plugin, payload):
```

python/neutron-understack/neutron_understack/trunk.py
```diff
@@ -26,3 +68,63 @@ def create(cls, plugin_driver):
...
+        subports = payload.metadata["subports"]
+        self._subports_added(subports)
+
+    def trunk_created(self, resource, event, trunk_plugin, payload):
```
skrobul, line 126:
```suggestion
    def on_trunk_create(self, resource, event, trunk_plugin, payload):
```

python/neutron-understack/neutron_understack/trunk.py
```diff
@@ -26,3 +68,63 @@ def create(cls, plugin_driver):
...
+        elif tenant_vlan_id != subport_seg_id:
+            subport.delete()
+            raise SubportSegmentationIDError(
+                seg_id=subport_seg_id,
+                net_id=ucvni_uuid,
+                nb_seg_id=tenant_vlan_id,
+                subport_id=subport.port_id,
+            )
+
+    def _subports_added(self, subports: list[SubPort]) -> None:
```
skrobul, line 108:
```suggestion
    def _configure_subports(self, subports: list[SubPort]) -> None:
```

skrobul (commented):
Overall looks good, but have a few minor suggestions for potential improvements (comments inline).
Also the `UnderStackTrunkDriver` class could really use some docstrings, at least on the public methods with a high level explanation of why we react to certain events and what we intend to do.

Another potential refactoring opportunity is in how the configure_tenant_vlan_id and _subports_added are split, consider something similar to this:

```python
    def _configure_subports(self, subports: list[SubPort]) -> None:
        for subport in subports:
            network_id = utils.fetch_subport_network_id(subport_id=subport.port_id)
            tenant_vid = self.nb.fetch_tenant_vid(
                network_id=network_id
            )

            if tenant_vid:
                self._update_vni(tvid, subport)
            elif tenant_vid != subport.segmentation_id:
                self._handle_mismatch(tenant_vid, subport, network_id)


    def _update_vni(tvid, subport):
        ucvni_uuid = utils.fetch_subport_network_id(subport_id=subport.port_id)
        self.nb.add_tenant_vlan_tag_to_ucvni(
            network_uuid=ucvni_uuid, vlan_tag=subport.segmentation_id
        )
        LOG.info(
            "Segmentation ID: %(seg_id)s is now set on Nautobot's UCVNI "
                "UUID: %(ucvni_uuid)s in the tenant_vlan_id custom field",
            {"seg_id": subport.segmentation_id, "ucvni_uuid": ucvni_uuid},
        )

    def _handle_mismatch(tenant_vlan_id, subport, subport_network_id):
        if tenant_vlan_id != subport.segmentation_id:
            err = SubportSegmentationIDError(
                seg_id=subport.segmentation_id,
                net_id=subport_network_id,
                nb_seg_id=tenant_vlan_id,
                subport_id=subport.port_id,
            )

            subport.delete()
            raise err
```

^ untested

python/neutron-understack/neutron_understack/trunk.py
mfencik: I disagree with this rename as on_ may suggest that this call happens before commit, where this happens after the fact.

python/neutron-understack/neutron_understack/trunk.py
mfencik: I disagree with this rename as on_ may suggest that this call happens before commit, where this happens after the fact.

python/neutron-understack/neutron_understack/trunk.py
mfencik: We are not configuring subports, merely reacting on the call.

skrobul: What happens during that reaction? seems like you are **configuring** tenant_vlan_id on a list of **subports**. I appreciate that it's not doing changes on the neutron side, but my gripe is mostly with two methods that are called almost exactly the same minus one character plus they reference each other so probably need slightly better name which does not have to be configure_subports.

skrobul (approved):

python/neutron-understack/neutron_understack/trunk.py
mfencik: refactored, we will use UnderstackDriver's nautobot client as the driver is passed on .create

PR #669, opened 2025-02-12
fix: Further updates to Nautobot api response logging

stevekeay:
Logs more information, but limits the size of any raw payloads we dump out to the logger.

Also extracts the "error" key instead of logging the whole json payload.

```
neutron_understack.nautobot.NautobotRequestError: Nautobot API ERROR 404 for https://nautobot.dev.undercloud.rackspace.net/testing-failures POST {'payload1': 123}: {'body': 'b\'\\n\\n\\n\\n<!DOCTYPE html>\\n<html lang="en">\\n<head>\\n    '
         '<title>Page Not Found - Nautobot</title>\\n    \\n\\n\\n\\n    <link '
         'rel="stylesheet" id="template-theme"\\n          '
         'href="/template.css">\\n    <lin...10971 more chars'}
```

### DIFF
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -65,6 +65,8 @@ class Nautobot:
             response_data = {"status_code": response.status_code, "body": ""}
 
         if response.status_code >= 300:
+            response_data = response_data.get("error", response_data)
+
             raise NautobotRequestError(
                 code=response.status_code, url=full_url, body=response_data
             )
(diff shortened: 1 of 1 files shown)


cardoe:
So I'm a -1 for pretty printing in log files because we should be sending this to stuff like ELK and those log lines get parsed and split up with the leading context. As soon as you force newlines in that messes that up.

cardoe (approved):

stevekeay:
> So I'm a -1 for pretty printing in log files because we should be sending this to stuff like ELK and those log lines get parsed and split up with the leading context. As soon as you force newlines in that messes that up.

Fair enough, I removed the pretty-printing and rebased on main.

PR #673, opened 2025-02-13
fix: tune openstack db backup settings

nicholaskuechler:
Adjusts the settings for backups. Users can override these settings if they need. Reducing volume and retention of backups will help in our smaller dev environments where I had just hit some disk space issues.


cardoe (approved):
Realistically we'll want defaults here and it be overridden in any deploy.

PR #674, opened 2025-02-13
chore: bump horizon and placement to latest helm chart versions

nicholaskuechler:


apps/appsets/openstack.yaml
```diff
@@ -38,7 +38,7 @@ spec:
                   repoURL: https://tarballs.opendev.org/openstack/openstack-helm
                   openstackRelease: 2024.2
                   # renovate: datasource=custom.openstackhelm depName=placement
-                  chartVersion: 0.3.17
+                  chartVersion: 2024.2.0
```
cardoe, line 41: This is Vlad's rolling release version. You've got to pick one with the git suffix like I did for keystone above for it to remain stable.

apps/appsets/openstack.yaml
```diff
@@ -58,7 +58,7 @@ spec:
                   repoURL: https://tarballs.opendev.org/openstack/openstack-helm
                   openstackRelease: 2024.2
                   # renovate: datasource=custom.openstackhelm depName=horizon
-                  chartVersion: 0.3.29
+                  chartVersion: 2024.2.0
```
cardoe, line 61: This is Vlad's rolling release version. You've got to pick one with the git suffix like I did for keystone above for it to remain stable.

cardoe (changes requested):

PR #675, opened 2025-02-14
feat(neutron): fix config paths for different startup case and drop rpc-server config

cardoe:
Due to https://bugs.launchpad.net/oslo.config/+bug/2098514 the path for the config file snippet that was originally used can change and we don't want that. So change paths and document why this is. Added a document explaining how you can provide further extensions for your deployment as well. Dropped the rpc-server section since we've got that disabled always now.


mfencik (approved):
LGTM

PR #678, opened 2025-02-18
feat(nova): build our own container with patches

cardoe:
Add patches to fix:
- ironic node validate error message log
- metadata trunk_details added

### DIFF
--- a/containers/nova/Dockerfile.nova
+++ b/containers/nova/Dockerfile.nova
@@ -1,0 +1,14 @@
+# syntax=docker/dockerfile:1
+
+ARG OPENSTACK_VERSION="required_argument"
+FROM docker.io/openstackhelm/nova:${OPENSTACK_VERSION}-ubuntu_jammy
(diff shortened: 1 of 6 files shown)


containers/nova/Dockerfile.nova
```diff
@@ -0,0 +1,14 @@
+# syntax=docker/dockerfile:1
+
+ARG OPENSTACK_VERSION="required_argument"
+FROM docker.io/openstackhelm/nova:${OPENSTACK_VERSION}-ubuntu_jammy
```
skrobul, line 4: These patches are likely quite specific to that version of the code, should we hardcode the base image here? Or just leave it and fix it when it breaks? Dunno.

cardoe: Yeah. I'd like to get to switching to building our own container from a branch we point to.

mfencik (approved):

PR #680, opened 2025-02-18
fix(renovate): change the OpenStack Helm version method

cardoe:
Change how we detect version changes for OpenStack Helm by using the helm data source template.

PUC-641 #done fixed up renovate for OpenStack Helm


cardoe:
@renovate debug

PR #698, opened 2025-02-19
feat: use MariaDB operator for creating OpenStack oslo db connection string

cardoe:
Rather than having to pass around the DB password in plain text to OpenStack Helm, let the MariaDB operator create a file snippet that contains the properly formatted db connection string for oslo.config to use and let each of our services load it.

Blocked until the following changes are merged and our chart is updated to a new enough version to have it:
- https://review.opendev.org/c/openstack/openstack-helm/+/942131 (placement)
- TODO (keystone)

The keystone change is an unknown to me because `.Values.pod.mounts.keystone_db_sync.keystone_db_sync.volumeMounts` needs to merge with https://opendev.org/openstack/openstack-helm/src/commit/34d1672a65b14ef7452d6308f8c79393b6bef0fb/keystone/templates/job-db-sync.yaml#L77


cardoe:
Based on upstream feedback, https://review.opendev.org/c/openstack/openstack-helm-infra/+/942294 will do it generically for all charts using the db-sync job.

This isn't necessarily complete because there are other pods that have the config mounted and we'll need to see if those need the database credentials and then make sure to get that loaded into those pods. There might be more fixed needed upstream to get the loading working.

PR #702, opened 2025-02-20
feat: add UnderstackDriver methods for trunks

mfencik:


python/neutron-understack/neutron_understack/tests/test_neutron_understack_mech.py
```diff
@@ -17,8 +18,19 @@ def current_context() -> dict:
...
 @pytest.fixture
-def context(current_context) -> MagicMock:
-    return MagicMock(current=current_context)
+def original_context(current_context) -> dict:
+    original = deepcopy(current_context)
+    return original
+
+
+@pytest.fixture
+def context(current_context, original_context) -> MagicMock:
```
cardoe, line 27:
```suggestion
def port_context(current_context, original_context) -> MagicMock:
```

So ultimately rather than a MagicMock here we should be returning a `PortContext`. There's quite a few helpers available in `neutron.tests.unit.fake_resources`

```
from neutron.tests.unit import fake_resources as fakes

def port_context(current_context, original_context) -> PortContext:
    return fakes.FakePortContext(current=current_context, original_context=original_context)
```

For example. Doesn't have to be done now but we'll want to improve on this.

PR #706, opened 2025-02-20
feat: modify anaconda deploy interface to support ESXi installs

cardoe:
This modifies Ironic so that we can utilize the anaconda deploy interface to install VMware ESXi. The changes are 3 patches, all of which have been submitted upstream:
- https://review.opendev.org/c/openstack/ironic/+/942496 to fix Ironic's handling of Glance image properties since switching to the OpenStackSDK from the glanceclient
- https://review.opendev.org/c/openstack/ironic/+/942590 to allow the anaconda deploy interface to use the provisioning network
- https://review.opendev.org/c/openstack/ironic/+/942849 to change how the config drive data is passed along so we can handle it within VMware's kickstart syntax

Lastly we ship a changed ipxe_config.template with iPXE kernel and module arguments and parameters which line up with what ESXi's multiboot loader needs.

This still requires you to craft a correct kickstart template, modify boot.cfg appropriately, and host the ESXi CD contents somewhere accessible.

### DIFF
--- a/apps/appsets/operators.yaml
+++ b/apps/appsets/operators.yaml
@@ -22,7 +22,7 @@ spec:
                   sources:
                     - repoURL: https://charts.rook.io/release
                       chart: rook-ceph
-                      targetRevision: v1.15.0
+                      targetRevision: v1.16.4
                       helm:
                         releaseName: rook-ceph
                         valueFiles:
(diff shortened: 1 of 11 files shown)


cardoe:
We've pivoted to using https://github.com/rackerlabs/esxi-img to create images that can be used to make images that can be used by the direct deploy interface.

PR #707, opened 2025-02-21
fix(nova): fix trunk details patch and use correct container

cardoe:
Fix a bug in the trunk details patch and switch to using our built container for nova.


mfencik (approved):
LGTM

PR #746, opened 2025-03-04
fix: adds 'bsdextrautils' package to IPA image for hexdump command

nicholaskuechler:
https://manpages.debian.org/bullseye/bsdextrautils/hexdump.1.en.html


cardoe (commented):
This should probably be a patch upstream.

nicholaskuechler:
https://review.opendev.org/c/openstack/ironic-python-agent-builder/+/943334

PR #750, opened 2025-03-05
feat: PUC-788: assign tenant to UCVNI when the network is created

haseebsyed12:


python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -113,6 +117,8 @@ def ucvni_create(
             payload["ucvni_id"] = segment_id
             payload["ucvni_type"] = "INFRA"
 
+        LOG.info("haseeb payload is")
+        LOG.info(json.dumps(payload))
```
cardoe, line 121: Please delete these two lines.

cardoe (changes requested):

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -98,12 +100,14 @@ def make_api_request(
...
         self,
         network_id: str,
+        project_id: str,
         ucvni_group: str,
         network_name: str,
         segment_id: int | None = None,
     ):
         payload = {
             "id": network_id,
+            "tenant": str(uuid.UUID(project_id)),
```
cardoe, line 110: So this is validating and I agree with the types but it should probably be on the helper method instead of taking a str.

python/neutron-understack/neutron_understack/utils.py
```diff
@@ -6,3 +8,10 @@ def fetch_subport_network_id(subport_id):
     context = n_context.get_admin_context()
     neutron_port = port_obj.Port.get_object(context, id=subport_id)
     return neutron_port.network_id
+
+
+def convert_to_uuid(str_id: str) -> str:
```
mfencik, line 13: convert_to_uuid name suggests that the return format is UUID, but it is a string, perhaps something like validate_uuid_format would better?

mfencik (changes requested):

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -98,12 +100,14 @@ def make_api_request(
     def ucvni_create(
         self,
         network_id: str,
+        project_id: str,
```
cardoe, line 103:
```suggestion
        project_id: UUID,
```

Is what I meant in my review. But we need to do that for all arguments.

cardoe (approved):
Just a note on a follow on improvement.

python/neutron-understack/neutron_understack/nautobot.py
haseebsyed12: we cannot send UUID type attributes directly into payload we need to convert back into string otherwise we get "UUID is not JSON serializable" error.
Removing the conversion logic completely, since Django can handle it.

cardoe: Right. Didn't mean we send it directly. Let's get this merged and we can mess with it afterwards.

cardoe (approved):

cardoe (approved):

mfencik (approved):

PR #758, opened 2025-03-06
feat: split global/regional deployment

ctria:
Added a new ArgoCD ApplicationSet that includes all the components of understack.

This applicationset includes everything that is included in:
- infra appset
- operators appset
- openstack appset
- components appset

In addition to that it utilizes the understack.rackspace.com/role label to distinguish between a global deployment, a regional and an aio.

The scope of this appset is to NOT include ArgoCD but instead to be able to be included in any ArgoCD deployment and deploy the needed components only on the clusters that have the above label.

You can review such a global cluster deployment at:
https://argocd.pvceng.rax.io/applications?cluster=test-uc-global+%28https%3A%2F%2Ftest-uc-global.k8s-api.pvceng.rax.io%3A443%29

And a regional cluster deployment at:
https://argocd.pvceng.rax.io/applications?cluster=test-uc-regional+%28https%3A%2F%2Ftest-uc-regional.k8s-api.pvceng.rax.io%3A443%29


components/ironic/ironic-ks-user-baremetal.yaml
```diff
@@ -142,7 +142,6 @@ spec:
         - name: PATH
           value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/
         - name: DEPENDENCY_SERVICE
-          value: openstack:keystone-api
```
ctria, line 145: That is not needed as the job depends on the ironic-ks-user already (which depends on the openstack:keystone-api service).

In case of a remote keystone, this service won't exist and we can't change this via helm values in the deploy repo (as we can for the ironic-ks-user job).

components/dex/values.yaml
```diff
@@ -110,3 +110,9 @@ envVars:
       secretKeyRef:
         name: grafana-sso
         key: client-secret
+
+ingress:
```
ctria, line 114: Removed the separate ingress resource definition and add the one from the helm-values as that allows us to override the configuration in the deploy repo.

skrobul (commented):
That looks good on first read (few comments above), but let's test it on actual deployment as discussed

components/dex/values.yaml
cardoe: I fixed up the existing appsets to work with this.

apps/appsets/understack.yaml
```diff
@@ -0,0 +1,897 @@
...
+                            hostname: 'nautobot.{{ .values.uc_dns_zone }}'
+                        valueFiles:
+                          - $understack/components/nautobot/values.yaml
+                          - $deploy/{{.name}}/helm-configs/nautobot.yaml
+                        ignoreMissingValueFiles: true
+                        fileParameters:
+                          - name: nautobot.config
+                            # until ArgoCD 2.13.0 is released we cannot use a reference
+                            # path: $understack/components/nautobot/nautobot_config.py
+                            path: 'https://raw.githubusercontent.com/rackerlabs/understack/{{ .values.uc_repo_ref }}/components/nautobot/nautobot_config.py'
```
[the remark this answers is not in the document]
cardoe: agreed. the existing application has this issue as well.

cardoe (approved):

PR #768, opened 2025-03-13
feat: PUC-783: creating VLAN in Nautobot when segment is created in neutron

haseebsyed12:


python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -257,3 +283,37 @@ def check_vlan_availability(self, interface_id: str | UUID, vlan_tag: int) -> bo
 
         response = self.make_api_request("GET", url, params=params) or {}
         return response.get("available", False)
+
+
+class IpamOperations:
```
cardoe, line 288: So we should just use pynautobot and not write our own http wrapper.

cardoe (changes requested):

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -77,7 +79,42 @@ def update_network_precommit(self, context):
...
+            for segment in context.network_segments
+            if segment.get("network_type") == "vlan"
+            and segment.get("physical_network") is not None
+        ]
+        if not vlan_segments:
+            return
+
+        missing_segments = []
+
+        for segment in vlan_segments:
```
stevekeay, line 94: I don't know how many network segments we expect to see on any given network, but this is O(n) round-trips to the nautobot api, and would scale better if we could somehow fetch all the relevant objects in a single API request.

If that is complicated to do, feel free to ignore this comment because we should not put effort into premature optimisation.

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -77,7 +79,42 @@ def update_network_precommit(self, context):
         pass
 
     def update_network_postcommit(self, context):
-        pass
+        network = context.current
+        vlan_segments = [
```
stevekeay, line 83: Openstack appears to be calling these network_segments, while Nautobot calls them vlans.  Any reason we don't use consistent naming here?

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -257,3 +287,17 @@ def check_vlan_availability(self, interface_id: str | UUID, vlan_tag: int) -> bo
 
         response = self.make_api_request("GET", url, params=params) or {}
         return response.get("available", False)
+
+    def get_vlan(self, segmentation_uuid: str):
```
stevekeay, line 291: Nautobot calls these vlans, not segmentations.  Since this method expresses the nautobot API it should use nautobot terminology.

python/neutron-understack/neutron_understack/neutron_understack_mech.py
cardoe: Yeah we need to use the pynautobot client and fetch these with the `.filter()` method to get the full list in 1 call.

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -26,6 +30,30 @@ class NautobotCustomFieldNotFoundError(exc.NeutronException):
...
+    status: str = "Active"
+
+    def to_dict(self) -> dict:
+        return {
+            "id": self.id,
+            "vid": self.vid,
+            "name": utils.derive_from_network(
+                network_id=self.network_id, network_name=self.network_name
+            ),
+            "vlan_group": {"name": self.vlan_group_name},
```
cardoe, line 49:
```suggestion
            "vlan_group": self.vlan_group_name,
```

They look up `name` and `id` the same way and flatten the dictionary internally in Django REST framework so there's no advantage here to that.

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -26,6 +30,30 @@ class NautobotCustomFieldNotFoundError(exc.NeutronException):
...
+            "status": {"name": self.status},
```
cardoe, line 50:
```suggestion
            "status": self.status,
```

They look up `name` and `id` the same way and flatten the dictionary internally in Django REST framework so there's no advantage here to that.

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -34,6 +62,37 @@ def _truncated(message: str | bytes, maxlen=200) -> str:
     return f"{input[:maxlen]}...{len(input) - maxlen} more chars"
 
 
+class NautobotClient:
```
cardoe, line 65: What's the goal of the wrapper class of the Nautobot client? Because other than the last call everything is a single line

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -34,6 +62,37 @@ def _truncated(message: str | bytes, maxlen=200) -> str:
...
+        return vlan
+
+    def delete_vlans(self, vlan_ids: list[str]):
+        return self.nautobot.ipam.vlans.delete(vlan_ids)
+
+    def create_vlans_and_associate_vlan_to_ucvni(self, vlans: list[VlanPayload]):
+        vlans_payload = [payload.to_dict() for payload in vlans]
+        vlans_endpoint: pynautobot.core.endpoint.Endpoint = self.nautobot.ipam.vlans
+        vlan_create_response = vlans_endpoint.create(vlans_payload)
+        return vlan_create_response
```
cardoe, line 85:
```suggestion
        return self.nautobot.ipam.vlans.create([payload.to_dict() for payload in vlans])
```

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -34,6 +62,37 @@ def _truncated(message: str | bytes, maxlen=200) -> str:
...
+        self.nautobot = pynautobot.api(
+            url=nb_url,
+            token=nb_token,
+        )
+
+    def fetch_vlans(self, vlan_ids: list[str]):
+        vlan: pynautobot.core.response.Record = self.nautobot.ipam.vlans.filter(
+            id=vlan_ids
+        )
+        return vlan
```
cardoe, line 76:
```suggestion
        return self.nautobot.ipam.vlans.filter(id=vlan_ids)
```

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -34,6 +62,37 @@ def _truncated(message: str | bytes, maxlen=200) -> str:
...
+        return self.nautobot.ipam.vlans.delete(vlan_ids)
+
+    def create_vlans_and_associate_vlan_to_ucvni(self, vlans: list[VlanPayload]):
+        vlans_payload = [payload.to_dict() for payload in vlans]
+        vlans_endpoint: pynautobot.core.endpoint.Endpoint = self.nautobot.ipam.vlans
+        vlan_create_response = vlans_endpoint.create(vlans_payload)
+        return vlan_create_response
+
+    def fetch_associated_vlans_ids_from_ucvni(self, ucvni_uuid: str):
+        ucvni: pynautobot.core.response.Record = (
```
cardoe, line 88: This type will be wrong with the next release.

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -81,7 +84,60 @@ def update_network_precommit(self, context):
         pass
 
     def update_network_postcommit(self, context):
-        pass
```
cardoe, line 84: :-1: Don't do this here. Register a function to be called on `resources.SEGMENTS` and `events.AFTER_CREATE` and `events.BEFORE_DELETE` 

Here's an example in the OVN mechanism... https://github.com/openstack/neutron/blob/60c572acb18e3181d93cf6cc60920362f264aa6b/neutron/plugins/ml2/drivers/ovn/mech_driver/mech_driver.py#L624
https://github.com/openstack/neutron/blob/60c572acb18e3181d93cf6cc60920362f264aa6b/neutron/plugins/ml2/drivers/ovn/mech_driver/mech_driver.py#L301

python/neutron-understack/neutron_understack/utils.py
```diff
@@ -6,3 +6,19 @@ def fetch_subport_network_id(subport_id):
...
+    short_uuid = network_id.replace("-", "")[:8]
+    name = f"{network_name}___{short_uuid}"
+    return name
+
+
+def valid_network_segments(network_segments: list[dict]):
+    return [
+        segment
+        for segment in network_segments
+        if segment.get("network_type") == "vlan"
```
cardoe, line 22:
```suggestion
        if segment.get("network_type") == neutron_lib.constants.TYPE_VLAN
```

cardoe (changes requested):

PR #771, opened 2025-03-17
feat: Dynamic network segments for each VLAN Group

stevekeay:
This was opened to address PUC-787

The old scheme used a Nautobot plugin to assign a VLAN number outside of
Openstack, and this is what was configured on the switch.  This caused
issues because Neutron had no knowledge of these VLANs, which make it
difficult to have Neutron drive any layer 3 services or even trunk ports.

Now we have Neutron allocate "Dynamic" network segments instead, and
these are pushed to Nautobot, to keep Nautobot's VLANs in sync with
Neutron's Network Segments.

To facilitate we are using the "physical network" of each Ironic
baremetal port to identify the vlan group where that switch port lives.
For every vlan group we have created a Network Segment Range with the
same name.  Dynamic network segments of type "VLAN" are assigned from
those ranges.

We implement some new API calls to replace the "prep switch interface"
and "detach port" Nautobot jobs which were previously making interface
updates in addition to assigning VLAN numbers.  Specifically, we now
update the vlans on the switch port and we toggle it's "provisioning"
state.

Undersync is invoked with a VLAN Group parameter.  We now identify the
VLAN Group by name instead of UUID.  The name is directly discoverable
from Neutron whereas the UUID would have to be looked up in Nautobot.
(Undersync was already updated to accept either type of parameter.)

In Neutron we now exclusively use type "VXLAN" and so all support for
type "VLAN" Networks has been removed.  This also includes the
"provisioning" network which has been recreated as type VXLAN in our
environment.

There is an opportunity to remove our special "vlan 4010" case handling
of the provisioning network and have this created by Neutron using the
same mechanisms as for as any other tenant network, but that has not
happened yet.

Trunk ports have first-class support in Neutron but they are handled by
a separate model, disjoint from "normal" ports, and using trunk ports
exercises totally different ml2 driver callbacks.   Trunk port support
is coming in a separate future PR.


python/neutron-understack/neutron_understack/vlan_group_name_convention.py
```diff
@@ -0,0 +1,27 @@
...
+    "-2": "network",
+    "-1f": "storage",
+    "-2f": "storage",
+    "-3f": "storage-appliance",
+    "-4f": "storage-appliance",
+    "-1d": "bmc",
+}
+
+
+def for_switch(switch_name: str) -> str | None:
```
cardoe, line 13:
```suggestion
def for_switch(switch_name: str) -> str:
```

if pyright complaints, put a `return ""` after the `raise`

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -283,22 +290,35 @@ def _configure_switchport_on_bind(self, context: PortContext) -> None:
...
+            if segment[api.NETWORK_TYPE] in RELEVANT_NETWORK_TYPES:
                 context.set_binding(
                     segment[api.ID],
                     portbindings.VIF_TYPE_OTHER,
                     {},
                     status=p_const.PORT_STATUS_ACTIVE,
                 )
                 LOG.debug("Bound segment: %s", segment)
+                self._find_or_create_network_segment(context)
                 self._configure_switchport_on_bind(context)
```
cardoe, line 313: Are we calling `.continue_binding()` in here?

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -307,17 +327,42 @@ def bind_port(self, context: PortContext) -> None:
...
+                and segment.get(api.PHYSICAL_NETWORK) == vlan_group_name
+            ):
+                LOG.info("Using existing VLAN network segment %s", segment)
+                return segment
+
+        segment_args = {
+            api.NETWORK_TYPE: p_const.TYPE_VLAN,
+            api.PHYSICAL_NETWORK: vlan_group_name,
+        }
+        segment = context.allocate_dynamic_segment(segment_args)
```
cardoe, line 363: We shouldn't have to do this part. The port will come in with a physical network request that should match a network-segment-range and we'll just bind on that.

python/neutron-understack/neutron_understack/vlan_group_name_convention.py
```diff
@@ -0,0 +1,27 @@
+# Please keep this in sync with data_center.py in this repo:
+VLAN_GROUP_SUFFIXES = {
+    "-1": "network",
+    "-2": "network",
+    "-1f": "storage",
+    "-2f": "storage",
+    "-3f": "storage-appliance",
+    "-4f": "storage-appliance",
+    "-1d": "bmc",
+}
```
cardoe, line 10: Reading through now, we won't need this code because we just need to create the correct network segment ranges for the "network" vlan groups which will match up with Nautobot.

stevekeay:
so... we should not need this PR at all, right?

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -28,7 +28,7 @@ class UnderstackDriver(MechanismDriver):
     resource_provider_uuid5_namespace = UUID("6eae3046-4072-11ef-9bcf-d6be6370a162")
 
     @property
-    def connectivity(self):
+    def connectivity(self):  # type: ignore
```
cardoe, line 31: What was the complaint on this line?

stevekeay: The complaint is that this property incorrectly overrides the superclass.  I think because there is no annotation, and it returns a literal value, the inferred type *is* the literal value.  Therefore the signatures don't match.

python/neutron-understack/neutron_understack/vlan_group_name_convention.py
stevekeay: It seems like we would need to find the baremetal port, perhaps by MAC address, and then look up its physical_network.  This complicates things, so leaving it as a task to clean up later once we get this all working.

cardoe:
Will you include a detailed description in the PR?

PR #790, opened 2025-03-20
feat: Adds workflow and tooling for troubleshooting ironic 'clean failed' events

nicholaskuechler:
feat: Adds workflow and tooling for troubleshooting ironic 'clean failed' events


examples/openstack-notifications/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Find OpenStack Ironic notifications for a specific provision state
+
+I needed to find the notification payload for OpenStack Ironic `clean failed` events:
+
+```
+python queue_dump.py -u $RABBIT_USER -p $RABBIT_PASS --provision-state 'clean failed'
+```
+
+## Drain a queueu and destroy all messages
```
skrobul, line 71:
```suggestion
## Drain a queue and destroy all messages
```

examples/openstack-notifications/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+```
+export RABBIT_USER=$(kubectl -n openstack get secret rabbitmq-default-user -o jsonpath="{.data.username}" | base64 --decode)
+export RABBIT_PASS=$(kubectl -n openstack get secret rabbitmq-default-user -o jsonpath="{.data.password}" | base64 --decode)
+```
+
+3. Open a port to the rabbitmq server:
+
+``` text
+kubectl -n openstack port-forward svc/rabbitmq 5672kubectl -n openstack port-forward svc/rabbitmq 5672
```
skrobul, line 21: duplicated line

examples/openstack-notifications/README.md
nicholaskuechler: Fixed, thanks

examples/openstack-notifications/README.md
nicholaskuechler: Fixed, thanks

skrobul (approved):
The code looks good. I only have one concern - if I understand correctly, this workflow will trigger every time cleaning fails. If there is some hardware issue, this will run in infinite loop - do you think there is an easy way to try N times **and** produce an alert for operator to investigate?

nicholaskuechler:
> The code looks good. I only have one concern - if I understand correctly, this workflow will trigger every time cleaning fails. If there is some hardware issue, this will run in infinite loop - do you think there is an easy way to try N times **and** produce an alert for operator to investigate?

Yes unfortunately it would re-try forever. I haven't experienced this in practice when I've been manually cleaning yet, though. Usually the problem happens during deploys when services are being restarted, so a single retry usually fixes it.

That being said, I think there might be something we can do to prevent infinite retries by using the ironic node history:

``` text
 osa-dev baremetal node history list bfa06d8a-7d2e-4934-aeb5-c9b185b83548 -f json
[
  {
    "UUID": "7426536e-dea3-4706-b589-f13eb55f4bdc",
    "Created At": "2025-02-28T14:37:35+00:00",
    "Severity": "ERROR",
    "Description of the event": "Deploy step deploy.switch_to_tenant_network failed: Error changing node bfa06d8a-7d2e-4934-aeb5-c9b185b83548 to tenant networks after deploy. NetworkError: Could not add public network VIF 4fc36b3c-11ff-4e3d-8009-f08d2fa5deac to node bfa06d8a-7d2e-493..."
  },
  {
    "UUID": "6766ad3c-a404-4cc0-9f63-f7fadb270453",
    "Created At": "2025-03-26T14:41:40+00:00",
    "Severity": "ERROR",
    "Description of the event": "Failed to prepare node bfa06d8a-7d2e-4934-aeb5-c9b185b83548 for cleaning: Unable to establish connection to http://neutron-server.openstack.svc.cluster.local:9696/v2.0/ports/25e79ca0-4961-4c5c-af50-2859b39d225a: ('Connection aborted.', RemoteDisconnec..."
  }
]
```

I'll investigate this further as a follow on.

PR #796, opened 2025-03-24
feat: PUC-785: crud ucvni-group range on network-segment-range events

haseebsyed12:
[tested end to end workflow](https://workflows.dev.undercloud.rackspace.net/workflows/argo-events/neutron-event-network-segment-range-hkbsv?tab=workflow&nodeId=neutron-event-network-segment-range-hkbsv&uid=0e7a2484-6a1c-4e3c-b824-9a68fa40bdfe)

### DIFF
--- a/python/understack-workflows/understack_workflows/main/sync_ucvni_group_range.py
+++ b/python/understack-workflows/understack_workflows/main/sync_ucvni_group_range.py
@@ -1,0 +1,203 @@
+import argparse
+import logging
+from argparse import Namespace
+from enum import StrEnum
+from typing import cast
+
+import requests
+from pynautobot.core.endpoint import DetailEndpoint
+from pynautobot.core.endpoint import Endpoint
+from pynautobot.core.response import Record
+
+from understack_workflows.helpers import credential
+from understack_workflows.helpers import parser_nautobot_args
+from understack_workflows.helpers import setup_logger
+from understack_workflows.nautobot import Nautobot
+
+logger = setup_logger(__name__, level=logging.INFO)
+
+_EXIT_SUCCESS = 0
+_EXIT_API_ERROR = 1
+_EXIT_EVENT_UNKNOWN = 2
+
+
+class SegmentRangeEvent(StrEnum):
+    CREATE = "network_segment_range.create.end"
+    UPDATE = "network_segment_range.update.end"
+    DELETE = "network_segment_range.delete.end"
+
+
+def argument_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description="Handle Network Segment Range Events")
+    parser.add_argument(
+        "event", type=SegmentRangeEvent, choices=list(SegmentRangeEvent)
+    )
+    parser.add_argument("segment_name", type=str, help="Segment name")
+    parser.add_argument("network_type", type=str, help="Network type")
+    parser.add_argument("segment_range_id", type=str, help="Segment range ID")
+    parser.add_argument("segment_min_range", type=int, help="Segment minimum range")
+    parser.add_argument("segment_max_range", type=int, help="Segment maximum range")
+    return parser_nautobot_args(parser)
+
+
+def _remove_subrange(
+    ranges: list[str], segment_range_id: str, notes_endpoint
+) -> list[str]:
+    range_record: Record = cast(Record, notes_endpoint.get(segment_range_id))
+    range_to_remove: str | None = range_record.note
+
+    """Remove sub-range and adjust existing ranges."""
+    if not range_to_remove:
+        return ranges
+
+    updated_ranges = []
+    try:
+        remove_start, remove_end = map(int, range_to_remove.split("-"))
+    except ValueError:
+        return ranges  # Return unchanged if range_to_remove is invalid
+
+    for r in ranges:
+        start, end = map(int, r.split("-"))
+
+        # If it's an exact match, remove it
+        if start == remove_start and end == remove_end:
+            continue
+
+        # If there's no overlap, keep the range
+        if end < remove_start or start > remove_end:
+            updated_ranges.append(r)
+        else:
+            # Handle partial overlap by splitting
+            if start < remove_start:
+                updated_ranges.append(f"{start}-{remove_start - 1}")
+            if end > remove_end:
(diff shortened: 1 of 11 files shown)


python/understack-workflows/understack_workflows/main/sync_ucvni_groups.py
```diff
@@ -0,0 +1,258 @@
...
+    nb_token = segment_args.nautobot_token or credential("nb-token", "token")
+    nautobot_session = requests.Session()
+    nautobot_session.auth = TokenAuth(nb_token)
+
+    ucvni_group_api_url = urljoin(
+        segment_args.nautobot_url, "api/plugins/undercloud-vni/ucvni_groups/"
+    )
+    ucvni_group_details = fetch_ucvni_group(
+        nautobot_session, ucvni_group_api_url, segment_args.segment_name
+    )
```
skrobul, line 204: is there any reason for reimplementing the Nautobot client again? There is already one that works in `nautobot.py`

haseebsyed12: it is using pynautobot and I am facing issue while using pynautobot for ucvni-groups because we have API endpoint like this `api/plugins/undercloud-vni/ucvni_groups` but pynautobot is converting into `api/plugins/undercloud-vni/ucvni-groups`

skrobul: Workaround that can get you started for development:

```
>>> endpoint = nb.plugins.undercloud_vni.ucvni_groups
>>> endpoint.url = endpoint.url.replace('ucvni-groups', 'ucvni_groups')
>>> endpoint.all()
[<pynautobot.core.response.Record ('from_api') at 0x7f80ba6f07d0>, <pynautobot.core.response.Record ('spine402-1.iad3') at 0x7f80ba884a10>]
>>> 
```

but the proper solution would be to fix [urls.py in vni_custom_model](https://github.com/RSS-Engineering/undercloud-rackspace/blob/3a06ec108c636973a0f1802ac2513374f4c17d18/vni-custom-model/vni_custom_model/urls.py#L9) plugin to follow Nautobot's conventions and use `ucvni-group`.

skrobul: I think it's fine to change it - there shouldn't be too many (if any) places using API endpoint for UCVNI groups.

skrobul (changes requested):
Now that https://github.com/RSS-Engineering/undercloud-rackspace/pull/415 has landed, please rework this to use pynautobot to avoid duplication

skrobul (commented):
this code still has multiple references to the `requests` library. There are few places where the code looks for RequestException, while the [pynautobot documentation](https://docs.nautobot.com/projects/pynautobot/en/latest/dev/code_reference/core/query/) states that the unhappy path raises Nautobot specific exceptions. Could you please verify which ones are actually used and adjust if needed?

python/understack-workflows/understack_workflows/main/sync_ucvni_group_range.py
```diff
@@ -0,0 +1,203 @@
...
+    try:
+        nautobot: Nautobot = Nautobot(
+            segment_args.nautobot_url, nb_token, logger=logger
+        )
+        ucvni_group_endpoint: Endpoint = (
+            nautobot.session.plugins.undercloud_vni.ucvni_groups
+        )
+        notes_endpoint: Endpoint = nautobot.session.extras.notes
+    except requests.exceptions.ConnectTimeout as e:
+        logger.error("Network error while connecting to Nautobot: %s", str(e))
```
haseebsyed12, line 170: @skrobul : here I added this to capture nautobot service down scenario. Please suggest if this can be handled using pynautobot exception. I did not find any such.

skrobul: Correct me if I'm wrong but this `except` block is essentially a dead code that can never be reached. Here is why:

1. Your code (correctly) initializes [`Nautobot`](https://github.com/rackerlabs/understack/blob/a9dc8861f97b818f48cddca58df49c042295389b/python/understack-workflows/understack_workflows/nautobot.py#L11) object providing URL, token and logger and accepting other defaults.
2. `Nautobot` looks for a provided `session` and if one isn't, it [creates one](https://github.com/rackerlabs/understack/blob/a9dc8861f97b818f48cddca58df49c042295389b/python/understack-workflows/understack_workflows/nautobot.py#L24) by calling `pynautobot.api`.
3. When `pynautobot.api` initializes, it [creates a requests.Session()](https://github.com/nautobot/pynautobot/blob/0dab87e618076217f8ef5b31d2d9dbd0cc563e0b/pynautobot/core/api.py#L91), again using `requests` library defaults. Unfortunately, these defaults are stupid. All of the requests created with a default session [do not time out](https://docs.python-requests.org/en/latest/user/advanced/#timeouts) and hang around forever.

In summary, with current setup, I don't think the `ConnectTimeout` will ever be raised, but if you managed to get this raised I'd be interested to see the code to replicate this.


Now, in terms of solutions  here - I think it would be reasonable to:
- move the Nautobot related code to `nautobot.py`
- once that's done, look into how we can fix the Connection Timeout handling on connections to Nautobot, not only for this call but also for other ones in `nautobot.py` which seem to suffer from the same problem

haseebsyed12: I tested connection timeout in my local without connecting to `Appgate` and I am able to catch `requests.exceptions.ConnectTimeout`.

Regarding setting timeout I followed this [documentation](https://docs.nautobot.com/projects/pynautobot/en/latest/user/advanced/session/#timeouts). But in my above scenario this is not useful because it's only added after the api() call has already returned in [nautobot.py.](https://github.com/rackerlabs/understack/blob/3fca283a8aa7dae115b3b70e481fb92e3f94c3eb/python/understack-workflows/understack_workflows/nautobot.py#L23)

```python
def api_session(self, url: str, token: str) -> NautobotApi:
       api = pynautobot.api(url, token=token)
       api.http_session.mount(api.base_url, TimeoutHTTPAdapter())
       return api

What do you suggest about my test without connecting to Appgate ?

skrobul (approved):

PR #800, opened 2025-03-24
feat: PUC-765: use segmentation_id generated by neutron as ucvni_id in nautobot

haseebsyed12:
Reason for code clean up related to VLAN type networks is,  we were experimenting with VLAN type networks but we switched back to VXLAN networks and in future also we want to use VXLAN type networks only.

### DIFF
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -128,7 +128,7 @@ class Nautobot:
         project_id: str,
         ucvni_group: str,
         network_name: str,
-        segment_id: int | None = None,
+        segmentation_id: int,
     ) -> dict:
         payload = {
             "id": network_id,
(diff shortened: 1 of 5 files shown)


python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -128,22 +128,21 @@ def ucvni_create(
         project_id: str,
         ucvni_group: str,
         network_name: str,
-        segment_id: int | None = None,
+        segment_id: int,
```
stevekeay, line 131: Consistency is king - please can we call this segmentation_id to match openstack, or ucvni_id to match nautobot?

python/neutron-understack/neutron_understack/neutron_understack_mech.py
```diff
@@ -94,20 +95,6 @@ def create_network_postcommit(self, context):
         )
         self._create_nautobot_namespace(network_id, external)
 
-        if provider_type != p_const.TYPE_VLAN:
```
stevekeay, line 97: If we are not doing this any more, please can we write the reasons for this change in the commit message?  Also, if we are no longer creating those vlans, do we need to remove the corresponding cleanup code in delete_network_postommit?

haseebsyed12: made a separate commit for removal of other network types other than VXLAN

cardoe (approved):

stevekeay (approved):

PR #803, opened 2025-03-25
fix(glance): add job annotations to fix sync issues

cardoe:
Due to the way OpenStack Helm loads their jobs and requires jobs to exist forever and that jobs are immutable, we need to have ArgoCD delete the existing or replace them when it runs a sync to apply any changes otherwise we hit immutable errors.

### DIFF
--- a/components/glance/values.yaml
+++ b/components/glance/values.yaml
@@ -140,3 +140,15 @@ annotations:
       argocd.argoproj.io/hook: Sync
       argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
       argocd.argoproj.io/sync-options: Replace=true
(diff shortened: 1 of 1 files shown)


components/glance/values.yaml
```diff
@@ -140,3 +140,15 @@ annotations:
       argocd.argoproj.io/hook: Sync
       argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
       argocd.argoproj.io/sync-options: Replace=true
+    glance_metadata_defs:
```
skrobul, line 143:
```suggestion
    glance-metadefs-load:
```

skrobul (changes requested):

PR #813, opened 2025-04-01
feat: Adds OpenStack Octavia load balancers

nicholaskuechler:
Adds OpenStack Octavia which providers load balancer services using the OVN driver.


.github/workflows/containers.yaml
```diff
@@ -127,6 +127,7 @@ jobs:
           - name: nova-flavors
           - name: ansible
           - name: understack-tests
+          - name: openvswitch
```
cardoe, line 130: Are we using the `openvswitch` container anymore? Cause I don't see it above in the matrix but its here and there's a Dockerfile for it.

components/octavia/values.yaml
```diff
@@ -0,0 +1,135 @@
+---
+
+
+labels:
+  # (nicholas.kuechler) TODO: remove this label override when
+  # https://review.opendev.org/c/openstack/openstack-helm/+/952722
```
cardoe, line 6: If you bump the chartVersion to one of the 2025.1 charts it should have your change and you can drop this.

components/octavia/values.yaml
```diff
@@ -0,0 +1,135 @@
...
+    external_policy_local: false
+    node_port:
+      enabled: false
+
+conf:
+  octavia:
+    api_settings:
+      enabled_provider_drivers: >-
+        ovn: "The Octavia OVN driver",
+        amphora: "The Octavia Amphora driver"
```
cardoe, line 56: Any reason to enable amphora? We don't have VMs so these would be pretty heavy haproxy nodes.

containers/octavia/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        patch \
+        quilt \
+        libunbound-dev \
+        libunwind-dev \
+        python3-unbound \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+RUN pip install -U ovn-octavia-provider ovsdbapp
```
cardoe, line 15: Do we need any specific version to match the OpenStack version?

containers/openvswitch/Dockerfile
```diff
@@ -0,0 +1,35 @@
+ARG FROM=docker.io/ubuntu:25.04
```
cardoe, line 1: Are we using this container anywhere?

apps/openstack/octavia.yaml
```diff
@@ -0,0 +1,4 @@
+---
+component: octavia
+repoURL: https://tarballs.opendev.org/openstack/openstack-helm
+chartVersion: 2024.2.4+f37dd907f
```
cardoe, line 4: This is where I was suggesting a 2025.1 version

cardoe (commented):
Just some questions.

apps/openstack/octavia.yaml
nicholaskuechler: Bumped it 👍

cardoe (approved):

PR #815, opened 2025-04-04
chore: utilize taskfile to make OpenStack Helm secret gen easier

cardoe:
Rather than the existing bash script which has a few rough edges, attempt to use https://taskfile.dev to provide a possibly friendlier interface to generating the deploy repo data.


taskfiles/openstack/creds.taskfile.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+    generates:
+      - '{{.DEPLOY_NAME}}/manifests/{{.SVC}}/secret-{{.TYPE}}-password.yaml'
+
+    vars:
+      TYPE: '{{ index .MATCH 0 }}'
+      SVC: '{{ index .MATCH 1 }}'
+      BASE_PATH: '{{ if isAbs .UC_DEPLOY }}{{.UC_DEPLOY}}/{{.DEPLOY_NAME}}/manifests/{{.SVC}}{{ else }}{{.USER_WORKING_DIR}}/{{.UC_DEPLOY}}/{{.DEPLOY_NAME}}/manifests/{{.SVC}}{{end}}'
+      CURRPASS:
+        sh: kubectl -n openstack get secret '{{.SVC}}-{{.TYPE}}-password' -o jsonpath='{.data.password}' 2> /dev/null | base64 -d
```
abhimanyu003, line 62: If kubectl fails here we will not get info about it because of `2> /dev/null`
Lets say if someone is having wrong `kube context` selected it will cause them have wrong things generated.

taskfiles/openstack/creds.taskfile.yaml
```diff
@@ -0,0 +1,81 @@
...
+      NEWPASS:
+        sh: dd bs=512 if=/dev/urandom count=1 2>/dev/null | LC_ALL=C tr -dc 'A-Za-z0-9\-' | head -c 32
+      USEPASS: '{{if .CURRPASS}}{{.CURRPASS}}{{else}}{{.NEWPASS}}{{end}}'
+    preconditions:
+      - which minijinja-cli
+      - which kubectl
```
abhimanyu003, line 68: Some cosmetic updates can be done preconditions:
```
    preconditions:
      - sh: command -v minijinja-cli >/dev/null
        msg: "minijinja-cli is required but not installed"
      - sh: command -v kubectl >/dev/null
        msg: "kubectl is required but not installed"
      - sh: "command -v asdf >/dev/null"
        msg: "kubeseal is required but not installed"
```

abhimanyu003 (changes requested):

taskfiles/openstack/creds.taskfile.yaml
cardoe: Thanks included this.

taskfiles/openstack/creds.taskfile.yaml
cardoe: I think this is really something temporary to try and create the same config file with the current data. We previously did it because we had to generate a new file every single time and therefore we needed to load some existing data if you were just replacing one credential. But like you pointed out, that'll be very dangerous. The user should instead just generate the single credential they want to replace since now there are separate tasks.

PR #817, opened 2025-04-07
chore(mariadb): increase max_connections to 1024

skrobul:


cardoe:
Any other tweaks or how do we know what's the right value?

skrobul:
We do not know for sure that it would be sufficient long term, but we know that the default of `150+1` is not enough for production environment where we run multiple replicas of neutron, nova and other openstack services. The 
I am not too worried about getting this value perfectly right at this time because it's a temporary change. Once we figure out how to safely migrate to HA method and adjust applications to leverage multiple nodes, the demand for max_connections will be smaller.

cardoe (approved):

PR #818, opened 2025-04-07
chore: Retrofit type hints to neutron ml2 driver plugin API to improve dx

stevekeay:
Hacking on the ml2 driver plugin feels very uncertain due to the large number of inter-dependencies that the driver code has, and the poorly-documented API with confusing concepts, like the "context" argument means different things in different places.

I don't know if this is a valid approach to documenting the API, but adding type hints or comprehensive documentation upstream seems unlikely to be a quick or easy win, while this gets a bit of autocomplete working with minimal effort.  While these type hints will surely get out of date, I assume that the ml2 API would be evolving quite slowly, so we won't see too much drift.

### DIFF
--- a/python/neutron-understack/neutron_understack/ml2_type_annotations.py
+++ b/python/neutron-understack/neutron_understack/ml2_type_annotations.py
@@ -1,0 +1,164 @@
+from typing import Any
+from typing import Literal
+from typing import NotRequired
+from typing import TypeAlias
+from typing import TypedDict
+
+NetworkTypeName: TypeAlias = Literal["vxlan", "vlan"]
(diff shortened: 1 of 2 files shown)


skrobul (approved):
LGTM, my only minor nitpick is usage of relative imports which [is prohibited in Openstack's style guide](https://docs.openstack.org/hacking/latest/user/hacking.html#imports) but I'm not sure if we need to follow the same guide

cardoe (approved):
Yeah I'm okay with the relative. I still think we should submit this upstream. One of the things they want to do is type annotate more parts of the code base this cycle and you've done that.

cardoe:
@stevekeay merge conflict with another PR of yours

PR #826, opened 2025-04-12
chore(argocd): attempt to get things more inline with the newer layout

cardoe:
The newer layout allows us to split things based on if they are regional or if they are global so we need to get ourselves inline with that change.

Please read the commit messages for the individual changes to understand the differences and the changes.


skrobul (approved):
LGTM

PR #839, opened 2025-04-15
feat: deploy-cli for scaffolding

abhimanyu003:
This is moved from 
https://github.com/RSS-Engineering/undercloud-deploy-cli

All the steps on how to run are README as well.

## How to run 

1. Export all these env
```sh
export UC_DEPLOY="<full_local_path_to_undercloud_deploy_repo>"
export DEPLOY_NAME="<cluster_name>"
export UC_DEPLOY_GIT_URL=git@github.com:RSS-Engineering/undercloud-deploy.git
export UC_DEPLOY_SSH_FILE=<path_to_ssh_private_key_file>
export DNS_ZONE=<cluster_name>.dev.undercloud.rackspace.net
export UC_DEPLOY_EMAIL="<your_email>"
export UC_AIO=yes
```
2. `go run *.go all`
3. `go run *.go help`

> Note: go run *.go is only used during dev if you don't have binary locally, or like to test changes.


go/deploy-cli/cmd/certManager/secrets.go
```diff
@@ -0,0 +1,68 @@
...
+        selector:
+          matchLabels:
+            authorizeWith: http
+      - dns01:
+          webhook:
+            groupName: acme.undercloud.rackspace.net
+            solverName: rackspace
+            config:
+              authSecretRef: cert-manager-webhook-rackspace-creds
+              domainName: dev.undercloud.rackspace.net
```
cardoe, line 56: So there's a few hardcoded values like this. I was also wondering if we can have the templates external and then compiled in so we can lint them somehow?

go/deploy-cli/cmd/argocd/secrets.go
```diff
@@ -0,0 +1,141 @@
...
+data:
+  config: {{ .Config | b64enc }}
+  name: {{ .Name | b64enc }}
+  server: {{ .Server | b64enc }}
+metadata:
+  name: {{ .DEPLOY_NAME }}-cluster
+  namespace: argocd
+  labels:
+    argocd.argoproj.io/secret-type: cluster
+    understack.rackspace.com/argocd: enabled
```
cardoe, line 38: I would drop this from the default

cardoe (commented):
Overall I think it covers what we've got and really shows the gaps of how we want to improve. Looking at this and the diff tool makes me think about some k8s integration for that and bringing it to this as another sub command.

cardoe:
Maybe a "goal" flow of this is...

```bash
# install understack

understack deploy init $DEPLOY_NAME # where it makes `pwd/$DEPLOY_NAME` and populates a directory structure?
```

A lot of those environment variables are a one time thing or not even needed if you're using a separate ArgoCD without baking it into your combined cluster.

skrobul (commented):
what do you think about removing the dependency on the kubeseal binary and using [the package](https://pkg.go.dev/github.com/bitnami-labs/sealed-secrets/pkg/kubeseal) directly?

cardoe:
> what do you think about removing the dependency on the kubeseal binary and using [the package](https://pkg.go.dev/github.com/bitnami-labs/sealed-secrets/pkg/kubeseal) directly?

I honestly don't even want to have kubeseal a requirement but some step that prior to committing your secrets you do something with them.

abhimanyu003:
> what do you think about removing the dependency on the kubeseal binary and using [the package](https://pkg.go.dev/github.com/bitnami-labs/sealed-secrets/pkg/kubeseal) directly?

That is what I was thinking as well, I have to check how it's working as it's having lots of parameter.

go/deploy-cli/.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
+# Make sure to check the documentation at https://goreleaser.com
+
+# The lines below are called `modelines`. See `:help modeline`
+# Feel free to remove those if you don't want/need to use them.
+# yaml-language-server: $schema=https://goreleaser.com/static/schema.json
+# vim: set ts=2 sw=2 tw=0 fo=cnqoj
+
+version: 2
```
cardoe, line 9: Do we need something like:

```
monorepo:
  tag_prefix: prefix-
  dir: go/deploy-cli

File: go/deploy-cli/cmd/certManager/templates/clusterIssuer.tmpl [hunk 4]
```diff
@@ -0,0 +1,27 @@
+---
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: understack-cluster-issuer
+  annotations:
+    argocd.argoproj.io/sync-wave: "5"
+spec:
+  acme:
+    email: undercloud-dev+uc-iad3-dev@rackspace.com
```
cardoe, line 10:
```suggestion
    email: $UC_DEPLOY_EMAIL
```

go/deploy-cli/cmd/certManager/templates/clusterIssuer.tmpl
```diff
@@ -0,0 +1,27 @@
...
+        selector:
+          matchLabels:
+            authorizeWith: http
+      - dns01:
+          webhook:
+            groupName: acme.undercloud.rackspace.net
+            solverName: rackspace
+            config:
+              authSecretRef: cert-manager-webhook-rackspace-creds
+              domainName: dev.undercloud.rackspace.net
```
cardoe, line 27:
```suggestion
              domainName: $DNS_ZONE
```

cardoe (commented):
These were still there from my prior review.

cardoe:
```
❯ go run *.go help
2025/04/23 10:00:13 WARN DNS_ZONE env not set for shell
exit status 1
```

I cannot see the help without having `DNS_ZONE` as an environment variable.

cardoe:
> Maybe a "goal" flow of this is...
> 
> ```shell
> # install understack
> 
> understack deploy init $DEPLOY_NAME # where it makes `pwd/$DEPLOY_NAME` and populates a directory structure?
> ```
> 
> A lot of those environment variables are a one time thing or not even needed if you're using a separate ArgoCD without baking it into your combined cluster.

So I see the name is still the same. Not that `understack` is a good binary name but were you planning on changing that in a follow on?

Further I think requiring the env vars all the time (and the `DNS_ZONE` to even see help) isn't where we want to go. As I've played with this further I think we save some of these values.

```bash
understack deploy [--deploy-path $UC_DEPLOY] init $DEPLOY_NAME
```

Where `--deploy-path` defaults to the value of $UC_DEPLOY and lastly current working directory if unset entirely? It then creates a `$DEPLOY_NAME` directory and inside that directory we can create `config.yaml` with the contents...

```yaml
---
name: $DEPLOY_NAME
```

Then we can have further operations to set `dns_zone` in that file. Then it can be used for all other commands?

cardoe:
So biggest review comment would be to have a more descriptive commit messages in the end.

cardoe (approved):
We'll iterate further in tree.

PR #845, opened 2025-04-16
feat: Accept either name or UUID for switch interface in local_link_info

stevekeay:
We wanted this to become a name rather than a nautobot UUID.  To support the transition from UUID to interface name, we accept a value of either kind.  If the value is not a UUID, we look it up in nautobot.


mfencik (approved):
LGTM

PR #847, opened 2025-04-16
feat:PUC-314: set default boot-interface to http-ipxe

haseebsyed12:
Use HTTP with our UEFI boot. So we need to switch to http-ipxe over ipxe.


cardoe (commented):
Please include a description in the PR.

stevekeay (approved):

PR #848, opened 2025-04-17
feat: etcd backup

skrobul:
This adds daily backups of etcd


skrobul:
> So rather than have to skip the templates what if we have them have a .yaml.tpl extension?

Didn't know that trick, that's certainly better - thank you!

> I also think this shouldn't be in the components appset. This is probably an infra item.

Moved and adapted to the new syntax

cardoe (approved):
This works. Remind me to show you component patches. For small templates it might be easier?

PR #857, opened 2025-04-23
feat: PUC-844: use ansible to create provisioning network

haseebsyed12:
Using Ansible openstack network module to create Provisioning network instead of creating using [openstack cli command](https://github.com/rackerlabs/understack/blob/5841c515f79a333e0afe9d1fe9bfcaa6bcef9c7d/workflows/openstack/sensors/sensor-k8s-neutron-deployment.yaml#L37) for uniformity.

[This PR](https://github.com/RSS-Engineering/undercloud-deploy/pull/348) from understack-deploy repo contains the provisioning network details.


ansible/roles/openstack_network/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  openstack.cloud.networks_info:
+    name: "{{ item.network_name }}"
+  loop: "{{ fabric_wide_layer_3_infrastructure_networks }}"
+  loop_control:
+    label: "{{ item.network_name }}"
+  register: existing_networks
+
+- name: Create network if not exists
+  openstack.cloud.network:
+    name: "{{ item.item.network_name }}"
```
cardoe, line 13: Why the double item.item?

ansible/roles/openstack_network/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+
+- name: Get network info
+  openstack.cloud.networks_info:
```
cardoe, line 4: Can we not just use the network module with state present? I don't think we need the explicitly check. Ansible does that for us.

haseebsyed12: I was getting this error `The following parameters cannot be updated: provider_network_type. You will need to use state: absent and recreate.` so using `openstack.cloud.networks_info` to check it before

ansible/roles/openstack_network/tasks/main.yml
haseebsyed12: `fabric_wide_layer_3_infrastructure_networks` is a list and result of each `openstack.cloud.networks_info` `existing_networks.results` is list again

stevekeay (approved):
I can't see any reason why this won't work.

PR #863, opened 2025-04-24
feat: add ironic client

mfencik:


cardoe (commented):
So my initial comments are that we need to make sure to do this only for bare metal ports for bare metal devices. We're going to have other port types like neutron router ports and the such and that will need to be ignored.

mfencik:
> So my initial comments are that we need to make sure to do this only for bare metal ports for bare metal devices. We're going to have other port types like neutron router ports and the such and that will need to be ignored.

This is now adjusted and it will be refactored when work in https://rackspace.atlassian.net/browse/PUC-904 is done

cardoe (approved):

PR #872, opened 2025-04-24
chore(neutron): replace manual API calls with pynautobot

cardoe:
Use the pynautobot API calls instead of making manual API calls in the neutron mechanism. While typing is a bit awkward and needs some casts currently, it will be better for long term maintainability to use the API library than writing our own calls everywhere. This can be seen by the ability to consistently catch errors and return them back to neutron which should eliminate a few 500 errors that users see without any context.

### DIFF
--- a/python/neutron-understack/neutron_understack/nautobot.py
+++ b/python/neutron-understack/neutron_understack/nautobot.py
@@ -1,23 +1,25 @@
-import inspect
 from dataclasses import dataclass
-from pprint import pformat
-from urllib.parse import urljoin
-from uuid import UUID
+from typing import cast
 
 import pynautobot
-import requests
 from neutron_lib import exceptions as exc
 from oslo_log import log
+from pynautobot.core.response import Record
 
 LOG = log.getLogger(__name__)
 
 
 class NautobotRequestError(exc.NeutronException):
-    message = "Nautobot API ERROR %(code)s for %(url)s %(method)s %(payload)s: %(body)s"
+    message = "Nautobot API ERROR %(code)s for %(url)s %(payload)s: %(body)s"
 
-
-class NautobotOSError(exc.NeutronException):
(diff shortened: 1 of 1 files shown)


python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -266,93 +198,57 @@ def set_port_vlan_associations(
...
-              }
-            }
-        """
-        variables = {"interface_id": interface_uuid}
-        current = self.api.graphql.query(query, variables).json["data"]["interface"]
-        LOG.debug("Nautobot %s query result: %s", variables, current)
-
-        url = f"/api/dcim/interfaces/{interface_uuid}/"
-        payload = {}
+        interface = self.api.dcim.interfaces.get(interface_uuid)
```
cardoe, line 226: wrap

python/neutron-understack/neutron_understack/nautobot.py
```diff
@@ -266,93 +198,57 @@ def set_port_vlan_associations(
...
-        This method checks if a VLAN number, specified as `vlan_tag`, is used
-        in any of the VLAN groups associated with the fabric to which the
-        interface, identified by `interface_id`, is connected.
-        """
-        url = "/api/plugins/undercloud-vni/vlan_availability_check"
-        params = {"interface_id": interface_id, "vlan_tag": str(vlan_tag)}
-
-        response = self.make_api_request("GET", url, params=params) or {}
-        return response.get("available", False)
+        return self.api.dcim.interfaces.update(id=interface_uuid, data=payload)
```
cardoe, line 251: The return type is changing here and this needs to wrap the exception.

cardoe:
A bunch of the `cast()` can go away with https://github.com/nautobot/pynautobot/pull/308

PR #876, opened 2025-04-28
feat: dexop - operator to manage Oauth2 clients in Dex

skrobul:
This pull request introduces a new operator that enables dynamic management of OAuth2 clients in Dex using standard Kubernetes objects.

Currently, creating a Client involves several manual steps:
- Generating a password locally on the developer's machine
- Storing the password in a compatible Kubernetes YAML Secret manifest
- Sealing the Secret on the running cluster, committing it to the repository
- Updating Dex's deployment to read the secret and export it as an environment variable
- Configuring Dex to reference the new environment variable

This process is not ideal, as it imposes a significant administrative burden and security risks due to the use of environment variables. Moreover, updating passwords becomes challenging.

By deploying the operator from this pull request, we can simplify this process significantly. The operator will handle generating secrets and updating Dex configurations for the listed Clients, eliminating the need for manual intervention. Additionally, it leverages dynamic configuration of Dex, allowing us to make changes without requiring a restart. As an added security benefit, the actual password values will never leave the cluster, reducing the risk of exposure and improving overall security.

Key benefits of this change include reduced administrative burden, improved security, and easier password updates.

> [!NOTE]  
To the reviewer:
Part of this PR is boiler plate that is automatically generated by [operator-sdk](https://sdk.operatorframework.io/). This can be ignored or skimmed through:
  - Stuff in `config/` folder is automatically generated with exception of `config/samples`
  - `dist/` contains an autogenerated YAML for non-helm installs
  - `Makefile` is autogenerated with adjustments to the `IMAGE_TAG_BASE` only


abhimanyu003 (changes requested):

go/dexop/.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.23.6
```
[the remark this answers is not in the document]
skrobul: It's from [mise](https://github.com/jdx/mise) which is the same concept, but better. I believe we do need this since go.mod only [lists minimum version](https://go.dev/doc/modules/gomod-ref#go), not the preferred one.

go/dexop/internal/controller/secret_manager.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/sethvargo/go-password/password"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type SecretManager struct {
+}
+
+func (s SecretManager) readSecret(r *ClientReconciler, ctx context.Context, name, namespace string) (string, error) {
```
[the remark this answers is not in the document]
skrobul: The methods in `SecretManager` do not alter the struct's state, and because `SecretManager` is currently an empty struct, there's no need for a pointer receiver. A value receiver suffices for stateless behavior and avoids unnecessary complexity of pointer management.

go/dexop/go.mod
```diff
@@ -0,0 +1,102 @@
+module github.com/rackerlabs/understack/go/dexop
+
+go 1.23
+
+toolchain go1.23.6
+
+require (
```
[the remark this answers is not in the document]
skrobul: Good catch,thanks! fixed in rebase

go/dexop/internal/controller/secret_manager.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if value, ok := secret.Data["client-secret"]; ok {
+		return string(value), nil
+	}
+	return "", fmt.Errorf("client-secret key not found")
+}
+
+func (s SecretManager) writeSecret(r *ClientReconciler, ctx context.Context, clientSpec *dexv1alpha1.Client, value string) (*v1.Secret, error) {
+	secret := &v1.Secret{
+		TypeMeta: metav1.TypeMeta{},
```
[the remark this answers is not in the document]
skrobul: I believe this is automatically inferred from the endpoint by the server, so I have removed it.

cardoe (commented):
Overall seems like what we're aiming to achieve.

go/dexop/dex/client.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", err
+	}
+	return result.GetIssuer(), nil
+}
+
+func newDexClient(hostAndPort, caPath, clientKey, clientCrt string) (dexapi.DexClient, error) {
+	cPool := x509.NewCertPool()
+	caCert, err := os.ReadFile(caPath)
+	if err != nil {
+		return nil, fmt.Errorf("invalid CA crt file: %s", caPath)
```
abhimanyu003, line 107: Slight update here to capture more info about err. 
Added `err` in below code at the end.
```go
return nil, fmt.Errorf("invalid CA crt file (%s): %w", caPath, err)
```

go/dexop/dex/client.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	if !cPool.AppendCertsFromPEM(caCert) {
+		return nil, fmt.Errorf("failed to parse CA crt")
+	}
+
+	clientCert, err := tls.LoadX509KeyPair(clientCrt, clientKey)
+	if err != nil {
+		return nil, fmt.Errorf("invalid client crt file: %s", clientCrt)
```
abhimanyu003, line 115: Same can be done here. :)

go/dexop/internal/controller/client_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	// resource was deleted but it is being finalized
+	if clientSpec == nil {
+		return ctrl.Result{}, nil
+	}
+	reqLogger.Info("Reconciling Client")
+
+	deleteRequested, err := r.handleDeletion(ctx, clientSpec)
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if deleteRequested {
```
abhimanyu003, line 71: nitpick: personally I prefer to multiple if instead of else if.
```go
if err != nil {
	return ctrl.Result{}, err
} 
if deleteRequested {
    return ctrl.Result{}, nil
}
```

go/dexop/internal/controller/client_controller.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if err = ctrl.SetControllerReference(clientSpec, secret, r.Scheme); err != nil {
+		return "", err
+	}
+
+	if err = r.Update(ctx, secret); err != nil {
+		return "", err
+	}
+
+	if secret.Data["secret"] == nil {
```
abhimanyu003, line 213: This can cause panic if secret.Data map is nil or secret key is not present.

we can add additional check, in that way it will first check if map is not not nil
```go
if secret.Data == nil || secret.Data["secret"] == nil {
```

go/dexop/dex/client.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+type DexManager struct {
+	Client dexapi.DexClient
+	Issuer string
+}
+
+// Creates new Oauth2 client in Dex
+func (d *DexManager) CreateOauth2Client(clientSpec *dexv1alpha1.Client) (*dexapi.CreateClientResp, error) {
+	request := &dexapi.CreateClientReq{
```
abhimanyu003, line 24: nitpic: not sure if we do need to check if `clientSpec` is not nill here.

abhimanyu003 (changes requested):

go/dexop/internal/controller/client_controller.go
skrobul: I generally prefer that too, cannot remember why I wrote it this way - fixed as per suggestion, thanks!

go/dexop/dex/client.go
skrobul: done

go/dexop/dex/client.go
skrobul: good point, thx!

go/dexop/internal/controller/client_controller.go
skrobul: good catch, fixed - thx!

go/dexop/dex/client.go
skrobul: it shouldn't be, but input validation is always good so I added it anyway

abhimanyu003 (approved):
LGTM

PR #886, opened 2025-04-29
feat(cli): provide consistent way to define path to deployment repo

cardoe:
Provide a consistent way to define the path to your deployment repo so that all commands can behave the same way.


skrobul (approved):
LGTM

cardoe:
So I handled the lint error and adjusted when the help text is produced to better report to the user what is happening.

abhimanyu003 (approved):

go/understack/cmd/root.go
```diff
@@ -20,10 +34,66 @@ var RootCmd = &cobra.Command{
...
+
 // Execute will execute the root command
 func Execute() error {
 	return RootCmd.Execute()
 }
 
 func init() {
+	// bind our flag
+	if err := viper.BindEnv(deployRepoFlag, deployRepoEnvVar); err != nil {
+		panic(fmt.Sprintf("failed to bind %s env var: %v", deployRepoFlag, err))
```
abhimanyu003, line 85: Instead of panic we can simply do.
```
log.Fatal("failed to bind ", "env", deployRepoFlag, "err", err)
os.Exit(1)
```
output will be like this.

image

go/understack/cmd/root.go
```diff
@@ -20,10 +34,66 @@ var RootCmd = &cobra.Command{
...
+	if !info.IsDir() {
+		return fmt.Errorf("path '%s' is not a directory", expandedRepoDir)
+	}
+
+	// Switch working directory
+	if err := os.Chdir(expandedRepoDir); err != nil {
+		return fmt.Errorf("failed to change working directory to '%s': %w", expandedRepoDir, err)
+	}
+
+	log.Printf("deployment repo path: %s", expandedRepoDir)
```
abhimanyu003, line 72: Can be update to for better print.

```
	log.Infof("deployment repo path: %s", expandedRepoDir)
```

abhimanyu003 (changes requested):

go/understack/cmd/root.go
cardoe: Thanks. Fixed up.

abhimanyu003 (approved):

PR #893, opened 2025-04-30
feat: understack cli github release action

abhimanyu003:
This will build and release understack cli binary once new tag is added.

**How this is work**
* New github tag is added `v0.0.1-understack-cli` ( suffix -understack-cli is important to  identify just the cli release  )
* goreleaser will build binary for mac-linux-windows and attach artifcats to release.

### DIFF
--- a/.github/workflows/understack-cli-release.yaml
+++ b/.github/workflows/understack-cli-release.yaml
@@ -1,0 +1,40 @@
+name: understack-cli-release
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+-understack-cli"
+
+permissions:
+  contents: write
+  packages: write
+  id-token: write
+  issues: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/understack-cli-release.yaml
```diff
@@ -0,0 +1,40 @@
+name: understack-cli-release
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+-understack-cli"
```
cardoe, line 5: @skrobul suffix or prefix? other stuff has used a prefix in the repo. I just wanna be consistent.

abhimanyu003: prefix is now allowed as per semver, if prefix is required need to check how to force goreleaser to use that.

cardoe: There example https://goreleaser.com/customization/monorepo/ uses prefixes. Whatever we do I just want consistency. So if we go with suffix then we switch everything to a suffix.

PR #894, opened 2025-04-30
feat: Adds OpenStack Skyline dashboard to Understack

nicholaskuechler:
Adds the OpenStack Skyline dashboard to Understack. Skyline is a modern version of the OpenStack Horizon dashboard we already have available in Understack.

Screenshot 2025-05-05 at 11 19 16 AM

### DIFF
--- a/components/skyline/values.yaml
+++ b/components/skyline/values.yaml
@@ -1,0 +1,65 @@
+---
+endpoints:
+  skyline:
+    host_fqdn_override:
+      public:
+        tls:
+          secretName: skyline-tls-public
+          issuerRef:
+            name: understack-cluster-issuer
+            kind: ClusterIssuer
+
+network:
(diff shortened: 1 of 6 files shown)


components/skyline/values.yaml
```diff
@@ -0,0 +1,65 @@
+---
+endpoints:
+  skyline:
+    host_fqdn_override:
+      public:
+        tls:
+          secretName: keystone-tls-public
```
cardoe, line 7:
```suggestion
          secretName: skyline-tls-public
```

cardoe (changes requested):
I really wanna suggest that value as a default to the upstream OSH

components/skyline/values.yaml
nicholaskuechler: Fixed

cardoe (approved):

PR #896, opened 2025-04-30
feat(cli): nest commands under 'deploy' and rename 'init' to 'quickstart'

cardoe:
We've discussed that we'll add more than just deployment to this tool so move the existing commands under 'deploy' today so we don't break things. Since the 'init' command does all the operations and is opinionated rename it to 'quickstart'.


skrobul (approved):

cardoe (commented):
+1 from me. I started it so I cannot review it.

PR #916, opened 2025-05-08
feat: argocd: allow operators to control syncPolicy

skrobul:
This is a followup to https://github.com/pvceng-sre/argocd-infra/pull/65

The aim is to allow operators to temporarily pause the automated synchronization for particular apps to facilitate debugging, migrations and similar.

ref: https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Controlling-Resource-Modification/#allow-temporarily-toggling-auto-sync


cardoe (approved):

cardoe:
Let's revisit this with @ctria

PR #929, opened 2025-05-14
feat: "router trunks"

mfencik:


skrobul:
So we jumped on call with @mfencik and after some troubleshooting found the culprit. All of the database queries described above were failing because of the mismatch in the `binding:host_id` field between the subport and the parent port.
The binding levels are actually created with a host_id populated, while the subport keeps it empty. The workaround for this is to determine what the subport's parent port is, obtain it's `host_id` and then query for binding levels.
With those results, we get the `segment_id` that can be used to look up actual segment.

After making those adjustments, we were able to add and remove router interface without leaving anything over in OVN. It all works.

skrobul:
In the interest of keeping everyone updated - the problem with Neutron restart losing ports has been resolved, but during testing we found another seemingly bigger bug.
 
 We can successfully attach and detach network to a router and get routing/NAT services with a single router for tenant A. As per ADR015, we want to use shared PUBLICNET networks for outside connectivity. 
When other tenant (B) adds a second router that shares the same PUBLICNET network, stuff starts to break down.
On addition - we ran into a limitation that prevents segments that have already been added to the trunk to be added again. Milan fixed that one already.
 
Things get more tricky when the tenant A decides to delete their router or detach it from PUBLICNET network because OVN and Neutron simply deletes the port without regard to the fact that tenant B is still very much using PUBLICNET for their router.
 
We have some ideas about how to solve this, but the long story short is that we cannot merge this PR just yet.

PR #932, opened 2025-05-14
feat: Allow node owners and operators to disable cleaning in ironic

nicholaskuechler:
Allows node owners and operators to disable cleaning if they need or want. Currently the request is blocked:
```
 openstack baremetal node set --no-automated-clean 7ca98881-bca5-4c82-9369-66eb36292a95
"baremetal:node:disable_cleaning": "role:admin and system_scope:all" requires a scope of ['system'], request was made with project scope. (HTTP 500)
```

### DIFF
--- a/components/ironic/values.yaml
+++ b/components/ironic/values.yaml
@@ -27,7 +27,10 @@ conf:
   # Update policies for better integration with OpenStack services
   # https://docs.openstack.org/ironic/latest/configuration/sample-policy.html
   policy:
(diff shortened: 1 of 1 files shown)


skrobul (commented):
Why is this needed? I am not sure if giving powers to disable cleaning to `node.lessee` is a good idea. This would potentially put us at risk for accidentally leaking data or failing to execute required maintenance tasks before the node is returned to the pool for next tenant to use.

PR #935, opened 2025-05-15
device_type data provided as volume mount

haseebsyed12:
Providing [nautobot-device-types-repo](https://github.com/RSS-Engineering/undercloud-nautobot-device-types) as volume.

Previously we were doing this way https://github.com/RSS-Engineering/undercloud-rackspace/blob/main/ansible/roles/device_types/tasks/main.yml#L4-L45


skrobul (commented):
Where is this volume mount set up? Also, can we have a customizable path in case someone wants to run same ansible locally, not from the container?

PR #936, opened 2025-05-19
feat: creating neutron objects in openstack using ansible

haseebsyed12:


stevekeay (commented):
Does this need to be different playbooks for neutron/ironic?  Is that a technical limitation in ansible?

It seems like we would want all of this or none of this, I can't see a scenario where we would want to populate ironic but not neutron.

PR #938, opened 2025-05-22
fix: Sync all openstack projects to nautobot, not just the user ones

stevekeay:
This --only-domain parameter prevents projects from being created or updated in nautobot unless they are part of the "default" domain.

This means the infra project is not created in nautobot, which in turn means that out "infra" networks can't be created in nautobot.

We can't just create a user-style project to contain those networks because rbac restrictions prevent it.  So, just allow the infra project to appear in nautobot.  Why not, anyhow?

### DIFF
--- a/python/understack-workflows/tests/test_sync_keystone.py
+++ b/python/understack-workflows/tests/test_sync_keystone.py
@@ -42,7 +42,6 @@ def test_parse_object_id(arg_list, context, expected_id):
         assert args.object == expected_id
 
 
-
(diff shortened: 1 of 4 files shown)


stevekeay:
Given that:
- openstack "domain" acts a namespace for projects
- nautobot tenants are all in a flat namespace
- nautobot requires tenant names to be unique

A better idea seems to be to create tenants in nautobot with names in the form:

    "{domain_name}:{project_name}"

PR #972, opened 2025-06-10
feat: nautobot operator

abhimanyu003:
Closing thi PR in favor of https://github.com/rackerlabs/understack/pull/1057


cardoe:
We'll definitely want a rebase and to squish things together until we get passing.

cardoe (approved):
I think we should squash the commits together since a bunch fail and are just fixups.

Other questions Nautobot is a cluster scoped definition? So we couldn't have a Nautobot in different namespaces to test different ones in the same cluster?

There also seems to be plural vs non-plural in some cases.

PR #973, opened 2025-06-10
feat: run inspection on enroll

haseebsyed12:
this fix also contains patch https://review.opendev.org/c/openstack/ironic/+/951716 for using update_pxe_enabled flag to enable/disable port updation while inspect. In our case we disable it which is done in [https://github.com/RSS-Engineering/undercloud-deploy/pull/350](https://github.com/RSS-Engineering/undercloud-deploy/pull/350)

### DIFF
--- a/components/images-openstack.yaml
+++ b/components/images-openstack.yaml
@@ -22,12 +22,12 @@ images:
     keystone_fernet_setup: "ghcr.io/rackerlabs/understack/keystone:2024.2-ubuntu_jammy"
 
     # ironic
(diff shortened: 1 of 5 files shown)


cardoe (dismissed):
You can change the container references for testing but you need to revert that before merging. The PR will go away once it is merged.

skrobul:
>  You can change the container references for testing but you need to revert that before merging. The PR will go away once it is merged.

@cardoe  https://github.com/rackerlabs/understack/pull/974

PR #974, opened 2025-06-11
feat: check if the pr-XXX images are to be merged

skrobul:
This should stop us from accidentally merging code with reference to the ephermeral container images.

Example failure: https://github.com/rackerlabs/understack/actions/runs/15578200340


stevekeay (approved):
Nice idea.

PR #975, opened 2025-06-11
fix: use admin context when adding router subports

mfencik:


skrobul (approved):
+1 Please update the PR description to say why this was needed (RBAC issues)

PR #1015, opened 2025-06-13
fix(rabbitmq): set TTL to prevent notifications queues from growing indefinitely

nicholaskuechler:
References:
* https://www.rabbitmq.com/docs/ttl
* https://www.rabbitmq.com/kubernetes/operator/using-operator#override
* https://github.com/rabbitmq/cluster-operator/tree/main/docs/examples/import-definitions

### DIFF
--- a/components/glance/glance-rabbitmq-queue.yaml
+++ b/components/glance/glance-rabbitmq-queue.yaml
@@ -57,3 +57,20 @@ spec:
   rabbitmqClusterReference:
     name: rabbitmq  # rabbitmqCluster must exist in the same namespace as this resource
     namespace: openstack
+---
+apiVersion: rabbitmq.com/v1beta1
+kind: Policy
+metadata:
+  name: glance-notifications-ttl
+  namespace: openstack
+spec:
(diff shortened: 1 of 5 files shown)


skrobul (commented):
other resources in these files call out the `metadata->namespace: openstack` explicitly - do we want to set it for the policies too?

components/glance/glance-rabbitmq-queue.yaml
```diff
@@ -57,3 +57,20 @@ spec:
   rabbitmqClusterReference:
     name: rabbitmq  # rabbitmqCluster must exist in the same namespace as this resource
     namespace: openstack
+---
+apiVersion: rabbitmq.com/v1beta1
+kind: Policy
+metadata:
+  name: glance-ttl-policy
```
cardoe:
```suggestion
  name: glance-notifications
```

cardoe (commented):
minor nit so we can have different policies applied and it be obvious change these like this.

cardoe (approved):

PR #1026, opened 2025-06-17
fix(ironic): do not override OpenStack Helm default paths

cardoe:
These paths are defaulted within /var/lib/openstack-helm and should remain that way to avoid cross filesystem operations.

### DIFF
--- a/components/ironic/values.yaml
+++ b/components/ironic/values.yaml
@@ -70,8 +70,6 @@ conf:
     oslo_messaging_rabbit:
       rabbit_ha_queues: true
     pxe:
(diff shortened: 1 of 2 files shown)


cardoe:
> Do we want to update this one too:
> 
> ```
> containers/dnsmasq/dnsmasq/dnsmasq.conf.j2
> 98:tftp-root={{ env['TFTP_DIR'] | default('/var/lib/understack/master_iso_images') }}
> ```
> 
> Also bunch of things in deploy repo still refer to those paths

Good find. Yes to both.

skrobul (approved):

PR #1044, opened 2025-06-27
feat: add deployment of OpenStack Cinder

cardoe:
This adds a deployment of OpenStack Cinder via OpenStack Helm.


cardoe:
This is lacking any manipulation of the conf section. I need someone to take this over and craft some defaults.

PR #1070, opened 2025-07-10
feat: Adds alert automation using argo-events for NeutronAgentDown alerts

nicholaskuechler:
Adds an argo-events sensor and workflow for capturing alerts from AlertManager and running remediation commands to delete a neutron agent if the baremetal node no longer exists in ironic.


cardoe (commented):
This should really live with the workflows piece. This component directory is suppose to be for the deployment of the operator.

cardoe (approved):

PR #1074, opened 2025-07-15
PUC-1010: Netapp Cinder driver for loading cinder cleanly with minimal cinder conf

nidzrai:
[Bootstrapping the driver cleanly with our static config, Just enough to start the cinder volume service successfully](https://rackspace.atlassian.net/browse/PUC-1010)

This commit adds the initial structure for the NetApp Dynamic Cinder driver

The goal of this change is to ensure that the Cinder service loads without any errors when this driver is enabled, even with minimal configuration.

Key points:
- Initializes required methods and capabilities expected by Cinder
- Registers a backend under the `dynamic_netapp` config group
- Accepts metadata from volume type extra_specs ( hostname, username, password, protocol etc)
- Parses these dynamic values to prepare for backend selection
- [WIP]The NetApp REST client is defined but actual volume provisioning and connection logic (e.g., NVMe subsystem creation) is still pending

### DIFF
--- a/python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
+++ b/python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
@@ -1,0 +1,198 @@
+"""NetApp Dynamic SVM Cinder Driver"""
+from oslo_config import cfg
+from oslo_log import log as logging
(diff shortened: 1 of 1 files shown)


cardoe (changes requested):
Can you squash your lint fix commit into your main commit?

cardoe (approved):

PR #1076, opened 2025-07-16
docs(images): Add config drive mandatory option when creating images

nicholaskuechler:


cardoe (approved):
You can set the property on image creation.

PR #1086, opened 2025-07-17
feat: pwsafe external secrets

abhimanyu003:


cardoe:
So this is part of the open source generic project, we won't have any pwsafe references in here. That's why we're using External Secrets so that any backend could be configured by the deploy project.

abhimanyu003:
This is not required any-longer as this is part of internal deployment "deploy" repo. 
But can be used as refrence.

PR #1090, opened 2025-07-19
netapp_nvme with netapp credentials

haseebsyed12:


components/cinder/values.yaml
```diff
@@ -56,9 +60,16 @@ pod:
         volumeMounts:
           - mountPath: /var/lib/cinder
             name: var-lib-cinder
+#          - mountPath: /etc/cinder/conf/backends.conf
```
haseebsyed12, line 63: I cannot mount this `/etc/cinder/conf/backends.conf` because In existing openstack-helm deployment this mountpath exists.

PR #1095, opened 2025-07-23
feat(rally): Add run script for using json config instead of environment variables

nicholaskuechler:
We're able to perform deeper configuration using JSON config format than we're able to do with ENV variable config.


cardoe:
Can you give me an example of how this gets used?

PR #1099, opened 2025-07-24
PUC-1066: Adding entrypoint for cinder volume

haseebsyed12:


python/cinder-understack/pyproject.toml
```diff
@@ -48,3 +49,6 @@ filterwarnings = "ignore::DeprecationWarning"
 # use our default and override anything we need specifically
 extend = "../pyproject.toml"
 target-version = "py310"
+
+[project.entry-points."cinder.volume.drivers"]
+netapp-understack = "cinder_understack.dynamic_netapp_driver:NetappCinderDynamicDriver"
```
cardoe, line 54: I think you won't be able to use a dash here. I think you'll have to name it something like `NetappCinderDynamicDriver`

python/cinder-understack/pyproject.toml
```diff
@@ -24,6 +24,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dependencies = [
+    "cinder>=26.1.0",
```
cardoe, line 27: We're on 2024.2 so I bet we're at 25.x

PR #1103, opened 2025-07-24
fix(nautobot): Performance tuning for nautobot

nicholaskuechler:
This sets Nautobot to use the same limits for all our environments. For celery, I'm also limiting concurrency. We don't run a lot of jobs currently so we maybe need to tune that more in the future.

Concurrency defaults to number of CPUs in the system: https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/guides/celery-queues/

A nautobot issue related to OOM issues: https://github.com/nautobot/helm-charts/issues/433 says "If your concurrency is set too high for the resource requests and limits then you'll have a large amount of memory taken by all the worker processes."

### DIFF
--- a/components/nautobot/values.yaml
+++ b/components/nautobot/values.yaml
@@ -54,6 +54,12 @@ nautobot:
     timeoutSeconds: 60
 
 celery:
+  # Concurrency defaults to the number of CPUs detected on the system.
+  # If your concurrency is set too high for the resource requests and limits then
+  # you'll have a large amount of memory taken by all the worker processes.
+  # https://github.com/nautobot/helm-charts/issues/433
+  # https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/guides/celery-queues/#concurrency-setting
+  concurrency: 2
   replicaCount: 1
   extraEnvVarsSecret:
     - nautobot-django
@@ -66,7 +72,6 @@ celery:
     initialDelaySeconds: 60
     periodSeconds: 120
     timeoutSeconds: 60
-
 
 postgresql:
   enabled: false


components/nautobot/values.yaml
```diff
@@ -46,14 +46,15 @@ nautobot:
       cpu: "1000m"
       memory: "1536M"
     limits:
-      cpu: "2000m"
-      memory: "8704M"
+      cpu: "8000m"
+      memory: "34816M"
```
cardoe, line 50: So my concern here is that this will still result in overloaded nodes. You're telling the k8s scheduler to expect that the pod will only take 1536mb of RAM when it selects a node but then the hard limit is 32GB of RAM. We're really saying that our API nodes should burst up to 32GB of RAM?

nicholaskuechler: @skrobul thought so: https://github.com/RSS-Engineering/undercloud-deploy/blame/main/delta-uc-dfw3-prod/helm-configs/nautobot.yaml#L23-L30

components/nautobot/values.yaml
```diff
@@ -43,8 +43,8 @@ nautobot:
   resources:
     # Increases CPU as the defaults are a little too low for our needs
     requests:
-      cpu: "1000m"
-      memory: "1536M"
+      cpu: "2000m"
```
cardoe, line 46: Is 2 whole cores really correct here? Typically this should be more horizontal pods. 

https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-resource-requests-and-limits

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

components/nautobot/values.yaml
```diff
@@ -43,17 +43,18 @@ nautobot:
   resources:
     # Increases CPU as the defaults are a little too low for our needs
     requests:
-      cpu: "1000m"
-      memory: "1536M"
+      cpu: "6000m"
```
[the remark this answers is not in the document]
nicholaskuechler: Removed it 👍

components/nautobot/values.yaml
nicholaskuechler: Removed it 👍

cardoe:
https://github.com/nautobot/helm-charts/issues/593

PR #1112, opened 2025-07-29
feat: Add Proper Pool Discovery and FlexVol Integration

nidzrai:
https://rackspace.atlassian.net/browse/PUC-1104
This update enhances the NetApp dynamic driver to support real FlexVol discovery and accurate pool capacity reporting, enabling more reliable volume scheduling and placement in OpenStack Cinder.

**Real FlexVol Discovery** -Retrieves actual FlexVol names and capacity details directly from the NetApp backend using the REST API.
**Pool Name Extraction** -Maps FlexVols to Cinder pool names correctly, ensuring volumes are routed to the right pool selected by the scheduler.
**Enhanced Logging** -Adds detailed logs during pool discovery and update operations to aid in debugging and traceability.
**Scheduler Integration**-Exposes real-time capacity and pool stats to the scheduler for accurate volume placement decisions.

**Namespace** - map  namespace after volume 
WIP: subsystem 
WIP : self.client here is temporary . (we have to think of solution to replace it ) [done] used caching via dic {}
Added namespace table to fix  add after volume

### DIFF
--- a/python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
+++ b/python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
@@ -1,53 +1,57 @@
-"""Metadata-based backend config."""
+"""NetApp NVMe driver with dynamic multi-SVM support."""
 
+from cinder import context
 from cinder import exception
+from cinder.objects import volume_type as vol_type_obj
 from cinder.volume import driver as volume_driver
 from cinder.volume.drivers.netapp import options
-from cinder.volume.drivers.netapp.dataontap.block_cmode import (
-    NetAppBlockStorageCmodeLibrary,
-)
 from cinder.volume.drivers.netapp.dataontap.client.client_cmode_rest import (
     RestClient as RestNaServer,
 )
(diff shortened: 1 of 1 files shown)


python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
```diff
@@ -156,67 +504,289 @@ def update_provider_info(self, *args, **kwargs):
...
-            # Add host initiator to subsystem
-            # Create namespace backed by FlexVol
-            # Map namespace to subsystem
-        else:
-            LOG.info(" .WIP. ")
+
+        # Everything else from cinder.conf
+        hostname = self.configuration.netapp_server_hostname
+        username = self.configuration.netapp_login
+        password = self.configuration.netapp_password
```
cardoe, line 645: Cause you're fiddling with the configuration pieces here.

cardoe (approved):
We can merge this to iterate on this against an actual device but we've gotta go back and work on tests before we extend anything any further.

PR #1116, opened 2025-07-30
fix(renovate): simplify the container match rule

cardoe:
This should hopefully simplify the whole match and just support docker container image references in general. Looks like some of the prior things I had been doing are no longer required but are still showing up in searches and results on renovate's site but its all been upstreamed.


mfencik (approved):
LGTM for another try

PR #1174, opened 2025-08-22
fix: yaml template error

stevekeay:
This is a shot in the dark - I simply have no way to test if it will work.  Don't merge unless you're confident it won't break anything.

This value was not being interpolated - those braces are passed verbatim to the output:

kubectl --context=uc_iad3_rxdb_lab --namespace argo get configmaps workflow-controller-configmap -o yaml apiVersion: v1
data:
  sso: |-
    # This is the root URL of the OIDC provider (required).
    issuer: https://dex.{{.dns_zone}}


cardoe:
So I've not merged this in because I've realized that issuer is suppose to be dex and then the other places are the cluster address. Which wouldn't solve the issue you're seeing where they're different.

PR #1191, opened 2025-09-01
fix(cinder-understack): handle SVM name stripped in pool name and bad values passed from nova

skrobul:
The follow error can be seen in the Cinder logs because the SVM name has already been stripped from the pool name in the volume host due to one call calling another driver call. To handle this we will detect this case and detect the SVM name from the project ID and not fail.

```
ERROR cinder_understack.dynamic_netapp_driver [req-f6486c4b-bae2-4c4c-8a33-5ecdca412712 req-4b7cca6a-4421-4f8e-bfec-8858176fb723 c3fd23244abd42631651813496e0ee22bd2377690c1d9b0e7851804e0497775f 6c2fb34446bf4b35b4f1512e51f2303d - - - -] No such SVM vol_6c2fb34446bf4b35b4f1512e51f2303d instantiated
```

The second fix here is a workaround for the fact that nova ironic sets the key "initiator" with the necessary data when the cinder NetApp driver expects the key "nqn" to store the data. This copies "initator" over to "nqn" until upstream can resolve this.

### DIFF
--- a/python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
+++ b/python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
@@ -402,6 +402,9 @@ class NetappCinderDynamicDriver(volume_driver.BaseVD):
 
     def initialize_connection(self, volume, connector):
         """Initialize connection to volume."""
+        # TODO: the nova ironic driver sends the field 'initiator' but the NetApp
+        # cinder driver expects the field to be 'nqn' so copy the field over
+        connector["nqn"] = connector["initiator"]
         with self._volume_to_library(volume) as lib:
             return lib.initialize_connection(volume, connector)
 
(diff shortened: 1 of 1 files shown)


python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
```diff
@@ -328,11 +328,22 @@ def _volume_to_library(self, volume) -> Generator[NetAppMinimalLibrary]:
...
         svm_pool_name = volume_utils.extract_host(original_host, level="pool")
+        if not svm_pool_name:
+            raise exception.InvalidInput(
+                reason=f"svm_pool_name not found in {original_host}"
+            )
+
         svm_name = svm_pool_name.split(_SVM_NAME_DELIM)[0]
+        # workaround when the vol_ name is detected instead of svm name
+        if svm_name.startswith("vol_"):
+            svm_name = f"os-{svm_name[4:]}"
```
cardoe, line 339: There's a config option for what the SVM name should be prefixed with. The check should probably be if the svm_name does not start with that prefix.

nicholaskuechler (approved):

PR #1193, opened 2025-09-02
fix(keystone): avoid resetting keys on each deploy

cardoe:
Due to the way OpenStack Helm sets the helm.sh/hook annotation on the secrets always it causes a removal and re-apply of the secrets but dropping all the generated data every time the keystone chart is deployed. We do not want this behavior so disable OpenStack Helm creating this secret for us and instead switch to just loading the secret ourselves. ref:PUC-1118


cardoe:
https://review.opendev.org/c/openstack/openstack-helm/+/959251 that's an upstream fix that would solve it without needing to merge this

components/keystone/secret-keystone-keys.yaml
```diff
@@ -0,0 +1,16 @@
+# Explicitly define this secret as empty so that OpenStack Helm does not
+# create it for us because it will put helm hook annotations on the one
+# it generates. This causes the secret to get re-generated by subsequent
+# helm runs. Specifically ArgoCD cleans up anything with a helm hook
+# before applying the chart again. We do not want this to go away and
+# instead allow other jobs to update it so it should persist.
```
skrobul, line 6:
```suggestion
# instead allow other jobs to update it so it should persist.
# TODO: remove after https://review.opendev.org/c/openstack/openstack-helm/+/959251 is released.
```

skrobul (approved):

PR #1256, opened 2025-09-22
feat: Automate Nautobot service account provisioning

haseebsyed12:
When service account details are created in Vault (PasswordSafe), a Kubernetes
secret is generated. Argo Events then triggers a Job that runs an Ansible
playbook to ensure the user is created in Nautobot and a corresponding token
is provisioned.

### DIFF
--- a/ansible/roles/users/library/nautobot_token.py
+++ b/ansible/roles/users/library/nautobot_token.py
@@ -2,23 +2,24 @@ from ansible.module_utils.basic import AnsibleModule
 import requests
 
 
-def get_existing_token(base_url, username, password, user_token):
-    """Return the token dict if it exists, otherwise None."""
+def get_existing_token(base_url, username, password, user_token, module):
     headers = {"Accept": "application/json"}
     tokens_url = f"{base_url}/api/users/tokens/"
 
     try:
         response = requests.get(tokens_url, headers=headers, auth=(username, password))
         response.raise_for_status()
-    except requests.exceptions.RequestException:
-        return None
+    except requests.exceptions.RequestException as e:
+        module.fail_json(
+            msg=f"Failed to fetch existing tokens for user {username}: {str(e)}"
+        )
 
     tokens = response.json().get("results", [])
     return next((t for t in tokens if t.get("key") == user_token), None)
 
 
-def create_new_token(base_url, username, password, user_token, description):
-    """Create a new Nautobot token using Basic Auth. Returns the token dict or None."""
+def create_new_token(base_url, username, password, user_token, description, module):
+    """Create a new Nautobot token using Basic Auth."""
     tokens_url = f"{base_url}/api/users/tokens/"
     headers = {"Content-Type": "application/json", "Accept": "application/json"}
     payload = {"key": user_token, "description": description, "write_enabled": True}
@@ -28,24 +29,12 @@ def create_new_token(base_url, username, password, user_token, description):
             tokens_url, headers=headers, json=payload, auth=(username, password)
         )
         response.raise_for_status()
-    except requests.exceptions.RequestException:
-        return None
+    except requests.exceptions.RequestException as e:
+        module.fail_json(
+            msg=f"Failed to create new token for user {username}: {str(e)}"
+        )
 
     return response.json()
-
-
-def format_token_response(token):
-    """Normalize token dict fields for output."""
-    if not token:
-        return None
-    return {
-        "id": str(token.get("id")),
-        "display": str(token.get("display")),
-        "created": str(token.get("created")),
-        "expires": str(token.get("expires")),
-        "write_enabled": bool(token.get("write_enabled")),
-        "description": str(token.get("description", "No description")),
-    }
 
 
 def run_module():
@@ -66,23 +55,25 @@ def run_module():
     token_description = module.params["token_description"]
 
     # fetch existing token
-    token = get_existing_token(base_url, username, password, user_token)
+    token = get_existing_token(base_url, username, password, user_token, module)
     if token:
         module.exit_json(
             changed=False,
-            message=f"Found existing token for {username}",
-            token=format_token_response(token),
+            username=username,
+            message=f"Found existing Nautobot token for user {username}",
         )
 
     # No token found → try creating new
-    new_token = create_new_token(base_url, username, password, user_token, token_description)
+    new_token = create_new_token(
+        base_url, username, password, user_token, token_description, module
+    )
     if not new_token:
-        module.fail_json(msg="Failed to create new token")
+        module.fail_json(msg=f"Failed to create new token for user {username}")
 
     module.exit_json(
         changed=True,
-        message=f"No token found, created new token for {username}",
-        token=format_token_response(new_token),
+        username=username,
+        message=f"No token found, created new Nautobot token for user {username}",
     )
 
 
--- a/components/nautobot/secretstore-nautobot.yaml
+++ b/components/nautobot/secretstore-nautobot.yaml
@@ -29,6 +29,7 @@ rules:
   - ansible-token
   - openstack-token
   - workflow-token
+  - undersync-token
 - apiGroups:
   - authorization.k8s.io
   resources:
--- a/components/undersync/nautobot-token.yaml
+++ b/components/undersync/nautobot-token.yaml
@@ -1,0 +1,28 @@
+---
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
+  name: nautobot-token
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: nautobot
+  target:
+    name: nautobot-token
+    creationPolicy: Owner
+    deletionPolicy: Delete
+    template:
+      engineVersion: v2
+      data:
+        token: "{{ .token }}"
+        bearer_token: "Token {{ .token }}"
+  data:
+  - secretKey: token
+    remoteRef:
+      key: undersync-token
+      property: token
+      # necessary to avoid argoproj/argo-cd#13004
+      conversionStrategy: Default
+      decodingStrategy: None
+      metadataPolicy: None


workflows/argo-events/eventsources/k8s-secret-nautobot-token.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: argoproj.io/v1alpha1
+kind: EventSource
+metadata:
+  name: k8s-secret-nautobot-token
+  namespace: argo-events
```
cardoe, line 5: This should not be in the namespace and shouldn't be part of the argo-events deployment.

workflows/argo-events/kustomization.yaml
```diff
@@ -27,3 +27,7 @@ resources:
   - sensors/alertmanager-webhook-sensor.yaml
   - eventsources/alertmanager-webhook-eventsource.yaml
   - workflowtemplates/alert-automation-neutron-agent-down.yaml
+  # noutobot secret
+  - serviceaccounts/k8s-secret-events-nautobot.yaml
+  - eventsources/k8s-secret-nautobot-token.yaml
+  - sensors/k8s-nautobot-secret.yaml
```
cardoe, line 33: All this should be with nautobot.

workflows/argo-events/sensors/k8s-nautobot-secret.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Sensor
+metadata:
+  name: secret-nautobot-token
+  namespace: argo-events
```
cardoe, line 5: Not in this namespace.

workflows/argo-events/serviceaccounts/k8s-secret-events-nautobot.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: argo-events
```
cardoe, line 5: Not in this namespace.

workflows/argo-events/serviceaccounts/k8s-secret-events-nautobot.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: k8s-events-secret-nautobot
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
cardoe, line 10: This doesn't need to be a cluster role but can just be singular namespace scoped Role

abhimanyu003:
This is looking good, but we need to document the flow using some flow chart + we need to have a table where all the required passwords are listed.

docs/deploy-guide/generate-external-secrets.md
```diff
@@ -0,0 +1,121 @@
...
+  - ref: understack
+    path: 'components/secretstore-gen-secrets'
+    helm:
+      releaseName: nautobot-secrets
+      valueFiles:
+        - $deploy/&#123;&#123;.name&#125;&#125;/helm-configs/secretstore-nautobot-secrets.yaml
+      ignoreMissingValueFiles: true
+  - ref: deploy
+    path: '&#123;&#123;.name&#125;&#125;/manifests/secret-store'
+```
```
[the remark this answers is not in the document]
haseebsyed12: superuser creds of site cluster will be created in aio cluster by ansible in nautobot and k8s secret of that superuser will be created in site cluster. and this super user token will created all other users as usual in both aio and site cluster.

skrobul (commented):
lgtm, please rebase

ansible/roles/users/library/nautobot_token.py
```diff
@@ -0,0 +1,113 @@
...
+    tokens_url = f"{base_url}/api/users/tokens/"
+
+    try:
+        response = requests.get(tokens_url, headers=headers, auth=(username, password))
+        response.raise_for_status()
+    except requests.exceptions.RequestException as e:
+        return None, f"Failed to fetch tokens: {e}"
+
+    data = response.json()
+    tokens = data.get("results", [])
```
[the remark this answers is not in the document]
haseebsyed12: for no tokens
``` json
{
    "count": 0,
    "next": null,
    "previous": null,
    "results": []
}
```

ansible/roles/users/library/nautobot_token.py
```diff
@@ -0,0 +1,113 @@
...
+
+    base_url = module.params["base_url"].rstrip("/")
+    username = module.params["username"]
+    password = module.params["password"]
+    user_token = module.params["user_token"]
+    create_if_notfound = module.params["create_if_notfound"]
+    token_description = module.params["token_description"]
+
+    if module.check_mode:
+        module.exit_json(**result)
```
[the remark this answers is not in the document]
haseebsyed12: removed it. I was [playing](https://gist.github.com/amarao/2b211c342926f4c0f8275840b72ed5bc) with it left mistakenly.

cardoe:
So I'm still struggling to see the deployment configuration here. The questions I've got:

- You'll either need to provide the list of service accounts / tokens that need to be defined in the environment that runs Nautobot (where an admin credential to create that will exist) OR if the individual site environments are defining their own then they'll need the admin credential. This isn't very clear to me from the code which path you intent to take.
- You define a site specific Application deployment of a `kustomization.yaml` into a site namespace `nautobot` but to what purpose? The sites need the token credentials in the namespaces where their applications run.

ansible/roles/users/library/nautobot_token.py
```diff
@@ -0,0 +1,96 @@
...
+def get_existing_token(base_url, username, password, user_token):
+    """Return the token dict if it exists, otherwise None."""
+    headers = {"Accept": "application/json"}
+    tokens_url = f"{base_url}/api/users/tokens/"
+
+    try:
+        response = requests.get(tokens_url, headers=headers, auth=(username, password))
+        response.raise_for_status()
+    except requests.exceptions.RequestException:
+        return None
```
stevekeay, line 14: Maybe log the exception here.

ansible/roles/users/library/nautobot_token.py
```diff
@@ -0,0 +1,96 @@
...
+    headers = {"Content-Type": "application/json", "Accept": "application/json"}
+    payload = {"key": user_token, "description": description, "write_enabled": True}
+
+    try:
+        response = requests.post(
+            tokens_url, headers=headers, json=payload, auth=(username, password)
+        )
+        response.raise_for_status()
+    except requests.exceptions.RequestException:
+        return None
```
stevekeay, line 32: definitely log the exception here.

stevekeay (commented):
Like Doug says, if this is going to run in each site, the credential in passwordsafe would list the "central" nautobot URL, and every site will try to create the user and token in the central nautobot.

Should each site have its own username, like argo.iad3, argo.rxdb-lab, and so on?

haseebsyed12:
> So I'm still struggling to see the deployment configuration here. The questions I've got:
> 
>     * You'll either need to provide the list of service accounts / tokens that need to be defined in the environment that runs Nautobot (where an admin credential to create that will exist) OR if the individual site environments are defining their own then they'll need the admin credential. This isn't very clear to me from the code which path you intent to take.
> 
>     * You define a site specific Application deployment of a `kustomization.yaml` into a site namespace `nautobot` but to what purpose? The sites need the token credentials in the namespaces where their applications run.



- We define superuser username/pwd and token for site in [passwordsafe](https://passwordsafe.corp.rackspace.com/projects/41569/credentials/373539)

- global cluster (staging) creates site cluster (rxdb-lab) [secret of super-user](https://github.com/RSS-Engineering/undercloud-deploy/blob/7852ee79c19b15b4a7d73f13849cc8963ab245d8/charlie-uc-iad3-staging/manifests/nautobot/secret-rxdblab-superuser.yaml) and also creates user and token in nautobot. 

- Now site cluster only [creates secret of super-user](https://github.com/RSS-Engineering/undercloud-deploy/blob/7852ee79c19b15b4a7d73f13849cc8963ab245d8/echo-uc-iad3-rxdb-lab/manifests/nautobot/secret-nautobot-superuser.yaml) it did not do anything in nautobot.

- Now by using site's superuser token. site creates all its other users in nautobot.

haseebsyed12:
> Let's take one example: the secret that a site's argo-workflows is going to use, containing a token allowing it to log in to Nautobot. In which namespace will that be created?
>  Answer: Secret will be created in nautobot namespace and then will be copied to argo-events workflow using [nautobot ClusterSecretStore
](https://github.com/rackerlabs/understack/blob/486e75098f4d7356cccbe2667518a8cedafd2115/workflows/argo-events/secrets/nautobot-token.yaml#L11)> How was that secret previously created? If it is managed by Argo CD, is there an old manifest that we need to delete?
Answer: Previously we were using [nautobot superuser token](https://github.com/rackerlabs/understack/blob/main/workflows/argo-events/secrets/nautobot-token.yaml) only everywhere.
Also I updated the manifest to use [new workflow specific token](https://github.com/rackerlabs/understack/blob/global-secrets/workflows/argo-events/secrets/nautobot-token.yaml#L24) instead of [superuser token](https://github.com/rackerlabs/understack/blob/main/workflows/argo-events/secrets/nautobot-token.yaml#L24)

stevekeay (approved):

ansible/roles/users/library/nautobot_token.py
```diff
@@ -0,0 +1,85 @@
...
+    base_url = module.params["base_url"].rstrip("/")
+    username = module.params["username"]
+    password = module.params["password"]
+    user_token = module.params["user_token"]
+    token_description = module.params["token_description"]
+
+    # fetch existing token
+    token = get_existing_token(base_url, username, password, user_token, module)
+    if token:
+        module.exit_json(
```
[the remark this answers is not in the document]
haseebsyed12: yes
``` python
    def exit_json(self, **kwargs) -> t.NoReturn:
        """ return from the module, without error """
        _skip_stackwalk = True

        self.do_cleanup_files()
        self._return_formatted(kwargs)
        sys.exit(0)

```

PR #1257, opened 2025-09-22
fix: helm chart and path for external-dns

abhimanyu003:
This PR update the chart path for **external-dns**
- Now it's using oci directly in operators.
- Allow setting up values from deploy helm paths.


apps/operators/external-dns.yaml
```diff
@@ -1,5 +1,16 @@
 ---
 component: external-dns
 sources:
-  - ref: understack
-    path: 'operators/external-dns'
+  - repoURL: ghcr.io/rackerlabs/charts
+    path: .
+    chart: external-dns-rackspace
+    targetRevision: 0.0.1
+    includeCRDs: true
```
cardoe, line 8: This line isn't valid. https://argo-cd.readthedocs.io/en/stable/user-guide/helm/

apps/operators/external-dns.yaml
```diff
@@ -1,5 +1,16 @@
 ---
 component: external-dns
 sources:
-  - ref: understack
-    path: 'operators/external-dns'
+  - repoURL: ghcr.io/rackerlabs/charts
+    path: .
```
cardoe, line 5: The `path: .` isn't valid cause you're loading the chart below. You're not extracting an OCI image and then loading that.

cardoe (changes requested):

PR #1307, opened 2025-10-02
feat: upgrade to OpenStack 2025.1

cardoe:
This upgrades every place we use OpenStack 2024.2 (or its components) to OpenStack 2025.1 versions.


nicholaskuechler:
Need to update the container build so that we actually build the 2025.1 containers. We're only building 2024.2 currently: https://github.com/rackerlabs/understack/blob/main/.github/workflows/containers.yaml#L26

nicholaskuechler (approved):

PR #1308, opened 2025-10-02
fix(rally): We no longer need to use a custom user plugin

nicholaskuechler:
The original issue necessitating the use of the `users_custom` plugin was that we had to wait for the project creation sync from keystone to nautobot to complete. This process has been refactored and we no longer need to wait for the sync.


cardoe:
Where is the custom code? Should that be removed as well?

cardoe (approved):

PR #1387, opened 2025-11-10
feat: add a noop deploy interface patch

mfencik:
For infrastructure and any other devices that we do not want Ironic to manage, yet they need to be present in the Ironic inventory, we're using fake drivers when enrolling them. This is causing a problem for conductor. This patch will enable us to use noop deploy-interface with driver set to manual-management, which will save resources lookup and management that Ironic is doing.
backport patch - https://review.opendev.org/c/openstack/ironic/+/964448


cardoe:
We no longer need to patch this in as we've upgraded to a version of Ironic which ships this.

PR #1867, opened 2026-03-25
feat: undersync helm chart

skrobul:
This removes bunch of unnecessary duplication and allows us to control settings in slightly easier way.

Must be deployed together with https://github.com/RSS-Engineering/undercloud-deploy/pull/1618


charts/undersync/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+  type: ClusterIP
+  port: 80
+replicaCount: 1
+image:
+  repository: ghcr.io/rss-engineering/undersync/undersync
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  # tag: ""
+imagePullSecrets:
+  - name: dockerconfigjson-github-com
```
cardoe: I'd like to see this not in the default values and something that the deploy overrides.

skrobul: Fixed and [updated on the other side too](https://github.com/RSS-Engineering/undercloud-deploy/pull/1618/commits/066c1f14f68ffee3b6b725eeb2efce8dcb53fa36)

cardoe (approved):

PR #1908, opened 2026-04-02
feat: enable site clusters to run Nautobot Celery workers on site clusters

haseebsyed12:
Nautobot currently runs entirely on the global cluster, including its Celery workers. Sites that generate heavy background task load have no way to offload that processing closer to where the work originates, and a single global worker pool becomes a bottleneck as sites scale.

This adds a site-scoped ArgoCD Application that deploys only the Celery worker portion of the Nautobot helm chart. The web server, Redis, and PostgreSQL are all disabled because they remain on the global cluster — site workers connect back to those shared services.

This lets operators scale worker capacity per-site independently, run queue-specific workers closer to the hardware they manage, and reduce cross-cluster task latency for site-driven automation.

- ArgoCD Application template (`application-nautobot-worker.yaml`) gated behind `site.nautobot_worker.enabled`
- It deploys only the Celery worker portion of the Nautobot helm chart into the `nautobot` namespace on the site cluster.

### DIFF
--- a/charts/argocd-understack/templates/application-nautobot-worker.yaml
+++ b/charts/argocd-understack/templates/application-nautobot-worker.yaml
@@ -19,7 +19,7 @@ spec:
     helm:
       fileParameters:
       - name: nautobot.config
-        path: $understack/components/nautobot/nautobot_config.py
+        path: {{ $.Values.site.nautobot_worker.nautobot_config }}
       ignoreMissingValueFiles: true
       releaseName: nautobot-worker
       valueFiles:
--- a/charts/argocd-understack/templates/application-nautobot-worker.yaml
+++ b/charts/argocd-understack/templates/application-nautobot-worker.yaml
@@ -1,0 +1,50 @@
+{{- if eq (include "understack.isEnabled" (list $.Values.site "nautobot_worker")) "true" }}
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: {{ printf "%s-%s" $.Release.Name "nautobot-worker" }}
+  finalizers:
+  - resources-finalizer.argocd.argoproj.io
+  annotations:
+    argocd.argoproj.io/compare-options: ServerSideDiff=true,IncludeMutationWebhook=true
+spec:
+  destination:
+    namespace: nautobot
+    server: {{ $.Values.cluster_server }}
+  project: understack
+  sources:
+  - chart: nautobot
+    helm:
+      fileParameters:
+      - name: nautobot.config
+        path: $understack/components/nautobot/nautobot_config.py
+      ignoreMissingValueFiles: true
+      releaseName: nautobot-worker
+      valueFiles:
+      - $understack/components/nautobot-worker/values.yaml
+      - $deploy/{{ include "understack.deploy_path" $ }}/nautobot-worker/values.yaml
+    repoURL: https://nautobot.github.io/helm-charts/
+    targetRevision: 2.5.6
+
+  - path: components/nautobot-worker
+    ref: understack
+    repoURL: {{ include "understack.understack_url" $ }}
+    targetRevision: {{ include "understack.understack_ref" $ }}
+  - path: {{ include "understack.deploy_path" $ }}/nautobot-worker
+    ref: deploy
+    repoURL: {{ include "understack.deploy_url" $ }}
+    targetRevision: {{ include "understack.deploy_ref" $ }}
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    managedNamespaceMetadata:
+      annotations:
+        argocd.argoproj.io/sync-options: Delete=false
+    syncOptions:
+    - CreateNamespace=true
+    - ServerSideApply=true
+    - RespectIgnoreDifferences=true
+    - ApplyOutOfSyncOnly=true
+{{- end }}
--- a/charts/argocd-understack/templates/application-nautobot-worker.yaml
+++ b/charts/argocd-understack/templates/application-nautobot-worker.yaml
@@ -27,8 +27,9 @@ spec:
       - $deploy/{{ include "understack.deploy_path" $ }}/nautobot-worker/values.yaml
       {{- with index $.Values.appLabels "understack.rackspace.com/partition" }}
       values: |
-        celery:
-          taskQueues: {{ . | quote }}
+        workers:
+          default:
+            taskQueues: {{ . | quote }}
       {{- end }}
     repoURL: https://nautobot.github.io/helm-charts/
     targetRevision: 2.5.6
--- a/charts/argocd-understack/templates/application-nautobot-worker.yaml
+++ b/charts/argocd-understack/templates/application-nautobot-worker.yaml
@@ -8,6 +8,7 @@ metadata:
   - resources-finalizer.argocd.argoproj.io
   annotations:
     argocd.argoproj.io/compare-options: ServerSideDiff=true,IncludeMutationWebhook=true
+{{- include "understack.appLabelsBlock" $ | nindent 2 }}
 spec:
   destination:
     namespace: nautobot
@@ -24,6 +25,11 @@ spec:
       valueFiles:
       - $understack/components/nautobot-worker/values.yaml
       - $deploy/{{ include "understack.deploy_path" $ }}/nautobot-worker/values.yaml
+      {{- with index $.Values.appLabels "understack.rackspace.com/partition" }}
+      values: |
+        celery:
+          taskQueues: {{ . | quote }}
+      {{- end }}
     repoURL: https://nautobot.github.io/helm-charts/
     targetRevision: 2.5.6
 
@@ -31,6 +37,15 @@ spec:
     ref: understack
     repoURL: {{ include "understack.understack_url" $ }}
     targetRevision: {{ include "understack.understack_ref" $ }}
+    kustomize:
+      patches:
+        - patch: |
+            - op: replace
+              path: /data/UNDERSTACK_PARTITION
+              value: "{{ index $.Values.appLabels "understack.rackspace.com/partition" | default "" }}"
+          target:
+            kind: ConfigMap
+            name: cluster-data
   - path: {{ include "understack.deploy_path" $ }}/nautobot-worker
     ref: deploy
     repoURL: {{ include "understack.deploy_url" $ }}
--- a/charts/argocd-understack/templates/application-nautobot-worker.yaml
+++ b/charts/argocd-understack/templates/application-nautobot-worker.yaml
@@ -29,6 +29,9 @@ spec:
       values: |
         workers:
           default:
+            enabled: false
+          {{ . }}:
+            enabled: true
             taskQueues: {{ . | quote }}
       {{- end }}
     repoURL: https://nautobot.github.io/helm-charts/
--- a/components/nautobot-worker/kustomization.yaml
+++ b/components/nautobot-worker/kustomization.yaml
@@ -1,0 +1,10 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+configMapGenerator:
+  - name: cluster-data
+    literals:
+      - UNDERSTACK_PARTITION=""
+    options:
+      disableNameSuffixHash: true
--- a/components/nautobot-worker/values.yaml
+++ b/components/nautobot-worker/values.yaml
@@ -1,0 +1,69 @@
+# Nautobot Worker (site-level)
+#
+# Deploys only Celery workers that connect back to the global Nautobot
+# database and Redis. The web server is disabled because it lives on
+# the global cluster. Redis and PostgreSQL are disabled because the
+# workers reach the global instances over the network.
+---
+
+# Disable the Nautobot web server — workers only
+nautobot:
+  enabled: false
+  replicaCount: 0
+
+  db:
+    engine: "django.db.backends.postgresql"
+    # Override in deploy repo values to point at the global CNPG service
+    host: ""
+    port: 5432
+    name: "app"
+    user: "app"
+    existingSecret: "nautobot-db"
+    existingSecretPasswordKey: "password"
+
+  django:
+    existingSecret: nautobot-django
+
+  superUser:
+    enabled: false
+
+  redis:
+    # Override in deploy repo values to point at the global Redis service
+    host: ""
+    port: 6379
+    ssl: false
+    username: ""
+
+celery:
+  enabled: true
+  concurrency: 2
+  replicaCount: 1
+  extraEnvVarsSecret:
+    - nautobot-django
+  livenessProbe:
+    initialDelaySeconds: 60
+    periodSeconds: 120
+    timeoutSeconds: 60
+  readinessProbe:
+    initialDelaySeconds: 60
+    periodSeconds: 120
+    timeoutSeconds: 60
+
+# Disable celery beat — scheduling runs on the global cluster only
+workers:
+  beat:
+    enabled: false
+
+# Do not deploy local Redis — use the global instance
+redis:
+  enabled: false
+
+# Do not deploy local PostgreSQL — use the global CNPG instance
+postgresql:
+  enabled: false
+
+ingress:
+  enabled: false
+
+metrics:
+  enabled: false
--- a/components/nautobot-worker/values.yaml
+++ b/components/nautobot-worker/values.yaml
@@ -38,6 +38,8 @@ celery:
   enabled: true
   concurrency: 2
   replicaCount: 1
+  extraEnvVarsCM:
+    - cluster-data
   extraEnvVarsSecret:
     - nautobot-django
   livenessProbe:
--- a/components/nautobot-worker/values.yaml
+++ b/components/nautobot-worker/values.yaml
@@ -66,6 +66,3 @@ postgresql:
 
 ingress:
   enabled: false
-
-metrics:
-  enabled: false
--- a/components/nautobot/nautobot_config.py
+++ b/components/nautobot/nautobot_config.py
@@ -64,7 +64,7 @@ from nautobot.core.settings_funcs import is_truthy
 if DATABASES["default"]["ENGINE"].endswith("mysql"):  # noqa F405
     DATABASES["default"]["OPTIONS"] = {"charset": "utf8mb4"}  # noqa F405
 
-# SSL/mTLS options for PostgreSQL connections.
+# mTLS options for PostgreSQL connections.
 # When NAUTOBOT_DB_SSLMODE is set to "verify-ca" or "verify-full", the client
 # certificate, key, and CA root cert must be present at the configured paths.
 _db_sslcert = os.getenv("NAUTOBOT_DB_SSLCERT", "/etc/nautobot/mtls/tls.crt")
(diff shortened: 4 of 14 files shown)


components/nautobot-worker/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
```
cardoe, line 2: You shouldn't need this file.

charts/argocd-understack/templates/application-nautobot-worker.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: nautobot.config
+        path: $understack/components/nautobot/nautobot_config.py
+      ignoreMissingValueFiles: true
+      releaseName: nautobot-worker
+      valueFiles:
+      - $understack/components/nautobot-worker/values.yaml
+      - $deploy/{{ include "understack.deploy_path" $ }}/nautobot-worker/values.yaml
+    repoURL: https://nautobot.github.io/helm-charts/
+    targetRevision: 2.5.6
+  - path: components/nautobot-worker
```
cardoe, line 40: Remove this path line and you won't need that kustomization.yaml

haseebsyed12: sorry could you plz provide which line to be removed ?

skrobul: just line 29 (and move the dash to line 30)

charts/argocd-understack/templates/application-nautobot-worker.yaml
```diff
@@ -0,0 +1,50 @@
+{{- if eq (include "understack.isEnabled" (list $.Values.site "nautobot_worker")) "true" }}
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: {{ printf "%s-%s" $.Release.Name "nautobot-worker" }}
+  finalizers:
+  - resources-finalizer.argocd.argoproj.io
+  annotations:
+    argocd.argoproj.io/compare-options: ServerSideDiff=true,IncludeMutationWebhook=true
```
skrobul, line 10: please add appLabels (like in other Applications):

```yaml
{{- include "understack.appLabelsBlock" $ | nindent 2 }}
```

components/nautobot-worker/values.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+# Do not deploy local PostgreSQL — use the global CNPG instance
+postgresql:
+  enabled: false
+
+ingress:
+  enabled: false
+
+metrics:
+  enabled: false
```
skrobul: I think we still want metrics to be exposed for the worker to monitor it correctly.
We don't need the  `metrics.nginxExporter.enabled` and `metrics.uWSGI.enabled` but they are false by default.

components/nautobot/nautobot_config.py
```diff
@@ -64,6 +64,55 @@
 if DATABASES["default"]["ENGINE"].endswith("mysql"):  # noqa F405
     DATABASES["default"]["OPTIONS"] = {"charset": "utf8mb4"}  # noqa F405
 
+# SSL/mTLS options for PostgreSQL connections.
```
skrobul, line 70:
```suggestion
# mTLS options for PostgreSQL connections.
```

SSL is a protocol that has been deprecated in favour of TLS

components/nautobot/nautobot_config.py
```diff
@@ -64,6 +64,55 @@
...
+        "sslcert": _db_sslcert,
+        "sslkey": _db_sslkey,
+        "sslrootcert": _db_sslrootcert,
+    }
+
+# SSL/mTLS options for Redis connections.
+# When NAUTOBOT_REDIS_SSL env var is "true" (set by Helm `nautobot.redis.ssl`),
+# the Helm chart switches the URL scheme to rediss://.  We still need to tell
+# the Python redis client *which* certs to use for mutual TLS.
+import ssl as _ssl  # noqa: E402
```
skrobul:
```suggestion
from ssl import CERT_REQUIRED
```
if you just need one constant, import only that

docs/deploy-guide/components/nautobot-worker.md
```diff
@@ -0,0 +1,547 @@
...
+Site Cluster                          Global Cluster
++------------------+                  +---------------------------+
+| Worker Pod       |  TLS+ClientCert  | Envoy Gateway             |
+|  - celery        | ---------------> |  port 5432 (passthrough)  | --> CNPG PostgreSQL
+|  - mtls certs    | ---------------> |  port 6379 (passthrough)  | --> Redis
++------------------+                  +---------------------------+
+```
+
+The worker pods mount a client certificate (issued by a dedicated
+internal CA via cert-manager) and present it during the TLS handshake.
```
skrobul, line 65: Can we add a sentence with a link to section clarifying where is this CA located? (It's already well explained in ##Certificate infrastructure) so just a link is fine.

docs/deploy-guide/components/nautobot-worker.md
```diff
@@ -0,0 +1,547 @@
...
+defaults pointing to `/etc/nautobot/mtls/`) and sets
+`DATABASES["default"]["OPTIONS"]`. When the env var is unset or empty
+(as on the global cluster), no SSL options are applied and pods connect
+with password-only auth.
+
+#### pg_hba Rule Order
+
+The CNPG `pg_hba` rules are evaluated top-to-bottom:
+
+1. `host all all 10.0.0.0/8 scram-sha-256` -- local pods on the global
```
skrobul: 10.0.0.0/8 covers way more than just a local pods. This should be set to a pod networking value - currently 10.64.48.0/20 in our clusters but it may change in future - ideally this should be configurable

docs/deploy-guide/components/nautobot-worker.md
```diff
@@ -0,0 +1,547 @@
...
+
+- The mTLS CA hierarchy deployed (issuers, root CA, CA issuer)
+- Server TLS certificates for PostgreSQL and Redis
+- A global `nautobot-mtls-client` certificate (for Redis `authClients`)
+- CNPG configured with `serverTLSSecret`, `serverCASecret`, and `pg_hba`
+- Redis TLS enabled with `authClients: true`
+- Envoy Gateway TLS passthrough routes on ports 5432 and 6379
+
+You also need the mTLS CA key pair stored in your external secrets
+provider so the site cluster can pull it.
```
skrobul: Why? The CA private key should never leave the container, not to mention the cluster. Only public key should be distributed if needed to establish trust, not the whole key pair.

docs/deploy-guide/components/nautobot-worker.md
```diff
@@ -0,0 +1,547 @@
...
+  usages:
+    - client auth
+  privateKey:
+    algorithm: ECDSA
+    size: 256
+  issuerRef:
+    name: mtls-ca-issuer
+    kind: Issuer
+```
+
```
skrobul, line 322: I skipped reading those section in detail because the process seems to be fundamentally wrong.
The CA should live only on the global cluster and there is absolutely no need to create a copy of a CA on a site clusters. 
Having CA keys exposed in each of the sites means that it would require only one site to be compromised to give attacker ability to impersonate not only other sites, but also impersonate a server/global. It also allows to perform MITM attacks very easily.  It is like giving every branch office the key-making machine to the company's building badges, nut just a badge for that office.


Normally the process can be:

1. Provision CA and ClusterIssuer on a global cluster
2. When you need to deploy the site, you create a relevant Certificate object on a global cluster so that it can be issued.
3. The issued certificate and key gets transferred to a site-level (possibly through a secrets provider)
4. Workloads on the site use that certificate.

The alternative process could be similar and slightly more automated:
1. Provision CA and ClusterIssuer on a global cluster
2. When you need to deploy the site components, the site-cluster creates a [`CertificateRequest`](https://cert-manager.io/docs/usage/certificaterequest/).
3. It kicks off periodic checks if the certificate has been issued (as indicated by `Ready` condition
4. Operator approves the certificate issuance on a global
5. The site-level automation downloads the certificate and places it in appropriate paths, then sleeps until a certificate expiration is coming up and repeats everything from step 2 again.
6. Workloads on the site use that certificate.


This would be very similar to how kube-apiserver CA and kubelet certificate management works today, just with cert-manager.

I understand this may be complicated so the fully automated rotation can be implemented as a separate PR, but if you choose to go down this route at minimum, we would need to:
- switch to a 3 year client certificate expiration time
- explicitly document that transfer of certificates happens through manual operator intervention and is required every [EXPIRATION_PERIOD] years for each cluster. This can quickly grow into big maintenance burden if it's 2 certs per each site.

components/nautobot/nautobot_config.py
skrobul, line 1: as @cardoe  found out - this introduces 3rd place or 3rd copy of the `nautobot_config.py`:
- the default container already has `/opt/nautobot/nautobot_config.py`
- our custom container build adds `/opt/nautobot_config/nautobot_config.py`
- this PR adds another one that is provided through helm.file

Can we standardise on using just one way of delivering that config?
Ideally I think that should just be a volume that mounts to a Nautobot default's path and allows to be changed on per-env basis.

skrobul (changes requested):

docs/deploy-guide/components/nautobot-worker.md
skrobul: Actually after thinking about this for a while I realised that this is not entirely accurate. When the connections reach postgres pod, the source IP will not be preserved due to the way EnvoyGateway works. All of the connections that are going through TLSRoute will be sourced from an Envoy Gateway's pod IP address (172.28.x.x in dev-global for example). It may be different in other clusters. The same point remains, we should limit this to just pod networking.

components/nautobot-worker/kustomization.yaml
haseebsyed12: It now contains a configMapGenerator for the cluster-data ConfigMap that the ArgoCD Application patches with the site's UNDERSTACK_PARTITION value.

charts/argocd-understack/templates/application-nautobot-worker.yaml
haseebsyed12: Done -- appLabelsBlock added

components/nautobot/nautobot_config.py
haseebsyed12: Done

components/nautobot/nautobot_config.py
haseebsyed12: Done

docs/deploy-guide/components/nautobot-worker.md
haseebsyed12: Done

docs/deploy-guide/components/nautobot-worker.md
haseebsyed12: Implemented as suggested 

- CA hierarchy (selfsigned issuer, root CA, CA issuer) lives on the global cluster only. 
- site clusters pull only the pre-issued client cert+key and the CA public cert via ExternalSecret -- no CA private key, no cert-manager Issuer on site clusters.
- per-site client certificates are issued on the global cluster by cert-manager (e.g. certificate-nautobot-mtls-client-rax-dev-iad3.yaml), then the issued cert is uploaded to the secrets provider. (Automation of this manual coping this cert is not implemented yet).

components/nautobot/nautobot_config.py
haseebsyed12: global [application-nautobot.yaml](https://github.com/rackerlabs/understack/blob/main/charts/argocd-understack/templates/application-nautobot.yaml#L22) and the site application-nautobot-worker.yaml reference the same single `nautobot_config.py` from `$understack/components/nautobot/nautobot_config.py` via the Helm chart's fileParameters mechanism.

It is hardcoded to `/opt/nautobot/nautobot_config.py`
https://github.com/nautobot/helm-charts/blob/develop/charts/nautobot/templates/celery-deployment.yaml#L160
https://github.com/nautobot/helm-charts/blob/develop/charts/nautobot/templates/nautobot-deployment.yaml#L144

shall I address issue of default container path(/opt/nautobot/nautobot_config.py)
and the custom container build path(/opt/nautobot_config//nautobot_config.py) in another PR ?

components/nautobot-worker/values.yaml
haseebsyed12: copied the same from [disabled in favor of explicit definitions through kustomize](https://github.com/rackerlabs/understack/blob/main/components/nautobot/values.yaml#L94)

docs/deploy-guide/components/nautobot-worker.md
haseebsyed12: shall we use TLS for everything removing `host all all 10.0.0.0/8 scram-sha-256` rule ?

components/nautobot/nautobot_config.py
skrobul: Nope, it has to be fixed before merge - switching the global cluster to the `site-nautobot-worker` branch results in missing SSO and lack of installed plugins:

image

docs/deploy-guide/components/nautobot-worker.md
skrobul: Sounds like a good idea, but verify that the liveness/readiness probes still work after that (I have not checked but sometimes app have them configured with plaintext connection string rather than proper app config). It's probably okay, just check.

components/nautobot-worker/values.yaml
skrobul: ahh right, I completely forgot about those - my bad! This will be resolved after we upgrade to the newer chart that allows configuring `scrapeProtocols`.

docs/deploy-guide/components/nautobot-worker.md
skrobul: Thanks, that's looking much better now. 
I am okay with not having the automation for this yet, but if we are leaving this as manual work, can we please:

1. Get the detailed documentation (in Operator Guide) on what the process of adding new site (already there) and renewing certificates for existing site is?
2. The current client certificate lifetime is 1 year:

```
nautobot@nautobot-default-bddd598b9-j82hf:~$ openssl x509 -in /etc/nautobot/mtls/tls.crt -noout -text | grep Not
            Not Before: Apr 14 12:58:22 2026 GMT
            Not After : Apr 14 12:58:22 2027 GMT
nautobot@nautobot-default-bddd598b9-j82hf:~$
```

Can we bump this to 2 or 3 years to give us more breathing room, especially in rapid growth period where we expect many sites to be deployed in a short span of time?

skrobul (commented):
left few comments inline

components/nautobot/nautobot_config.py
haseebsyed12: fixed both the issues and updated the docs with relevant details on root cause and fix

docs/deploy-guide/components/nautobot-worker.md
haseebsyed12: Tested everything is working. Now pb_hba contains only `- hostssl all all 0.0.0.0/0 cert`. Both local and remote authenticate via certs.

docs/deploy-guide/components/nautobot-worker.md
```diff
@@ -185,88 +194,138 @@ Before starting, ensure the global cluster already has:
...
+```yaml title="global-cluster/nautobot/certificate-nautobot-mtls-client-<site>.yaml"
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: nautobot-mtls-client-<site>
+  namespace: nautobot
+spec:
+  secretName: nautobot-mtls-client-<site>
+  duration: 8760h    # 1 year
```
skrobul: as per other comment, can we make it 3 years, please?

docs/deploy-guide/components/nautobot-worker.md
```diff
@@ -185,88 +194,138 @@ Before starting, ensure the global cluster already has:
...
+      rewrite:
+        - regexp:
+            source: "(.*)"
+            target: "client_$1"
+    - extract:
+        key: "<ca-cert-credential-id>"
+      rewrite:
+        - regexp:
+            source: "(.*)"
+            target: "ca_$1"
```
skrobul: This expects very specific type of key/certificate and can be fragile.
Consider switching from a freeform text field to a JSON.

For [example](https://passwordsafe.corp.rackspace.com/projects/41569/credentials/602777), if you store the credentials as JSON:

```json
{
"tls.crt":"LS0tLS1CRUdJTiBDRVJ...",
"tls.key":"LS0tLS1CRUdJTiBFQyB...."
}
```

then you can use simpler and more robust template that does not need regular expressions:

```yaml
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: nautobot-mtls-client-marektest
  namespace: nautobot
spec:
  dataFrom:
  - extract:
      key: "602777"
  refreshInterval: 1h
  secretStoreRef:
    kind: ClusterSecretStore
    name: pwsafe
  target:
    creationPolicy: Owner
    deletionPolicy: Retain
    template:
      data:
        tls.crt: '{{ index (.password | fromJson) "tls.crt" | b64dec }}'
        tls.key: '{{ index (.password | fromJson) "tls.key" | b64dec }}'
      engineVersion: v2
      metadata: {}
      type: kubernetes.io/tls
```

Alternatively, if you really want to store those as plaintext, please consider using [PEM block filters](https://external-secrets.io/latest/guides/templating/#filter-pem-blocks) provided by ESO instead of regexes.

docs/operator-guide/nautobot-celery-queues.md
```diff
@@ -0,0 +1,260 @@
...
+## Verifying Workers are Listening
+
+To confirm a site worker is consuming from the correct queue:
+
+```bash
+# Check the CELERY_TASK_QUEUES env var in the running pod
+kubectl get deploy -n nautobot \
+  -l app.kubernetes.io/component=nautobot-celery-rax-dev \
+  -o jsonpath='{.items[0].spec.template.spec.containers[0].env}' \
+  | python3 -m json.tool | grep -A1 CELERY_TASK_QUEUES
```
skrobul:
```suggestion
kubectl -n nautobot get deploy nautobot-worker-celery-rax-dev \
            -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="CELERY_TASK_QUEUES")].value}'
```

docs/operator-guide/nautobot-mtls-certificate-renewal.md
```diff
@@ -0,0 +1,135 @@
...
+
+Check certificate status on the global cluster:
+
+```bash
+# List all mTLS client certificates and their expiry
+kubectl get certificate -n nautobot -o custom-columns=\
+NAME:.metadata.name,\
+READY:.status.conditions[0].status,\
+EXPIRY:.status.notAfter,\
+RENEWAL:.status.renewalTime
```
skrobul: Consider quoting the value of custom-columns, this does not paste well into zsh

docs/operator-guide/nautobot.md
```diff
@@ -1,5 +1,429 @@
...
+
+If the CNPG pods have not picked up updated certificate secrets (e.g.
+`client-ca.crt` still shows the old CA), restart them one at a time:
+
+```bash
+kubectl delete pod -n nautobot nautobot-cluster-2
+# wait for ready
+kubectl delete pod -n nautobot nautobot-cluster-3
+# wait for ready
+kubectl delete pod -n nautobot nautobot-cluster-1
```
skrobul: `kubectl cnpg restart` and `kubectl cnpg reload` will handle this more gracefully (the pods are not always labelled -1-2-3 for example)

docs/operator-guide/nautobot.md
```diff
@@ -1,5 +1,429 @@
...
+    _name = os.path.splitext(os.path.basename(_path))[0]
+    with open(_path) as _f:
+        PLUGINS_CONFIG[_name] = _interpolate_env(json.load(_f))
+```
+
+This gives each plugin its own readable file, makes PRs easy to review,
+and keeps the `${ENV_VAR}` interpolation for secrets. It can be
+implemented alongside the current env var approach without breaking
+existing deployments.
+
```
skrobul, line 432: good idea, but let's keep this PR focused on the task at hand. This can be done as a follow up

skrobul (commented):
This looks almost ready, added few comments in another review

docs/deploy-guide/components/nautobot-worker.md
haseebsyed12: done. using PEM block filters

skrobul: Nice, thanks. If you want to eliminate the regex and simplify the `dataFrom` section, sourcing data from [multiple credentials like here](https://github.com/RSS-Engineering/undercloud-deploy/blob/main/delta-uc-dfw3-prod/nautobot-api-tokens/secret-dfw3-workflow.yaml) is possible too, just need to assign `secretKey` for each remoteRef.

It's okay to leave it as is - your call.

skrobul (approved):

PR #1909, opened 2026-04-02
fix: Rework enrol workflow

stevekeay:
This is somewhat of an opinionated move from a large, complex workflow
that calls lots of different python scripts to a minimal workflow that
executes one single script to take all required steps.

In theory there are benefits to the multiple workflow steps, but in
practice I don't think we really benefit from the hardcore argo
functionality, and the workflows are fragile, difficult to read, almost
impossible to test, inscrutable to troubleshoot.  By contrast the
python script is easily maintainable.

The python script is "beefed up" with extra capabilities and it now
handles the transitioning of the node through the various stages,
including configuring RAID, running inspection, etc.

Logging and error handling is improved.

Ability to cope with non-standard cabling is improved: we used
to rely completely on a cabling convention to determine which port
should be used for PXE.

We now use the LLDP data reported by the BMC (when available - note that
we have hardware where the LLDP feature does not work), and we enable HTTP
boot on ALL interfaces that the BMC reports to be connected.   The server
will attempt each one in turn until it gets a DHCP response.

We persist the set of PXE interfaces in the extra field of the baremetal
node.  This is then used by the inspection hook to set the "pxe" flag on
those same interfaces.


workflows/argo-events/workflowtemplates/enroll-server.yaml
```diff
@@ -36,77 +36,19 @@ spec:
...
-                  value: "provide"
-                - name: device_id
-                  value: "{{steps.enroll-server.outputs.result}}"
-            when: "{{steps.server-manage-state.outputs.result}} == manageable"
     - name: enroll-server
       container:
         image: ghcr.io/rackerlabs/understack/ironic-nautobot-client:latest
         command:
-          - enroll-server
+          - enrol-server
```
skrobul: The package only [exposes](https://github.com/rackerlabs/understack/blob/87bf1faf6e6ecbb3f1f7141a83b103f9e0b63fec/python/understack-workflows/pyproject.toml#L35) `enroll-server` so this will fail

skrobul (changes requested):

workflows/argo-events/workflowtemplates/enroll-server.yaml
stevekeay: Good point, fixed in 0aaf487

cardoe (commented):
So overall I'm down for the part that gets rid of the argo-workflow and calls it with Python. I'm -1 on all the Redfish API calls because we should be doing this via Ironic. The logic on retrying and rebooting still doesn't cover all the behaviors of the hardware that have been tested via Ironic but there's no point in adding it here.

stevekeay:
> So overall I'm down for the part that gets rid of the argo-workflow and calls it with Python. I'm -1 on all the Redfish API calls because we should be doing this via Ironic. The logic on retrying and rebooting still doesn't cover all the behaviors of the hardware that have been tested via Ironic but there's no point in adding it here.

Note that this PR doesn't add that redfish interaction - it is already happening, I'm mostly just refactoring.

To move in the openstack direction requires significant changes that deserve careful planning and testing.  I'm happy to take this on, but I am concerned that we are postponing delivery of a very time-sensitive capability, notably the ability to on-board the RXDB gear.  The way they cabled up those racks we have no way to enrol them.

I would therefore suggest the following as follow-on PRs:

I don't think we can easily remove all the redfish interaction without adding significant new features to Ironic.  Some of it should be do-able though:

1) setting the password (and other BMC settings) is probably something that needs to stay in enroll, since Ironic expects the credentials to work right away.

2) ironic redfish inspection is probably sufficient to handle the "discovery" part, including enumerating the interfaces on the box

3) setting the BIOS settings can likely be done in a clean step which is defined using the data from (2)

4) Alan's raid thing can probably become a step or a hook

5) We discussed previously - some of this should be happening on every clean, like the RAID and BIOS needs to be ship-shape and we can set a BIOS password but we can not stop them messing with RAID.

I think we still need the external orchestration because it's not obvious how to do all this glue via steps/runbooks/hooks/etc.

skrobul:
Something to explore in future (not for this PR) https://www.dell.com/support/manuals/en-us/poweredge-r7615/idrac9_7.xx_ug/configuring-servers-and-server-components-using-auto-config?guid=guid-d38bd838-ad66-414b-a61f-0cb241f6459e&lang=en-us

stevekeay:
> Something to explore in future (not for this PR) https://www.dell.com/support/manuals/en-us/poweredge-r7615/idrac9_7.xx_ug/configuring-servers-and-server-components-using-auto-config?guid=guid-d38bd838-ad66-414b-a61f-0cb241f6459e&lang=en-us

Yes this is an interesting feature (although I'd put money on it not working properly).   It is irksome that they chose the "golden" model, like you configure a server how you like it and then "save" that config to a file.  I'd prefer if they documented the options/schema for the file to support us building it ourselves, in true declarative style.  It requires a file tailored to the exact hardware (NICs and raid/disks, etc) so we still need the same kind of logic to support this.

Autodiscovery is designed as a one-shot initial setup.  I guess you could maybe reset the iDRAC and get it to re-configure the server but there is no real support for managing the server during its lifetime.

skrobul:
Yeah, @cardoe in the other chat said that the server profiles are also iDRAC firmware version dependent so probably not a great use for us.

python/understack-workflows/understack_workflows/pxe_port_heuristic.py
```diff
@@ -1,55 +1,64 @@
...
-
-    Finally, score the interface name according to the list above.
-    """
-    is_eligible = not any(x in interface.name.upper() for x in DISQUALIFIED)
 
-    link_detected = interface.remote_switch_port_name is not None
+def guess_pxe_interfaces(
+    device_info: ChassisInfo, pxe_switch_macs: set[str] | None = None
+) -> list[str]:
+    """First 8 interface names, most probable PXE interfaces first."""
```
skrobul, line 15: Off by one - the code returns first 7 elements

python/understack-workflows/tests/test_enroll_server.py
skrobul, line 1: This file will not be run in a CI (needs to be called `test_XXX.py`) - is that intentional?

skrobul (commented):
`guess_pxe_interfaces()` now hard-filters disqualified names instead of deprioritizing them, so it can return an empty list.
That empty list is saved to extra.enrolled_pxe_ports during enroll, and later PortBiosNameHook treats an empty list as “no PXE interfaces” (not as fallback), which can disable PXE on all ports - I was wondering if maybe it's better to bail out and fail the enrollment earlier?

skrobul (approved):
Feel free to include above nits in followup PRs if you want to merge earlier

python/understack-workflows/understack_workflows/pxe_port_heuristic.py
stevekeay: Those damned python ranges get me every time.

python/understack-workflows/tests/test_enroll_server.py
stevekeay: Not intentional - chatgpt should have known about that when it created the file!

stevekeay:
> `guess_pxe_interfaces()` now hard-filters disqualified names instead of deprioritizing them, so it can return an empty list. That empty list is saved to extra.enrolled_pxe_ports during enroll, and later PortBiosNameHook treats an empty list as “no PXE interfaces” (not as fallback), which can disable PXE on all ports - I was wondering if maybe it's better to bail out and fail the enrollment earlier?

This is a good catch - that would be a box with no NICs but I changed the hook to treat the empty list as "no data" and it will now use the fallback behaviour in that case.

skrobul (approved):

PR #1917, opened 2026-04-07
feat(argocd-understack): create split rook-ceph Applications

cardoe:
This allows us to install the operator and upgrade the operator
independently from the cluster as well as ensuring that the cluster
syncs after the operator does. This does not remove the existing
deployment to give us an opportunity to transition over.

### DIFF
--- a/charts/argocd-understack/templates/application-rook-ceph-cluster.yaml
+++ b/charts/argocd-understack/templates/application-rook-ceph-cluster.yaml
@@ -1,0 +1,53 @@
+{{- $installApp := eq (include "understack.componentOptionAny" (list $ "rook_ceph_cluster" "installApp" true)) "true" }}
+{{- $installConfigs := eq (include "understack.componentOptionAny" (list $ "rook_ceph_cluster" "installConfigs" false)) "true" }}
+{{- if or $installApp $installConfigs }}
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: {{ printf "%s-%s" $.Release.Name "rook-ceph-cluster" }}
+  finalizers:
+  - resources-finalizer.argocd.argoproj.io
+  annotations:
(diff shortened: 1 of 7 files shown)


skrobul (approved):
LGTM.

I was thinking if we should add some validation to avoid enabling both the "old" and "new" style at the same time:

Something along those lines (untested)

```go
{{- $legacyRookEnabled := or
      (eq (include "understack.isEnabled" (list $.Values.global "rook")) "true")
      (eq (include "understack.isEnabled" (list $.Values.site "rook")) "true")
-}}
{{- $splitOperatorEnabled := or
      (eq (include "understack.componentOptionAny" (list $ "rook_ceph_operator" "installApp" false)) "true")
      (eq (include "understack.componentOptionAny" (list $ "rook_ceph_operator" "installConfigs" false)) "true")
-}}
{{- $splitClusterEnabled := or
      (eq (include "understack.componentOptionAny" (list $ "rook_ceph_cluster" "installApp" false)) "true")
      (eq (include "understack.componentOptionAny" (list $ "rook_ceph_cluster" "installConfigs" false)) "true")
-}}
{{- $splitAnyEnabled := or $splitOperatorEnabled $splitClusterEnabled -}}
{{- if and $legacyRookEnabled $splitAnyEnabled -}}
{{- fail "invalid rook config: legacy `rook.enabled` cannot be used with split `rook_ceph_operator`/`rook_ceph_cluster`; disable one mode before enabling the other" -}}
{{- end -}}
{{- if and $splitClusterEnabled (not $splitOperatorEnabled) -}}
{{- fail "invalid rook config: `rook_ceph_cluster` requires `rook_ceph_operator` to be enabled first" -}}
{{- end -}}
```

Or we can just be careful during migration since it's going to be a one time thing.

cardoe:
I think I'll just aim to be safe.

PR #1923, opened 2026-04-08
feat(understackctl): deploy check uses kustomize build to validate

cardoe:
Validate the manifests by running 'kustomize build' against each enabled
component.

### DIFF
--- a/go/understackctl/cmd/deploy/check.go
+++ b/go/understackctl/cmd/deploy/check.go
@@ -3,11 +3,22 @@ package deploy
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/charmbracelet/log"
 	"github.com/spf13/cobra"
 )
+
+// kustomizeBuildArgs mirrors the kustomize.buildOptions configured in
+// components/argocd/values.yaml so local validation matches ArgoCD's behaviour.
+var kustomizeBuildArgs = []string{
+	"--enable-alpha-plugins",
+	"--enable-exec",
+	"--enable-helm",
+	"--load-restrictor", "LoadRestrictionsNone",
+}
 
 func newCmdDeployCheck() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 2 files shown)


skrobul (approved):
This works, but I was wondering what do you think about not relying on the kustomize binary to be present on the system and running kustomize natively instead?


```go
import (
	"fmt"
	"path/filepath"
	"sigs.k8s.io/kustomize/api/krusty"
	kusttypes "sigs.k8s.io/kustomize/api/types"
	"sigs.k8s.io/kustomize/kyaml/filesys"
)
// validateKustomizationNative runs the equivalent of:
// kustomize build --enable-alpha-plugins --enable-exec --enable-helm --load-restrictor LoadRestrictionsNone <compDir>
func validateKustomizationNative(compDir string) error {
	opts := krusty.MakeDefaultOptions()
	// --load-restrictor LoadRestrictionsNone
	opts.LoadRestrictions = kusttypes.LoadRestrictionsNone
	// --enable-alpha-plugins + --enable-helm
	opts.PluginConfig = kusttypes.EnabledPluginConfig(kusttypes.BploLoadFromFileSys)
	opts.PluginConfig.HelmConfig.Command = "helm"
	// --enable-exec (for function plugins)
	opts.PluginConfig.FnpLoadingOptions.EnableExec = true
	k := krusty.MakeKustomizer(opts)
	if _, err := k.Run(filesys.MakeFsOnDisk(), compDir); err != nil {
		return fmt.Errorf("%s: %w", filepath.Join(compDir, "kustomization.yaml"), err)
	}
	return nil
}
```
And in `runDeployCheck`, replace the `exec.Command(...) with:

```go
if err := validateKustomizationNative(compDir); err != nil {
	kustomizeErrors = append(kustomizeErrors, err.Error())
}
```

the import block will need:

```go
	"sigs.k8s.io/kustomize/api/krusty"
	kusttypes "sigs.k8s.io/kustomize/api/types"
	"sigs.k8s.io/kustomize/kyaml/filesys"
```


the docs:
https://pkg.go.dev/sigs.k8s.io/kustomize/api/krusty
https://pkg.go.dev/sigs.k8s.io/kustomize/api/types

cardoe:
That's probably better

PR #1926, opened 2026-04-09
chore(urn): Adds device URN template

nicholaskuechler:


cardoe (changes requested):
Thumbs down on this in this repo. This is a deployment specific.

PR #1929, opened 2026-04-10
feat(nautobot): plumb the partition data down to a ConfigMap

cardoe:
Use the partition data that's available as an annotation on the cluster
definitions down to a ConfigMap which can be used by Nautobot.

### DIFF
--- a/charts/argocd-understack/templates/application-nautobot.yaml
+++ b/charts/argocd-understack/templates/application-nautobot.yaml
@@ -31,6 +31,15 @@ spec:
     ref: understack
     repoURL: {{ include "understack.understack_url" $ }}
     targetRevision: {{ include "understack.understack_ref" $ }}
+    kustomize:
+      patches:
+        - patch: |
+            - op: replace
+              path: /data/UNDERSTACK_PARTITION
+              value: "{{ index $.Values.appLabels "understack.rackspace.com/partition" | default "" }}"
+          target:
+            kind: ConfigMap
+            name: cluster-data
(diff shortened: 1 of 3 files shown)


nicholaskuechler:
Can you re-add cluster.

nicholaskuechler (approved):

skrobul (commented):
@cardoe I am -1 on this. We already have [`cluster-metadata`](https://github.com/RSS-Engineering/undercloud-deploy/blob/main/charlie-uc-iad3-staging/argo-events/cm-base/cm-cluster-metadata.yaml) to hold this type of information. Adding `cluster-data` will only confuse things. Any reason we cannot use that?

cardoe:
> @cardoe I am -1 on this. We already have [`cluster-metadata`](https://github.com/RSS-Engineering/undercloud-deploy/blob/main/charlie-uc-iad3-staging/argo-events/cm-base/cm-cluster-metadata.yaml) to hold this type of information. Adding `cluster-data` will only confuse things. Any reason we cannot use that?
> 

cluster-metadata is manually created while this is the same data auto-created. So I would get rid of the former.

skrobul:
cluster-metadata has quite a lot more than just a partition info:

- DNS server IPs
-  NAUTOBOT_URL - arguably can be derived from partition
- NTP server ips
- UCVNI_GROUP_NAME
- UNDERSTACK_CLUSTER_NAME
-  UNDERSTACK_ENV: staging - is the equivalent of partition

PR #1934, opened 2026-04-14
feat(nautobotop): add UCVNI group support to VLAN group sync

abhimanyu003:
Add UCVNI group support to VLAN group sync.

Example of yaml config.
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: vlan-groups
  namespace: nautobot
data:
  vlan-groups.yaml: |
    - name: management-vlans
      location: iad3
      range: "100-199"
      ucvni_group: spine402-1.iad3
    - name: guest-vlans
      location: dfw1
      range: "200-299"
      ucvni_group: spine402-1.iad3
    - name: production-vlans
      location: iad3
      range: "300-399"
      ucvni_group: spine402-1.iad3
```

### DIFF
--- a/go/nautobotop/internal/nautobot/ipam/ucvniGroup.go
+++ b/go/nautobotop/internal/nautobot/ipam/ucvniGroup.go
@@ -3,14 +3,11 @@ package ipam
 import (
 	"context"
 	"fmt"
-	"io"
 	"net/http"
 
 	"github.com/charmbracelet/log"
 	"github.com/rackerlabs/understack/go/nautobotop/internal/nautobot/cache"
 	"github.com/rackerlabs/understack/go/nautobotop/internal/nautobot/client"
-	"github.com/rackerlabs/understack/go/nautobotop/internal/nautobot/helpers"
-	"k8s.io/apimachinery/pkg/util/json"
 )
 
 // UcvniGroup represents a UCVNI group from the undercloud-vni plugin API.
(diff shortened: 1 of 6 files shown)


go/nautobotop/internal/nautobot/sync/vlanGroup.go
```diff
@@ -58,10 +60,19 @@ func (s *VlanGroupSync) syncSingleVlanGroup(ctx context.Context, vlanGroup model
...
+		ucvniGroup := s.ucvniGroupSvc.GetByName(ctx, vlanGroup.UcvniGroup)
+		if ucvniGroup.ID != "" {
+			relationships := map[string]nb.ApprovalWorkflowDefinitionRequestRelationshipsValue{
+				"ucvnigroup_vlangroup": helpers.BuildRelationshipSource(ucvniGroup.ID),
+			}
+			vlanGroupRequest.Relationships = &relationships
+		} else {
+			log.Info("ucvni group not found, skipping relationship", "name", vlanGroup.UcvniGroup)
+		}
 	}
```
skrobul, line 73: if I understand this correctly, when the ucvnigroup is removed now, it won't be reconciled - what do you think about changing to something closer to this:

```go
relationships := map[string]nb.ApprovalWorkflowDefinitionRequestRelationshipsValue{
	"ucvnigroup_vlangroup": helpers.BuildRelationshipSource(), // empty => clear relation
}
if vlanGroup.UcvniGroup != "" {
	ucvniGroup := s.ucvniGroupSvc.GetByName(ctx, vlanGroup.UcvniGroup)
	if ucvniGroup.ID != "" {
		relationships["ucvnigroup_vlangroup"] = helpers.BuildRelationshipSource(ucvniGroup.ID)
	} else {
		log.Info("ucvni group not found; leaving relationship empty", "name", vlanGroup.UcvniGroup)
	}
}
vlanGroupRequest.Relationships = &relationships
```

go/nautobotop/internal/nautobot/ipam/ucvniGroup.go
```diff
@@ -0,0 +1,93 @@
...
+// GetByName looks up a UCVNI group by name
+func (s *UcvniGroupService) GetByName(ctx context.Context, name string) UcvniGroup {
+	if group, ok := cache.FindByName(s.client.Cache, "ucvnigroups", name, func(g UcvniGroup) string {
+		return g.Name
+	}); ok {
+		return group
+	}
+
+	baseURL := s.client.Config.Servers[0].URL
+	url := fmt.Sprintf("%s/plugins/undercloud-vni/ucvni-groups/?name=%s&depth=0&limit=1000&offset=0", baseURL, name)
```
skrobul: this does not urlencode the name correctly, how about trying something like this:

```go
	baseURL := s.client.Config.Servers[0].URL
	u, err := url.Parse(baseURL)
	if err != nil {
		s.client.AddReport("GetUcvniGroupByName", "invalid base URL", "base_url", baseURL, "error", err.Error())
		return UcvniGroup{}
	}
	u.Path = "/plugins/undercloud-vni/ucvni-groups/"
	q := u.Query()
	q.Set("name", name)
	q.Set("depth", "0")
	q.Set("limit", "1000")
	q.Set("offset", "0")
	u.RawQuery = q.Encode()
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
	if err != nil {
		s.client.AddReport("GetUcvniGroupByName", "failed to create request", "name", name, "error", err.Error())
		return UcvniGroup{}
	}
```

https://pkg.go.dev/net/url#example-URL.Query

abhimanyu003: Thanks so much, Updated to use the req.cool client, which is already the default HTTP client in our Nautobot SDK.

go/nautobotop/internal/nautobot/sync/vlanGroup.go
abhimanyu003: This make sense, thanks for noticing this. Updated it

skrobul (approved):
LGTM

PR #1950, opened 2026-04-14
fix: Rework enrol workflow

stevekeay:
Make enroll leverage more of the openstack features.

We perform virtual-media boot for the initial inspection.

We update our handling of ironic ports to make sure the proper values are set at all points in the server life-cycle.


skrobul (commented):
LGTM, needs a cosmetic fix (see CI)

python/understack-workflows/understack_workflows/bmc_chassis_info.py
```diff
@@ -142,22 +85,25 @@ def bmc_interface(bmc) -> dict:
...
-    bmc_name = "iDRAC" if get_system_vendor(bmc) == "Dell" else "iLO"
-    bmc_description = (
-        "Dedicated iDRAC interface" if (bmc_name == "iDRAC") else data.get("name")
+
+    if get_system_vendor(bmc) == "Dell":
+        bmc_name = "iDRAC"
+        bmc_description = "Dedicated iDRAC interface"
+    else:
+        bmc_name = "iLO"
+        bmc_description = str(data.get("name"))
```
skrobul, line 94: else -> raise HardwareNotSupportedError (or similar)?

skrobul (commented):
it looks good - one small nit - would you be opposed standardising  on the `enroll` spelling to avoid potential mistakes? (it will work as is now, but it feels like maintaining two versions can lead to unnecessary issues)

python/understack-workflows/understack_workflows/bmc_chassis_info.py
stevekeay: fixed in e633b7c6 although I think this logic was originally used for creating things in Nautobot, and that responsibility moved elsewhere.  This code might shrink away altogether in future.

stevekeay:
> it looks good - one small nit - would you be opposed standardising on the `enroll` spelling to avoid potential mistakes? (it will work as is now, but it feels like maintaining two versions can lead to unnecessary issues)

I changed the spelling for comments and private references, but I left the physical network name as "enrol" for now because this is referenced externally - changing it requires that we first change the undersync API, to have it accept the new value.

skrobul (approved):

PR #1957, opened 2026-04-17
feat(nautobot-target-proxy): Moves nautobot-target-proxy in to Understack

nicholaskuechler:
* Moves nautobot-target-proxy app in to understack repo
* Adds a helm chart for nautobot-target-proxy
* Adds tests for nautobot-target-proxy
* Converts nautobot-target-proxy from poetry to uv
* Adds docs for nautobot-target-proxy


charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if eq (include "understack.isEnabled" (list $.Values.global "nautobot_target_proxy")) "true" }}
```
cardoe:
```suggestion
{{- if eq (include "understack.isEnabled" (list $.Values.site "nautobot_target_proxy")) "true" }}
```

no since you want this to be at the site?

charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
```diff
@@ -0,0 +1,44 @@
...
+    targetRevision: {{ include "understack.deploy_ref" $ }}
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    managedNamespaceMetadata:
+      annotations:
+        argocd.argoproj.io/sync-options: Delete=false
+    syncOptions:
+    - CreateNamespace=true
```
cardoe, line 40: The namespace is probably created by the site worker so we don't want this line.

charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
```diff
@@ -0,0 +1,44 @@
...
+    ref: deploy
+    repoURL: {{ include "understack.deploy_url" $ }}
+    targetRevision: {{ include "understack.deploy_ref" $ }}
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    managedNamespaceMetadata:
+      annotations:
+        argocd.argoproj.io/sync-options: Delete=false
```
cardoe, line 38: We won't want this if the namespace is created by the site worker

charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
nicholaskuechler: fixed

charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
nicholaskuechler: Doesn't this protect against deletions which is what we want?

charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
nicholaskuechler: I question whether it matters, but if it does matter, I think the assumption that nautobot-worker creates it is wrong, since nautobot-target-proxy comes before nautobot-worker alphabetically.

charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
cardoe: No it doesn't. It just creates a fight for ownership.

charts/argocd-understack/templates/application-nautobot-target-proxy.yaml
cardoe: It's not alphabetical. It's by sync order and this should come after.

cardoe (commented):
Overall I still think this is the wrong way to monitor the hardware. The metrics endpoint for many servers that Nidhi is working on is the better tool.

PR #1959, opened 2026-04-20
chore(argocd-understack): switch to unified 'understack' AppProject

cardoe:
This switches us to the unified 'understack' AppProject for ArgoCD. From
here we'll be able to grant controls and actions on all Applications
that are part of Understack and start to craft proper limiting rules.

### DIFF
--- a/apps/kustomization.yaml
+++ b/apps/kustomization.yaml
@@ -3,4 +3,4 @@ kind: Kustomization
 resources:
 - appsets/project-understack-infra.yaml
 - appsets/project-understack-operators.yaml
-- appsets/project-understack.yaml
+- ../bootstrap/argocd-understack/appproject-understack.yaml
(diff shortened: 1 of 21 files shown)


skrobul (approved):
looks good syntax wise, but are we sure what ArgoCD is going to do when we move apps between projects or rename them ?:)

ctria:
I don't see any issue although I'm still not sure that all those apps should be part of understack deployment and not dependencies of understack (i.e. cilium, cert-manager) that need to be fulfilled before installing understack.

Regarding what ArgoCD will do. Moving apps between projects is harmless but renaming is not. My understanding is that it depends on what will be synced first, the existence of the new app or the removal of the old (for ArgoCD there is no rename, it is an app that removed and an app that is added).

cardoe:
> I don't see any issue although I'm still not sure that all those apps should be part of understack deployment and not dependencies of understack (i.e. cilium, cert-manager) that need to be fulfilled before installing understack.
> 
So that's why I originally had the separate AppProject because there was one set that was dependencies and one that was the actual parts.

PR #1960, opened 2026-04-21
fix: wrong usage of the python ironic client

stevekeay:


skrobul (changes requested):
`client.runbook` and other runbook features are available from python-ironic-client 5.8.0 onwards, let's update the [dependency list](https://github.com/rackerlabs/understack/blob/200ddf89f53b4d74e34d9154c52a5a478004c695/python/understack-workflows/pyproject.toml#L25) to reflect that

skrobul (approved):

PR #1964, opened 2026-04-22
fix: Update enroll timeout for cleaning, etc., to 90 mins

stevekeay:
We have some long-running operations like firmware upgrades that were hitting the timeout.


mfencik (approved):
LGTM

PR #1976, opened 2026-04-23
fix(nautobot): envoy's TLS passthrough incompatibility with postgres TLS ClientHello

haseebsyed12:
# PostgreSQL STARTTLS and Envoy `TCPRoute`

This PR is to fix a specific issue that affected Nautobot site workers
connecting from site clusters back to the global PostgreSQL service
through Envoy Gateway.

## Problem

PostgreSQL does not begin a connection with a TLS ClientHello. It first
sends a plaintext `SSLRequest` and only then upgrades to TLS
(`STARTTLS`).

Envoy `TLSRoute` passthrough expects to see a TLS ClientHello
immediately so it can inspect SNI and choose a backend. That works for
Redis and HTTPS, but not for PostgreSQL:

- Redis: direct TLS from the first packet, so `TLSRoute` works
- PostgreSQL: plaintext `SSLRequest` first, so `TLSRoute` cannot route
  correctly and the connection hangs or is closed

This was the reason site Nautobot Celery workers could reach Redis but
could not reliably connect to PostgreSQL through the external gateway on
port `5432`.

## Implemented Solution

We changed the Envoy configuration for PostgreSQL from `TLSRoute` to
`TCPRoute`.

That means:

- the gateway listener on port `5432` now uses `protocol: TCP`
- the PostgreSQL route is now a `TCPRoute`
- Envoy forwards raw bytes without trying to inspect TLS
- PostgreSQL still performs its normal STARTTLS negotiation end-to-end
- PostgreSQL still enforces client certificate authentication with
  `pg_hba ... cert`

This is not a downgrade in authentication or encryption. The gateway is
no longer relying on SNI for PostgreSQL, but PostgreSQL traffic is still
encrypted and still requires a valid client certificate.

## Why `TCPRoute` Is Correct

`TCPRoute` is appropriate when a protocol does not start with a TLS
ClientHello, or when a dedicated port already maps to a single backend
and SNI-based routing adds no value.

That is exactly the case for the Nautobot PostgreSQL endpoint:

- dedicated listener: `5432`
- single backend: `nautobot-cluster-rw`
- protocol behavior: PostgreSQL STARTTLS

## Validation

After deploying the route change, validate from a site worker pod.

### 1. PostgreSQL connection

```bash
kubectl exec -n nautobot <pod> -- python3 -c "
import os, psycopg2
conn = psycopg2.connect(
    host='nautobot-db.<env>.example.com',
    port='5432', dbname='app', user='app',
    password=os.environ.get('NAUTOBOT_DB_PASSWORD', ''),
    sslmode='verify-ca',
    sslcert='/etc/nautobot/mtls/tls.crt',
    sslkey='/etc/nautobot/mtls/tls.key',
    sslrootcert='/etc/nautobot/mtls/ca.crt',
    connect_timeout=10)
print('DB CONNECTION OK - SSL:', conn.info.ssl_in_use)
conn.close()
"
```

Expected:

```text
DB CONNECTION OK - SSL: True
```

### 2. Celery worker control ping

Use the Nautobot wrapper, not `celery -A nautobot.celery`, because this
image exposes Celery through `nautobot-server`:

```bash
kubectl exec -n nautobot deploy/nautobot-worker-celery-<partition> -- sh -lc \
'nautobot-server celery inspect ping --destination "celery@$HOSTNAME"'
```

Expected:

```text
->  celery@<pod-name>: OK
        pong

1 node online.
```

### 3. Worker logs

```bash
kubectl logs -n nautobot deploy/nautobot-worker-celery-<partition> --tail=200
```

Expected log lines include:

```text
Connected to rediss://...
mingle: sync complete
celery@<pod-name> ready.
```

### DIFF
--- a/components/envoy-configs/templates/gw-external.yaml.tpl
+++ b/components/envoy-configs/templates/gw-external.yaml.tpl
@@ -50,6 +50,21 @@ spec:
           from: {{ .from | default "All" }}
           {{- end }}
     {{- end }}
+    {{- range .Values.routes.tcp }}
+    {{- $listenerName := .name | default (index (splitList "." .fqdn) 0) }}
+    - name: {{ $listenerName }}
+      port: {{ .gatewayPort | default ($.Values.gateways.external.port | default 443) }}
+      protocol: TCP
+      allowedRoutes:
+        namespaces:
+          {{- if .selector }}
+          from: Selector
+          selector:
+            {{- .selector | toYaml | nindent 12 }}
+          {{- else }}
+          from: {{ .from | default "All" }}
+          {{- end }}
+    {{- end }}
 
   {{- if .Values.gateways.external.serviceAnnotations }}
   infrastructure:
--- a/components/envoy-configs/templates/tcproute.yaml.tpl
+++ b/components/envoy-configs/templates/tcproute.yaml.tpl
@@ -1,0 +1,23 @@
+{{- range .Values.routes.tcp }}
+{{- $listenerName := .name | default (index (splitList "." .fqdn) 0) }}
+---
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: TCPRoute
+metadata:
+  name: {{ $listenerName }}
+  namespace: {{ .namespace | default "envoy-gateway" }}
+  labels:
+    {{- include "envoy-configs.labels" $ | nindent 4 }}
+spec:
+  parentRefs:
+    - name: {{ $.Values.gateways.external.name }}
+      namespace: {{ $.Values.gateways.external.namespace }}
+      sectionName: {{ $listenerName }}
+  rules:
+    - backendRefs:
+        - name: {{ .service.name }}
+          {{- with .namespace }}
+          namespace: {{ . }}
+          {{- end }}
+          port: {{ .service.port }}
+{{- end }}
--- a/components/envoy-configs/values.schema.json
+++ b/components/envoy-configs/values.schema.json
@@ -69,7 +69,7 @@
     },
     "routes": {
       "type": "object",
-      "description": "HTTP and TLS routes configurations for Understack services",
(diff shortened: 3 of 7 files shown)


skrobul (commented):
did you try adding `sslnegotiation=direct` as described [here](https://github.com/envoyproxy/gateway/issues/4594#issuecomment-3275445982) ?

haseebsyed12:
> sslnegotiation=direct

our CNPG/PostgreSQL server image  is still PostgreSQL 15.5
and this feature is introduced in [version 17](https://www.postgresql.org/docs/17/libpq-connect.html#LIBPQ-CONNECT-SSLNEGOTIATION)

skrobul:
> > our CNPG/PostgreSQL server image is still PostgreSQL 15.5

> and this feature is introduced in [version 17](https://www.postgresql.org/docs/17/libpq-connect.html#LIBPQ-CONNECT-SSLNEGOTIATION)

Fair point. I don't think we have any good reason not to [upgrade](https://cloudnative-pg.io/docs/1.29/postgres_upgrades) to v17 or even v18.


The [operator supports it](https://cloudnative-pg.io/docs/1.29/supported_releases#support-status-of-cloudnativepg-releases). Nautobot just needs 14 or newer.

haseebsyed12:
> > > our CNPG/PostgreSQL server image is still PostgreSQL 15.5
> 
> > and this feature is introduced in [version 17](https://www.postgresql.org/docs/17/libpq-connect.html#LIBPQ-CONNECT-SSLNEGOTIATION)
> 
> Fair point. I don't think we have any good reason not to [upgrade](https://cloudnative-pg.io/docs/1.29/postgres_upgrades) to v17 or even v18.
> 
> The [operator supports it](https://cloudnative-pg.io/docs/1.29/supported_releases#support-status-of-cloudnativepg-releases). Nautobot just needs 14 or newer.

ok. I will upgrade cnpg

PR #1979, opened 2026-04-23
chore: Enrol can clear bios jobs

stevekeay:
(1) It's quite easy to create "conflicting" jobs, so that a stale job that was left behind prevents any progress because the new job can't be created until the old one is deleted.

This just clears all jobs on startup.  I know there is an ironic hook that ships with ironic and therefore is superior to adding code here, but running it here makes it run at the right time, when we need it to.

We also add an option to call the ironic "known_good_state" which reboots the idrac, just to make sure.   Unfortunately this is broken in our Ironic because it relies on the "ping" binary (like we are still in 2006) that is not present in our container.   It still resets the drac but then it errors out, forcing the user to unset maintenance mode, wait a while, and try again.

(2) If we are to update BIOS settings, we go ahead and take a second reboot, to ensure the BIOS settings change gets committed before we take on anything else.  This takes a very long time, but trying to stack the bios update alongside other updates seems to result in failures.  This change makes the process slower, but more likely to work first time.

(3) disable some more PXE devices - these can get left enabled by prior processes and confuse the boot process output.  We're not using PXE so disable it, period.

### DIFF
--- a/python/understack-workflows/tests/test_enroll_server.py
+++ b/python/understack-workflows/tests/test_enroll_server.py
@@ -202,8 +202,12 @@ def test_enrol_happy_path_uses_virtual_media_inspect_and_flips_back(mocker):
     ]
     fake_ironic, created_node = make_ironic_client(
         node_name="Dell-ABC123",
-        # OOB inspect, agent inspect, OOB inspect (post-RAID).
-        inspect_interfaces=["idrac-redfish", "idrac-redfish", "idrac-redfish"],
+        inspect_interfaces=[
+            "idrac-redfish",
+            "idrac-redfish",
+            "idrac-redfish",
+            "idrac-redfish",
+        ],
         inventory=inventory,
         ports=ports,
     )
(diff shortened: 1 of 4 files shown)


cardoe:
@stevekeay Are you wanting to call `known_good_state` in the normal path? It looks like it just resets the iDRAC and then clears the job queue per https://github.com/openstack/ironic/blob/1114d43ca8cde969e87dd99f5c70cc9c5a57fe93/ironic/drivers/modules/drac/management.py#L161

stevekeay:
> @stevekeay Are you wanting to call `known_good_state` in the normal path? It looks like it just resets the iDRAC and then clears the job queue per https://github.com/openstack/ironic/blob/1114d43ca8cde969e87dd99f5c70cc9c5a57fe93/ironic/drivers/modules/drac/management.py#L161

The plan was to do that ONLY on explicit request by a command line flag.

known_good_state doesn't do any harm but it adds a lot of time to a process which is already very slow.   If we find ourselves needing to do it a lot then we can make it the default.

cardoe:
> > @stevekeay Are you wanting to call `known_good_state` in the normal path? It looks like it just resets the iDRAC and then clears the job queue per https://github.com/openstack/ironic/blob/1114d43ca8cde969e87dd99f5c70cc9c5a57fe93/ironic/drivers/modules/drac/management.py#L161
> 
> The plan was to do that ONLY on explicit request by a command line flag.
> 
> known_good_state doesn't do any harm but it adds a lot of time to a process which is already very slow. If we find ourselves needing to do it a lot then we can make it the default.

Well before making it the default we'd replace the ping call with hitting the redfish API endpoint in a backoff timeout.

cardoe (commented):
It's not clear to me why the agent inspection here a second time vs the other location?

PR #1985, opened 2026-04-27
feat: Use inventory for raid details instead of BMC

stevekeay:
We now have full raid controller details available after redfish
inspection so there is no need to probe the BMC for them.

We also update the business logic to group disks by size, to avoid
building RAID arrays out of mismatched disks.  This behaviour is
ported from #1978

I deleted the old raid script script to avoid further confusion.

This also fixes a RAID delete problem.


mfencik (approved):
LGTM

python/understack-workflows/understack_workflows/raid.py
```diff
@@ -0,0 +1,120 @@
...
+    disks = storage_controller.get("drives", [])
+
+    if "RAID" not in controller_id.upper():
+        return set()
+
+    return {
+        PhysicalDisk(
+            id=disk["id"],
+            controller=controller_id,
+            size_gb=disk["size"] // 10**9,
```
skrobul, line 97: is there any realistic scenario where the disk["size"] or disk["id"] is None? if yes, consider skipping that iteration and logging the warning

skrobul (approved):
lgtm

python/understack-workflows/understack_workflows/raid.py
stevekeay: Honestly this data is brand new, and I have not seen the code that generates it, so anything could happen.

If the data is not in the expected / advertised format though, is it not better for enrol to fail and expose the problem, rather than build raid arrays with missing devices?

skrobul: Fair point - it's better to fail early, I just would prefer to have a log message explaining why rather than `TypeError: unsupported operand type(s) for //: 'NoneType' and 'int'`

stevekeay: I'll improve this as a follow-on

PR #1994, opened 2026-04-30
feat(nautobotOP): Use UUID for Nautobot Operator Sync

abhimanyu003:
This PR focuses on Nautobot Operator Sync.
- During sync, the operator will create objects using the unique ID provided in the YAML file.
- This helps support renaming objects in Nautobot.
  - If the unique ID already exists, the operator will rename or update the object in Nautobot.
  - If the unique ID does not exist, the operator will create a new object with the given UUID.
- Delete behavior remains unchanged the operator will continue to check whether the object was created by the operator user. If it was not, the delete operation will be skipped.

```mermaid
flowchart TD
    A[Start sync] --> B[Read object from YAML]
    B --> C[Get unique ID from YAML]
    C --> D{Does unique ID exist in Nautobot?}

    D -- Yes --> E[Rename or update existing object]
    D -- No --> F[Create new object with given UUID]

    E --> G[Sync complete]
    F --> G

    H[Delete requested] --> I{Was object created by operator user?}
    I -- Yes --> J[Delete object from Nautobot]
    I -- No --> K[Skip delete]

    J --> L[Delete complete]
    K --> L
```


go/nautobotop/internal/nautobot/dcim/location.go
```diff
@@ -77,16 +99,6 @@ func (s *LocationService) ListAll(ctx context.Context) []nb.Location {
 }
 
 func (s *LocationService) Update(ctx context.Context, id string, req nb.LocationRequest) (*nb.Location, error) {
-	owned, err := s.client.IsCreatedByUser(ctx, id)
-	if err != nil {
-		s.client.AddReport("UpdateLocation", "failed to check ownership", "id", id, "error", err.Error())
-		return nil, err
-	}
-	if !owned {
```
nidzrai, line 85: if !owned ,  this check is gone from Update but still exists in destroy .  without this guard one can accidentally puts a UUID of a human-created or externally-managed Nautobot object in their YAML.  the operator will overwrite it without any warnin I see .

go/nautobotop/internal/nautobot/dcim/templates/console_port.go
```diff
@@ -22,29 +21,16 @@ func NewConsolePortTemplateService(nautobotClient *client.NautobotClient) *Conso
...
-	apiCall := func(ctx context.Context, batchIds []string) ([]nb.ConsolePortTemplate, *http.Response, error) {
-		list, resp, err := s.client.APIClient.DcimAPI.DcimConsolePortTemplatesList(ctx).Id(batchIds).Depth(2).DeviceType([]string{deviceTypeID}).Execute()
-		if err != nil {
-			return nil, resp, err
-		}
-		if list == nil {
-			return []nb.ConsolePortTemplate{}, resp, nil
-		}
-		return list.Results, resp, nil
+	list, resp, err := s.client.APIClient.DcimAPI.DcimConsolePortTemplatesList(ctx).Limit(10000).Depth(2).DeviceType([]string{deviceTypeID}).Execute()
```
nidzrai: It's a hardcoded ceiling

go/nautobotop/internal/nautobot/models/devicetype.go
```diff
@@ -5,6 +5,7 @@ type DeviceTypes struct {
 }
 
 type DeviceType struct {
+	ID            string          `yaml:"id"`
```
nidzrai, line 8: if we ever Marshals it to JSON    , the ID field would serialized  as "ID"  instead of "id" . I saw in cluster.go this pattern` ID           string   json:"id" yaml:"id"`

go/nautobotop/internal/nautobot/dcim/manufacturer.go
```diff
@@ -23,28 +22,20 @@ func NewManufacturerService(nautobotClient *client.NautobotClient) *Manufacturer
...
+	list, resp, err := s.client.APIClient.DcimAPI.DcimManufacturersList(ctx).Limit(10000).Depth(10).Execute()
+	if err != nil {
+		bodyString := helpers.ReadResponseBody(resp)
+		s.client.AddReport("ListAllManufacturers", "failed to list", "error", err.Error(), "response_body", bodyString)
+		return []nb.Manufacturer{}
+	}
+	if list == nil || len(list.Results) == 0 {
+		return []nb.Manufacturer{}
+	}
+	if list.Results[0].Id == nil {
```
nidzrai: In ListAll, if Results[0].Id == nil, the entire result set is discarded and an empty slice is returned. I m not sure if it is intentional . but let's say there are 10 items , it would check just 1st or the 0th position and would discard all . right ?

go/nautobotop/internal/nautobot/dcim/location.go
```diff
@@ -58,9 +58,31 @@ func (s *LocationService) GetByName(ctx context.Context, name string) nb.Locatio
...
+	if err != nil {
+		bodyString := helpers.ReadResponseBody(resp)
+		s.client.AddReport("GetLocationByID", "failed to get", "id", id, "error", err.Error(), "response_body", bodyString)
+		return nb.Location{}
+	}
+	if list == nil || len(list.Results) == 0 || list.Results[0].Id == nil {
+		return nb.Location{}
+	}
+
+	return list.Results[0]
```
nidzrai, line 82: so here if found in cache, no API call needed  otherwise  and it call API. bt  where would it update cache ? how does it write ?

go/nautobotop/internal/nautobot/sync/helpers.go
```diff
@@ -0,0 +1,10 @@
+package sync
+
+import nb "github.com/nautobot/go-nautobot/v3"
+
+func optionalID(id string) *string {
+	if id == "" {
+		return nil
+	}
+	return nb.PtrString(id)
```
nidzrai, line 9: what if I put in valid format but belongs to another object. and since there is not owner check . I think it would overwrite .

go/nautobotop/internal/nautobot/models/devicetype.go
abhimanyu003: I didn't get what you are saying here ?

go/nautobotop/internal/nautobot/dcim/manufacturer.go
abhimanyu003: This is just a safe guard to prevent app from panic and check if we recieved the data.

go/nautobotop/internal/nautobot/dcim/location.go
abhimanyu003: Can update will auto happen next time when reconcile happens, there cache is ony for one time use.
Sure we can look into it more, but in future update

go/nautobotop/internal/nautobot/sync/helpers.go
abhimanyu003: Not sure, why we want to owner check here ?
This function is a helper function that will create pointer of string

nidzrai (approved):

PR #1999, opened 2026-05-01
feat: Add Nautobot job queues ArgoCD application and Helm chart

haseebsyed12:
Adds the nautobot-job-queues component to ArgoCD and wires it into the global deployment flow. 
The chart provides a Kubernetes Job, ConfigMaps, values schema, and API token configuration for reconciling Nautobot job queue state after Nautobot and API tokens are available.


cardoe (approved):
I'm okay with this if we plan to move this to the operator.

PR #2017, opened 2026-05-07
chore: upgrade to latest dex container

cardoe:
Upgrade to the latest dex container and add comment so that Renovate can
track this for us. fixes PUC-1673


mfencik (approved):
LGTM

PR #2034, opened 2026-05-12
feat: a job that configures the nb constance config to set driver backend settings

khackworth:
### Tested locally against dev:

```
understack % kubectl logs -n nautobot job/nautobot-constance-config -f

Waiting for Nautobot to be ready...
DatabaseBackend          ... working
DefaultFileStorageHealthCheck ... working
MigrationsBackend        ... working
RedisBackend             ... working
Configuring Golden Config Constance settings...
19:53:25.549 DEBUG   git.cmd              cmd.py                                 execute() :
  Popen(['git', 'version'], cwd=/opt/nautobot, stdin=None, shell=False, universal_newlines=False)
19:53:25.597 DEBUG   git.cmd              cmd.py                                 execute() :
  Popen(['git', 'version'], cwd=/opt/nautobot, stdin=None, shell=False, universal_newlines=False)
197 objects imported automatically (use -v 2 for details).

============================================================
Golden Config Constance framework settings configured
============================================================
DEFAULT_FRAMEWORK: {'cisco_nxos': 'netmiko', 'all': 'napalm'}
GET_CONFIG_FRAMEWORK: {'cisco_nxos': 'netmiko', 'all': 'napalm'}
============================================================
Configuration complete
```

### Showing test results on the NB shell

` kubectl exec -it -n nautobot deployment/nautobot-default -- nautobot-server shell`

##### Pre Deploy:

```
>>> from constance import config
>>> print(config.nautobot_golden_config__DEFAULT_FRAMEWORK)
{'all': 'netmiko'}
>>> print(config.nautobot_golden_config__GET_CONFIG_FRAMEWORK)
{}
```

#### Post Deploy:

```
>>> from constance import config
>>> print(config.nautobot_golden_config__GET_CONFIG_FRAMEWORK)
{'cisco_nxos': 'netmiko', 'all': 'napalm'}
>>> print(config.nautobot_golden_config__DEFAULT_FRAMEWORK)
{'cisco_nxos': 'netmiko', 'all': 'napalm'}
```

### DIFF
--- a/components/nautobot/job-nautobot-constance-config.yaml
+++ b/components/nautobot/job-nautobot-constance-config.yaml
@@ -1,0 +1,129 @@
+---
+# nautobot constance Configuration job
+# -----------------
+# Built to enforce proper mapping for cisco_nxos to use netmiko
+# resolves issues where workers using napalm were falling
+# back to nxapi
+
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: nautobot-constance-config
+  namespace: nautobot
+  labels:
+    app.kubernetes.io/name: nautobot
+    app.kubernetes.io/component: constance-config
+  annotations:
+    argocd.argoproj.io/hook: PostSync
+    argocd.argoproj.io/sync-wave: "2"
+    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
+spec:
+  backoffLimit: 3
+  ttlSecondsAfterFinished: 3600
+  template:
+    spec:
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 999
+        fsGroup: 999
+        seccompProfile:
+          type: RuntimeDefault
+      restartPolicy: OnFailure
+      serviceAccountName: nautobot
+      imagePullSecrets:
+        - name: dockerconfigjson-github-com
+      containers:
+      - name: configure-constance
+        image: ghcr.io/rss-engineering/nautobot-rackspace:latest
+        imagePullPolicy: Always
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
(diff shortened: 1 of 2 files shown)


components/nautobot/job-nautobot-constance-config.yaml
```diff
@@ -0,0 +1,129 @@
...
+            drop:
+              - ALL
+          readOnlyRootFilesystem: false
+        resources:
+          requests:
+            cpu: "100m"
+            memory: "256Mi"
+          limits:
+            cpu: "500m"
+            memory: "512Mi"
```
skrobul: is that enough? The 512Mi seems overly restrictive to me given that nautobot-shell loads all of the models and whole Django machinery.

skrobul (approved):

components/nautobot/job-nautobot-constance-config.yaml
khackworth: I'm not sure, it tests fine. Any recommended values? My first stab at one of these. From the looks of it, i could use nb shell, or call via an api after it loads, or a UI option (ruled this one out).

skrobul: nb shell is absolutely fine, just saying that it is likely to try to consume more resources - if you already tested it on any of our environments, you can have a look at grafana graphs for the pod that this was run in. It will show if the CPU was throttled and potentially if the job was OOM killed at any point. But that's only if you want to get scientific about it, otherwise just estimate.

Given that this is an ephemeral pod/job, I'd probably go with slightly higher limits (not reservations). 1000m CPU and 2GB would be my pick. The code is not multi threaded so giving it anymore than 1000m does not make sense. The nautobot shell loads all of the models, so it can bloat memory usage

PR #2041, opened 2026-05-15
feat(ironic): Enables chassis_model inspection hook

nicholaskuechler:


nicholaskuechler:
Fixed in https://github.com/rackerlabs/understack/pull/2052

PR #2058, opened 2026-06-02
fix(neutron):  undersync  mechanism operate only on baremetal ports PUC-1713

nidzrai:


stevekeay (approved):
LGTM although I would vastly prefer if the commit message were amended to add a paragraph describing why this change was made.

skrobul (commented):
can you provide bit more context for this PR ? 

What's the rationale behind excluding 'normal' ports? I may be wrong here, but I think Undersync still requires visibility of those port types which are commonly applied on trunk subports and router interfaces.

stevekeay (changes requested):
There appears to be a merge commit in this PR?

stevekeay (approved):

PR #2059, opened 2026-06-02
feat(neutron): Upgrade neutron dependency to 2026.1 PUC-1685

nidzrai:
https://rackspace.atlassian.net/browse/PUC-1685


haseebsyed12 (approved):

python/neutron-understack/pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 dependencies = [
     "requests>=2,<3",
     "neutron-lib>=3,<4",
-    "neutron>=27,<28",
+    "neutron>=27",
```
skrobul: Wouldn't it be better to keep this constrained? If we know that version 28 is fine, we still don't know if version 29 will be fine so maybe keep the upper bound just in case

nidzrai: right, Thanks . Fixing this

skrobul (approved):

skrobul:
@nidzrai  please squash this before merging

PR #2060, opened 2026-06-02
fix: stop invalidating Keystone tokens on neutron ArgoCD sync

skrobul:
For a while we have been experiencing occasional issues with authentication between Neutron and Undersync. This was traced down to the tokens for service accounts being prematurely invalidated.

Turns out Keystone revokes all of the existing tokens when the password change for a given account is attempted. It does not matter if the password is already set to the exact same value or not. This PR works around that problem by executing password change API call only when the desired password does not authenticate.

Please note this fixes only Neutron where we have seen that problem the most. I am working on a more generic, upstream patch for other components.

related https://rackspace.atlassian.net/browse/PUC-1615

### DIFF
--- a/components/neutron/configmap-neutron-bin.yaml
+++ b/components/neutron/configmap-neutron-bin.yaml
@@ -797,8 +797,12 @@ data:
 
     # Manage user password (we do this in a separate step to ensure the password is updated if required)
     set +x
-    echo "Setting user password via: openstack user set --password=xxxxxxx ${USER_ID}"
-    openstack user set --password="${SERVICE_OS_PASSWORD}" "${USER_ID}"
+    if OS_USERNAME="${SERVICE_OS_USERNAME}" OS_PASSWORD="${SERVICE_OS_PASSWORD}" OS_USER_DOMAIN_NAME="" OS_USER_DOMAIN_ID="${USER_DOMAIN_ID}" OS_PROJECT_NAME="" OS_PROJECT_DOMAIN_NAME="" OS_PROJECT_ID="${USER_PROJECT_ID}" openstack token issue > /dev/null 2>&1; then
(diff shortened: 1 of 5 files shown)


skrobul:
Submitted upstream fix:

https://review.opendev.org/c/openstack/openstack-helm/+/991058

components/neutron/configmap-neutron-bin.yaml
```diff
@@ -0,0 +1,1800 @@
...
+        "${SERVICE_OS_USERNAME}");
+
+    # Manage user password (we do this in a separate step to ensure the password is updated if required)
+    set +x
+    if OS_USERNAME="${SERVICE_OS_USERNAME}" OS_PASSWORD="${SERVICE_OS_PASSWORD}" OS_USER_DOMAIN_NAME="" OS_USER_DOMAIN_ID="${USER_DOMAIN_ID}" OS_PROJECT_NAME="" OS_PROJECT_DOMAIN_NAME="" OS_PROJECT_ID="${USER_PROJECT_ID}" openstack token issue > /dev/null 2>&1; then
+      echo "Password for ${USER_ID} is already correct, skipping update."
+    else
+      echo "Setting user password via: openstack user set --password=xxxxxxx ${USER_ID}"
+      openstack user set --password="${SERVICE_OS_PASSWORD}" "${USER_ID}"
+    fi
```
skrobul, line 805: note for the reviewer: only code in lines 798-805 were adjusted. everything else is stock OSH

stevekeay (approved):

components/neutron/configmap-neutron-bin.yaml
cardoe: Pinging as a reminder to me to look at this to somehow make this work better in upstream.

PR #2065, opened 2026-06-08
fix(ironic): restrict required anti-affinity to conductors only

skrobul:
Conductors use host networking and bind a fixed port, so two conductor pods cannot run on the same host. The API has no such constraint and should use preferred anti-affinity to avoid unnecessary scheduling failures.


mfencik (approved):
LGTM

PR #2066, opened 2026-06-08
fix(image-upload): Harden the image-upload ansible against upload failures and restarts

nicholaskuechler:
2 small issues:
* after reserving and uploading we need the grab the status again as it changes from 'queued' to 'uploading', so it also needs to handle uploading status.
* adds better idempotent handling if we need to re-run the playbook, so we can pick up in the middle of reservation or uploading.

### DIFF
--- a/components/glance/glance-staging-pvc.yaml
+++ b/components/glance/glance-staging-pvc.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    undercloud.local/purpose: glance-staging
+  name: glance-staging
+  namespace: openstack
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: ceph-fs-ec
+  resources:
+    requests:
+      storage: 20Gi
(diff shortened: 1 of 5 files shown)


components/glance/glance-staging-pvc.yaml
```diff
@@ -0,0 +1,15 @@
...
+    undercloud.local/purpose: glance-staging
+  name: glance-staging
+  namespace: openstack
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: ceph-fs-ec
+  resources:
+    requests:
+      storage: 20Gi
```
skrobul, line 15: with windows images this might not be enough

skrobul (approved):

components/glance/glance-staging-pvc.yaml
nicholaskuechler: I agree but we just simply don't have enough free space. Also the workflow to upload an image using glance staging features is different than a normal end-user uploading a brand new glance image without staging it - the end-user, non-staged upload won't use this pvc.

PR #2069, opened 2026-06-10
feat: add convert_image_to_4k.py script - convert disk images for 4096-byte sector targets

skrobul:
### Problem

Our disk images are built on 512-byte-sector disks, but some bare-metal nodes have 4Kn (4096-byte logical sector) drives, or specifically they have RAID controllers like H965i which can present a VD only with 4k block size. 
Deploying a 512-sector image to such a node fails on two fronts:

- **Partition table:** the GPT's LBAs are in 512-byte units, so on a 4Kn device the partitions land at the wrong offsets / the table is misread — the disk layout is effectively invalid.
- **ESP:** even with the layout fixed, the FAT records `bytes/sector=512` in its BPB, which the Linux `vfat` driver won't mount on a 4Kn device — this is where the Ironic Python Agent fails during bootloader install (UEFI firmware would probably reject it later too).

Neither can be fixed by patching bytes in place; the table must be rewritten and the FAT rebuilt.

`scripts/convert_image_to_4k.py` turns a 512-sector image into an equivalent 4Kn image:

- **Rescales the GPT** to 4096-byte sectors, preserving each partition's byte offset and size.
- **Rebuilds the ESP** as a 4096-byte-sector FAT, copying its files across unchanged.
- **Copies all other partitions verbatim** (ext4 is sector-size-agnostic).
- Handles **raw and qcow2** (output mirrors input). No loop devices, no root needed.

### Dependencies

`python3`, `sfdisk`, `dd` always; `mtools` when an ESP is present; `qemu-img` for qcow2.

### Testing

Verified non-root on an Ubuntu 24.04 image (qcow2→qcow2): rebuilt ESP is FAT32/4096 with boot binaries byte-identical to the source; root and BIOS-boot partitions byte-identical; destination GPT verifies clean at 4096. **Confirmed end-to-end on real 4Kn hardware** — successfully deployed and booted on node `ba712e58-77de-4c25-a99f-0068da2fe36b`.

### DIFF
--- a/scripts/convert_image_to_4k.py
+++ b/scripts/convert_image_to_4k.py
@@ -1,0 +1,225 @@
+#!/usr/bin/env python3
+"""convert.py - Convert a 512-byte-sector raw disk image to a 4096-byte-sector
+(4Kn) layout, producing a new raw image.
+
+The new image preserves the byte offset and size of every partition, so the
+filesystems inside copy across unchanged. Only the partition table is rebuilt
+for 4096-byte logical sectors. Works directly on image files: no loop devices,
+no root.
+
+Usage:
+    convert.py <source.img> <output.img>
+
+Requires: python3, sfdisk (util-linux 2.26+), dd (GNU coreutils).
+This is a Python port of convert.sh; it shells out to sfdisk/dd for the table
+write and the copy, and does the sizing and LBA rescaling in Python.
+"""
+
+import os
+import re
+import shutil
+import subprocess
+import sys
+
+SRC_SECTOR = 512
+DST_SECTOR = 4096
+RATIO = DST_SECTOR // SRC_SECTOR  # 8
+HEADROOM = 64 * 1024 * 1024  # tail room for the backup GPT + alignment
+
+
+def die(msg):
+    print(f"ERROR: {msg}", file=sys.stderr)
+    sys.exit(1)
+
+
+def info(msg):
+    print(f">>> {msg}")
+
+
+def run(cmd, *, input=None, capture=False, what=None):
+    """Run a command; die with context on failure. Returns stdout if captured."""
+    res = subprocess.run(
+        cmd,
+        input=input.encode() if input is not None else None,
+        stdout=subprocess.PIPE if capture else None,
+    )
+    if res.returncode != 0:
+        die(f"{what or cmd[0]} failed (exit {res.returncode})")
+    return res.stdout.decode() if capture else None
+
+
+def rescale_value(match):
+    """Divide a `start=`/`size=` LBA by RATIO, refusing non-4K-alignable values."""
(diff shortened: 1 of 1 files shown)


scripts/convert_image_to_4k.py
```diff
@@ -0,0 +1,469 @@
...
+        )
+    return f"{key}={num // RATIO}"
+
+
+def transform_table(dump):
+    """Rebuild the sfdisk dump for 4096-byte sectors."""
+    out = []
+    for line in dump.splitlines():
+        if re.match(r"^(first-lba|last-lba|device):", line):
+            continue
```
stevekeay, line 79: Dropping the first-lba and last-lba numbers means that you;re leaving sfdisk to set those numbers.  I couldn't easily find out what the behaviour is.   Would it make more sense to scale these numbers too?

Not sure if this actually matters, but it seems like the sort of thing that might get rejected by EUFI firmware or whatever if it is inconsistent.

stevekeay (approved):
What could possibly go wrong? ;-)

scripts/convert_image_to_4k.py
skrobul: These are optional GPT header fields documented in sfdisk(8) → "Header lines": first-lba/last-lba specify the first/last usable sector, and when not provided sfdisk uses the "label specific default" - i.e. the usable bounds computed from the target device's
 geometry (after the GPT entry array, before the backup GPT). So dropping them isn't undefined; it hands the computation to sfdisk (as you said).

They can't be carried over or scaled, for two reasons:
 - `last-lba` depends on the destination disk size, which we change (+64 MiB headroom), so any scaled value would point to the wrong place.
 - `first-lba` depends on the GPT metadata size at the target sector size (a minimal GPT's first usable LBA is 34 at 512 B but 6 at 4096 B — and 34 / 8 isn't even an round integer).

This comes from the fact that [UEFI spec](https://uefi.org/specs/UEFI/2.10/05_GUID_Partition_Table_Format.html) states that partition entry array must be minimum of 16,384 bytes,
which with the 512B gives us 32 sectors (+1 for protective and +1 for header). With 4k sectors, this is just (4 sectors x 4096).

In theory we could calculate all ourselves, but I trust the sfdisk/kernel developers more than myself ;)

skrobul: I was curious how it's actually implemented, so here is some extracurricular material for those interested:

https://github.com/util-linux/util-linux/blob/5305e6c70b274f679329b79c0e1ef5a07e9dc1a6/libfdisk/src/gpt.c#L449

https://github.com/util-linux/util-linux/blob/5305e6c70b274f679329b79c0e1ef5a07e9dc1a6/libfdisk/src/gpt.c#L431

https://en.wikipedia.org/wiki/GUID_Partition_Table#Features

PR #2073, opened 2026-06-16
fix: populate per-network HCG for vrouters after 2026.1 upgrade

skrobul:
After upgrading to neutron 28 (2026.1), attaching a subnet to a router whose external gateway is on a vxlan-type network leaves the per-network unified `HA_Chassis_Group (neutron-<network_id>)` empty. External (baremetal) ports with OVN LSP `type=external` on that network reference this HCG; with it empty, no chassis owns them and ARP/routing breaks for all baremetal nodes on the network.

Root cause: for a vxlan-type external gateway neutron pins the Logical_Router to one chassis via `options:chassis` and creates a per-router HCG (`neutron-<router_id>`) carrying that chassis, but never sets ha_chassis_group on the gateway LRP. neutron's `link_network_ha_chassis_group` (called on internal LRP creation) bails at its [`if not gw_lrps[0].ha_chassis_group: return`](https://github.com/openstack/neutron/blob/4355bb2efe1940888809209572442c1ecda8655b/neutron/plugins/ml2/drivers/ovn/mech_driver/ovsdb/ovn_client.py#L2453) guard, so it never copies the chassis into `neutron-<network_id>`.

Fix: subscribe to `ROUTER_INTERFACE/AFTER_CREATE` at `PRIORITY_DEFAULT+1000` (after OVN's handler has created the LRP) with a new handler `link_vxlan_network_ha_chassis_group.`

The handler:

1. Detects a vxlan-type gateway by checking that the per-router HCG (`neutron-<router_id>`) exists and has non-empty `ha_chassis`; VLAN/FLAT type gateways have no router HCG and continue to be handled by neutron unchanged, even though we are not using them.
2. Calls `ovn_utils.sync_ha_chassis_group_network_unified` to populate the per-network HCG (`neutron-<network_id>`) from the router HCG chassis — this is what fixes external/baremetal ports that already reference it.
3. Anchors the internal router-interface LRP (`lrp-<port_id>`) to the same unified network HCG so the router port is also correctly owned.

Both operations run in a single OVN NB transaction. The `create_port_postcommit` hook is unaffected — the internal LRP does not exist at that point (it is created later, on the same `AFTER_CREATE` event OVN processes first).

### DIFF
--- a/python/neutron-understack/neutron_understack/neutron_understack_mech.py
+++ b/python/neutron-understack/neutron_understack/neutron_understack_mech.py
@@ -6,6 +6,7 @@ from keystoneauth1 import session as ks_session
 from neutron_lib import constants as p_const
 from neutron_lib.api.definitions import portbindings
 from neutron_lib.callbacks import events
+from neutron_lib.callbacks import priority_group
 from neutron_lib.callbacks import registry
 from neutron_lib.callbacks import resources
 from neutron_lib.plugins.ml2 import api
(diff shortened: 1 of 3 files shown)


nicholaskuechler (approved):

mfencik (approved):
LGTM

PR #2093, opened 2026-06-23
fix(routers/ovn): leftover uplink ports

skrobul:
## Fix OVN uplink LSP leak on router subnet detachment

When the last subnet is detached from a router (`openstack router remove subnet`), the ML2 driver is supposed to clean up the uplink it created — remove the trunk subport, delete the OVN localnet LSP, and delete the shared Neutron port. In practice the shared Neutron port and its OVN LSP (`<port-uuid>`) were left behind.

### Root causes

Two independent bugs combined to cause the leak:

**1. Cleanup anchored to a segment that no longer exists**

`_do_uplink_cleanup` started by looking up the dynamic VLAN segment for the network. On subnet detachment, Neutron releases that segment *before* the `ROUTER_INTERFACE AFTER_DELETE` event fires. The lookup returned `None` and the entire cleanup was skipped, including the shared port deletion.

**Fix:** anchor cleanup on the shared Neutron port (`uplink-<segment_id>`) instead — it persists until we delete it. The segment ID is derived from the port's name, eliminating the dependency on a live segment.

**2. OVO `.delete()` leaves the OVN LSP behind**

`shared_port.delete()` is an OpenStack Versioned Objects call that only removes the database row. It bypasses the ML2 mechanism manager, so the OVN mechanism driver's `delete_port_postcommit` never runs and the port's LSP (named by port UUID) is never removed.

**Fix:** added `delete_shared_port_lsp()` which explicitly calls `delete_lswitch_port(port_id, ...)` on the OVN NB IDL before the OVO delete, mirroring the existing `delete_uplink_port` pattern for the localnet LSP.

Both fixes apply to both deletion paths: subnet detachment (`ROUTER_INTERFACE AFTER_DELETE`) and router deletion (`PORT PRECOMMIT_DELETE`).

### Also included

- Documentation of the router interface lifecycle in `neutron-networking.md`, covering the uplink port concept, the two OVN LSPs it produces, the creation flow, and why two event handlers are needed with different port-count thresholds.

Closes https://rackspace.atlassian.net/browse/PUC-1741

### DIFF
--- a/docs/design-guide/neutron-networking.md
+++ b/docs/design-guide/neutron-networking.md
@@ -454,9 +454,17 @@ network over the physical fabric. This is done through an **uplink port**.
 
 #### The uplink port
 
-The uplink is a Neutron port plus a matching OVN localnet Logical Switch Port
-(LSP) named `uplink-<segment_id>`. It sits on the Network Node's leaf VLAN
-segment and gives OVN a way to send and receive that network's traffic over the
+The uplink is a Neutron port named `uplink-<segment_id>`. Creating it causes
+the OVN mechanism driver to create two OVN Logical Switch Ports (LSPs) on the
+network's logical switch:
+
+- `uplink-<segment_id>` — a `localnet`-type LSP created explicitly by understack;
+  it connects the logical switch to the physical network via the Network Node's
+  trunk VLAN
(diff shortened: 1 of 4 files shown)


stevekeay (commented):
LGTM but Milan should look at this because he is way more familiar with that code

mfencik (approved):
LGTM

PR #2095, opened 2026-06-23
feat(ironic-understack): stub hardware type for network devices

cardoe:
Add a basic stub hardware type to represent hardware network devices
that will live on the fabric and need their switch ports configured.

### DIFF
--- a/docs/design-guide/ironic.md
+++ b/docs/design-guide/ironic.md
@@ -1,0 +1,18 @@
+# Ironic
+
+## Custom Hardware Types
+
+UnderStack ships additional Ironic hardware types via the `ironic-understack`
+Python package, registered as `ironic.hardware.types` entry points.
+
+### netdev
+
+The `netdev` hardware type is a stub type for network devices (firewalls,
+load balancers) that Ironic tracks solely for Neutron physical port binding. It uses
+noop or no-* interfaces for everything except `network`, which is set to
+`neutron`. This means nodes of this type go through no deployment, inspection,
+BIOS, RAID, rescue, or firmware lifecycle — Ironic only manages their port
+bindings.
+
+To use it, add `netdev` to `enabled_hardware_types` in `ironic.conf` and
+ensure `noop` deploy and `neutron` network interfaces are also enabled.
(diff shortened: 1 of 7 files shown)


skrobul (commented):
LGTM once the dependency issues are resolved

python/ironic-understack/ironic_understack/tests/test_netdev_hardware.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_netdev_noop_interfaces():
+    hw = NetdevHardware()
+    assert hw.supported_console_interfaces == [noop.NoConsole]
+    assert hw.supported_firmware_interfaces == [noop.NoFirmware]
+    assert hw.supported_inspect_interfaces == [noop.NoInspect]
+    assert hw.supported_raid_interfaces == [noop.NoRAID]
+    assert hw.supported_rescue_interfaces == [noop.NoRescue]
+    assert hw.supported_storage_interfaces == [noop_storage.NoopStorage]
```
stevekeay, line 33: Is there a reason this name is inconsistent with the others?

docs/design-guide/ironic.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Custom Hardware Types
+
+UnderStack ships additional Ironic hardware types via the `ironic-understack`
+Python package, registered as `ironic.hardware.types` entry points.
+
+### netdev
+
+The `netdev` hardware type is a stub type for network devices (switches,
+routers) that Ironic tracks solely for Neutron physical port binding. It uses
```
stevekeay: This is confusing to me because we don't have switches or routers in Ironic - is this intended for firewalls and load balancers?  Would this also apply to our "infra" nodes?

stevekeay (approved):

python/ironic-understack/ironic_understack/tests/test_netdev_hardware.py
cardoe: Cause upstream the `noop` module doesn't export a `NoStorage` interface. There's a separate module nested under `storage` which exports one but it's called `NoopStorage`. Should probably be cleaned up for consistency upstream.

docs/design-guide/ironic.md
cardoe: I've updated the wording to `firewalls` and `load balancers` which is more correct.

It could apply to the infra nodes.

PR #2098, opened 2026-06-24
feat: adding device sync to Nautobot Operator

abhimanyu003:
Add device sync support to the Nautobot operator with full CRUD lifecycle (create/update/delete) following the same patterns as device types.

Sample ConfigMap/YAML files.

```yaml
---
id: "c3d4e5f6-a7b8-9012-cdef-123456789012"
name: "firewall-01"
device_type: "ASA5525-X"
role: "Infrastructure"
serial: "JMX2145L0XY"
asset_tag: "AST-FW-001"
status: "Active"
location: "dfw1"
rack: "dfw1-b01"
position: 38
face: "front"
comments: "Primary firewall for dfw1"
```

**Sample Config Map**

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nautobot-devices
  namespace: nautobot
data:
  device-switch-01.yaml: |
    ---
    id: "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
    name: "switch-01"
    device_type: "Nexus 9336C-FX2"
    role: "Infrastructure"
    serial: "FCW2345L0AB"
    asset_tag: "AST-SW-001"
    status: "Active"
    location: "iad3"
    rack: "iad3-a01"
    position: 40
    face: "front"
    comments: "Top of rack switch in iad3"
  device-switch-02.yaml: |
    ---
    id: "b2c3d4e5-f6a7-8901-bcde-f12345678901"
    name: "switch-02"
    device_type: "Nexus 9336C-FX2"
    role: "Infrastructure"
    serial: "FCW2345L0AC"
    asset_tag: "AST-SW-002"
    status: "Active"
    location: "iad3"
    rack: "iad3-a02"
    position: 40
    face: "front"
    comments: "Top of rack switch in iad3"
  device-firewall-01.yaml: |
    ---
    id: "c3d4e5f6-a7b8-9012-cdef-123456789012"
    name: "firewall-01"
    device_type: "ASA5525-X"
    role: "Infrastructure"
    serial: "JMX2145L0XY"
    asset_tag: "AST-FW-001"
    status: "Active"
    location: "dfw1"
    rack: "dfw1-b01"
    position: 38
    face: "front"
    comments: "Primary firewall for dfw1"
```


haseebsyed12 (approved):

go/nautobotop/internal/controller/nautobot_controller.go
```diff
@@ -111,6 +111,7 @@ func (r *NautobotReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		{name: "vlan", configRefs: nautobotCR.Spec.VlanRef, syncFunc: r.syncVlan},
 		// depends on: namespace, rir, location, vlanGroup, vlan, tenant, tenantGroup, role
 		{name: "prefix", configRefs: nautobotCR.Spec.PrefixRef, syncFunc: r.syncPrefix},
+		{name: "device", configRefs: nautobotCR.Spec.DeviceRef, syncFunc: r.syncDevice},
```
nidzrai: device sync is currently at the end of the sync order  but cluster sync runs earlier on line 99 and tries to assign devices during cluster sync. Since clusters depend on devices already existing, should device sync run before cluster sync? Otherwise new devices may not exist yet when cluster assignment happens.

go/nautobotop/config/crd/bases/sync.rax.io_nautobots.yaml
```diff
@@ -117,6 +117,31 @@ spec:
                   - configMapSelector
                   type: object
                 type: array
+              deviceRef:
+                items:
+                  description: ConfigMapRef defines a reference to a specific ConfigMap
```
geetikabatra, line 122: Nit: can we improve the description here? It sounds slightly unintuitive. I can't make sens eof this sentence, 'ie' reference to a specific ConfigMap' of what?

go/nautobotop/config/crd/bases/sync.rax.io_nautobots.yaml
```diff
@@ -117,6 +117,31 @@ spec:
...
+                  properties:
+                    configMapSelector:
+                      description: The name of the ConfigMap resource being referred
```
geetikabatra, line 125: The description says this selector "is" a name, but then on line 132 we have a separate key field — so it's unclear what the outer description is even referring to. This still looks confusing to me.

go/nautobotop/helm/crds/clients.yaml
```diff
@@ -117,6 +117,31 @@ spec:
                   - configMapSelector
                   type: object
                 type: array
+              deviceRef:
+                items:
+                  description: ConfigMapRef defines a reference to a specific ConfigMap
+                  properties:
+                    configMapSelector:
+                      description: The name of the ConfigMap resource being referred
```
geetikabatra, line 125: Same confusion as I described in [/config/crd/bases/sync.rax.io_nautobots.yaml](https://github.com/rackerlabs/understack/pull/2098/changes#diff-441145da41ea916e717e2704519efaf4c227669bd860e60e956f92e9e34786e8)

go/nautobotop/internal/controller/nautobot_controller.go
geetikabatra: I think this might be a problem later on. the current order relies on programmers  to manually remember the order of dependency when adding new resources. I don't think that should be the case. We should use a  DAG(Directed Acyclic Graphs) here and do a topological sort instead. This would guarantee correct processing order instead of relying on the order in the list. For example, if devices need to be processed before cluster sync, the DAG would enforce that automatically — rather than depending on someone remembering to place device above cluster.

geetikabatra (changes requested):

go/nautobotop/internal/controller/nautobot_controller.go
abhimanyu003: Thanks for suggestion added a dag ( topolocial sort ) its following this algo 
https://en.wikipedia.org/wiki/Topological_sorting
Slightely tweaked to account for names and have deterministic output

PR #2104, opened 2026-06-30
chore(keystone): cleanup patches - fixed in upstream

haseebsyed12:
https://review.opendev.org/c/openstack/openstack-helm/+/959251


cardoe (changes requested):
So we need to to re-enable letting OpenStack Helm create those things. Cause right now it's disabled and we were creating them with these resources. So removing these resources is good but then we need to enable it.

haseebsyed12:
> So we need to to re-enable letting OpenStack Helm create those things. Cause right now it's disabled and we were creating them with these resources. So removing these resources is good but then we need to enable it.

removed these two configs from values.yaml
secret_credential_keys: false
secret_fernet_keys: false

[Now these two keys are by default enabled](https://opendev.org/openstack/openstack-helm/src/branch/master/keystone/values.yaml#L1143)

cardoe (approved):
I hope my original commit is correct.

PR #2109, opened 2026-07-01
feat: Implement vrf router with auto-allocating evpn_vni attribute

stevekeay:
This creates a standalone plugin for 2026.1 to provide the vrf router behaviour we need, including auto-assignment of VNIs from a configured range.

Some of this overlaps work in progress upstream.   I have intentionally mirrored the --evpn-vni API surface of the evpn plugin.  We are not using any of the actual evpn plugin code but (in theory) once it gains feature parity, we should be able to switch over to it without affecting our users.

This got fairly big.  I was thinking it might be cleaner to break it out into a separate package (that we would install separately via pip install in the container) but I wanted to solicit opinions from the team.

To use this:
- we can roll neutron back to stock 2026.1 (removing the stuff we cherry-picked)
- change the neutron.conf `service_plugins` removing `evpn` and adding `understack_vni`
- Add a `[understack_vni]` section like `vni_ranges = 10000:19999`


python/neutron-understack/neutron_understack/config.py
```diff
@@ -139,6 +161,11 @@ def register_l3_svc_cisco_asa_opts(config):
     config.register_opts(_l3_svc_cisco_asa_opts, _OPT_GRP_L3_SVC_CISCO_ASA)
 
 
+def register_understack_vni_opts(config):
+    with contextlib.suppress(cfg.DuplicateOptError):
```
cardoe: Why was this triggering? This one is a bit concerning to me that we'll get bad data.

stevekeay: Yes, this should not have been present in the production code - I believe LLM added it to allow it to register different options for various test cases.  I have removed it in cbbc38a.

python/neutron-understack/neutron_understack/api/definitions/understack_vni.py
```diff
@@ -0,0 +1,37 @@
...
+
+EVPN_VNI = "evpn_vni"
+
+RESOURCE_ATTRIBUTE_MAP = {
+    COLLECTION_NAME: {
+        EVPN_VNI: {
+            "allow_post": True,
+            "allow_put": False,
+            "convert_to": converters.convert_to_int_if_not_none,
+            "default": 0,
```
skrobul, line 23: should that be `0` or `ATTR_NOT_SPECIFIED` ? I am concerned if having it as 0 will not prevent non-admin users from creating routers

python/neutron-understack/pyproject.toml
```diff
@@ -37,13 +37,24 @@ dependencies = [
...
+understack_vni = "neutron_understack.l3_router.vrf:UnderstackVniPlugin"
+
+[project.entry-points."oslo.policy.policies"]
+neutron-understack = "neutron_understack.conf.policies:list_rules"
+
+[project.entry-points."neutron.policies"]
+neutron-understack = "neutron_understack.conf.policies:list_rules"
+
+[project.entry-points."neutron.db.alembic_migrations"]
+neutron-understack = "neutron_understack.db.migration:alembic_migrations"
```
skrobul, line 57: that seems to be missing from `__init__.py`. So it probably needs to have `from neutron_understack.db.migration import alembic_migrations` or maybe this will work(untested) 

```suggestion
neutron-understack = "neutron_understack.db.migration.alembic_migrations"
```

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
```diff
@@ -0,0 +1,238 @@
...
+            context.session.add(allocation)
+            context.session.flush()
+            return never_used_vni
+
+        released = self._find_released_allocation(context, ranges)
+        if released:
+            released.router_id = router_id
+            released.released_at = None
+            context.session.flush()
+            return released.vni
```
skrobul, line 158: While slim chances with our workload, there is a risk of concurrent auto allocations picking the same never-used VNI. The faster one gets it, the loser will get 500 error. Consider catching a duplicate violation and retrying for more graceful handling:


```suggestion
  for _attempt in range(3):
        never_used_vni = self._find_never_used_vni(context, ranges)
        if never_used_vni is not None:
            try:
                allocation = vni_models.UnderstackRouterVNIAllocation(
                    vni=never_used_vni,
                    router_id=router_id,
                )
                context.session.add(allocation)
                context.session.flush()
                return never_used_vni
            except db_exc.DBDuplicateEntry:
                context.session.rollback()
                continue  # another worker won; recompute candidate

        released = self._find_released_allocation(context, ranges)
        if released:
            released.router_id = router_id
            released.released_at = None
            context.session.flush()
            return released.vni

        raise UnderstackVNINoAvailable(ranges=format_vni_ranges(ranges))
 raise UnderstackVNINoAvailable(ranges=format_vni_ranges(ranges))
```

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
```diff
@@ -0,0 +1,238 @@
...
+
+        raise UnderstackVNINoAvailable(ranges=format_vni_ranges(ranges))
+
+    def _allocate_specific_vni(self, context, router_id, vni):
```
skrobul, line 162: similarly to the above, it might be worth wrapping the `context.session.flush()` with catch for duplicates, although the retry would not make any sense here

```python
 try:
        context.session.flush()
    except db_exc.DBDuplicateEntry:
        context.session.rollback()
        raise UnderstackVNIInUse(vni=vni, router_id="unknown")
```
or something like that

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
skrobul: There is also `oslo_db_api.wrap_db_retry` decorator which may work here

```python
@oslo_db_api.wrap_db_retry(
    max_retries=3,
    retry_on_deadlock=True,
    exception_checker=lambda exc: isinstance(exc, db_exc.DBDuplicateEntry),
)
```

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
```diff
@@ -0,0 +1,238 @@
...
+from neutron_lib import constants as n_const
+from neutron_lib import exceptions as n_exc
+from oslo_config import cfg
+from oslo_utils import timeutils
+
+from neutron_understack.db import understack_vni as vni_models
+
+MIN_VNI = getattr(n_const, "MIN_VXLAN_VNI", 1)
+MAX_VNI = n_const.MAX_VXLAN_VNI
+DEFAULT_VNI_RANGES = [f"{MIN_VNI}:{MAX_VNI}"]
```
skrobul, line 11: unused?

python/neutron-understack/neutron_understack/api/definitions/understack_vni.py
stevekeay: I am a bit hazy on the logic for collections, but I think we want non-admin users to create routers, we just don't want them to be able to choose a VNI.

A value of zero is explicitly interpreted as "please auto-assign" in the assignment hook, so the value that actually gets persisted should be non-zero.

Technically it may be more correct to use the sentinel value but I thought this was going to the DB layer, I'm not sure what the best answer is.

skrobul: The python/neutron-understack/neutron_understack/conf/policies/evpn.py has:

```
   policy.DocumentedRuleDefault(
        name="create_router:evpn_vni",
        check_str=base.ADMIN,
        scope_types=["project"],
        description="Specify ``evpn_vni`` attribute when creating a router",
        operations=ACTION_POST,
...
```

which will translate to `"create_router:evpn_vni": "rule:admin_only"` inside a default policy (similarly to [stock one](https://docs.openstack.org/neutron/latest/configuration/policy-sample.html) and if the user provides any value of the attribute in the request, it will match the rule, effectively requiring admin privileges, so we may need to adjust `base.ADMIN` to sth else.

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
stevekeay: I'm not sure we can just retry here because this "PRECOMMIT" event happens inside a transaction, and by the time we notice the duplicate the transaction will already have failed.  The entire transaction needs to be retried, not just our bit.  (I don't want to get into nested transactions with mariadb.)

Good news is though, assuming that it does indeed raise a `DBDuplicateEntry`, I think this is already covered by the caller.

neutron/l3_db.py around line 310 defines create_router() and it's decorated with @db_api.retry_if_session_inactive.   cahtgpt assures me that our exception is listed in is_retriable(), so the wrapper should rollback and retry the failed exception.

stevekeay: I added a test, although it's not that meaningful to test this in isolation.

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
stevekeay: removed.

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
stevekeay: I *think* we're already covered by the same retry logic discussed earlier:

We query for an existing VNI allocation using the requested number, if it is found to be in use then we raise the exception.  Note that this read is serialised using "for update".

If two threads race for the VNI, the loser will attempt to insert a duplicate row which the DB will reject.  The transaction will get rolled back and retried. On the second attempt, its read should see the first (now committed) allocation, and raise the proper error.

python/neutron-understack/pyproject.toml
stevekeay: This appears to be the syntax for the migration runner.  Neutron has the same format in its pyproject.yaml.  Also, it did seem to work when I installed it in a devstack instance.

python/neutron-understack/neutron_understack/api/definitions/understack_vni.py
stevekeay: These are the results of my manual testing:

If a non-admin user provides a valid VNI, it fails:
```
ubuntu@devstack-ovn:~$ OS_CLOUD=devstack-alt-member openstack router create --flavor understack-vrf-demo  --extra-property type=int,name=evpn_vni,value=10099 steve-test-b
ForbiddenException: 403: Client Error for url: http://192.168.99.10/networking/v2.0/routers, ((rule:create_router and rule:create_router:flavor_id) and rule:create_router:evpn_vni) is disallowed by policy
```

(The above succeeds when performed as an admin user.)

If a non-admin user provides a VNI of zero, it still fails:
```
ubuntu@devstack-ovn:~$ OS_CLOUD=devstack-alt-member openstack router create --flavor understack-vrf-demo  --extra-property type=int,name=evpn_vni,value=0 steve-test-b
ForbiddenException: 403: Client Error for url: http://192.168.99.10/networking/v2.0/routers, ((rule:create_router and rule:create_router:flavor_id) and rule:create_router:evpn_vni) is disallowed by policy
```

(The above succeeds when performed as an admin user.)


If any user created a router but does not include a VNI, the call succeeds and a VNI is allocated from the configured range:

```
ubuntu@devstack-ovn:~$ OS_CLOUD=devstack-alt-member openstack router create --flavor understack-vrf-demo  steve-test-c
+---------------------------+--------------------------------------+
| Field                     | Value                                |
+---------------------------+--------------------------------------+
| admin_state_up            | UP                                   |
| availability_zone_hints   |                                      |
| availability_zones        |                                      |
| created_at                | 2026-07-08T13:22:11Z                 |
| description               |                                      |
| enable_default_route_bfd  | False                                |
| enable_default_route_ecmp | False                                |
| enable_ndp_proxy          | None                                 |
| enable_snat               | True                                 |
| evpn_vni                  | 10007                                |
| external_gateway_info     | null                                 |
| external_gateways         | []                                   |
| flavor_id                 | 00035dcb-968b-4be6-9057-026f4069ee1a |
| id                        | 92df1c90-78c4-4412-8eb4-e1c94288a4ef |
| name                      | steve-test-c                         |
| project_id                | fa2408453dfa4a658fca0aff301b94ba     |
| revision_number           | 1                                    |
| routes                    |                                      |
| status                    | ACTIVE                               |
| tags                      |                                      |
| updated_at                | 2026-07-08T13:22:11Z                 |
+---------------------------+--------------------------------------+
```

cardoe: So I had to walk the code and I still was confused. So I had to have Claude walk it with me to understand and run it in a little repl to test. It's the way for the non-default value (which the default value is 0). So the behavior is actually correct as Steve coded it. But holy cow that feels like a foot gun.

python/neutron-understack/pyproject.toml
cardoe: Yeah this is the correct syntax. It's effectively `from neutron_understack.db.migration import alembic_migrations as neutron-understack` (the dash is really an underscore) into the namespace.

python/neutron-understack/neutron_understack/l3_router/understack_vni_db.py
cardoe: So I think this is a good point out but I looked and Steve's right. It happens in a transaction and has `DBDuplicateEntry`.

cardoe (approved):
I think it would be good to get some operator docs.

skrobul (approved):

python/neutron-understack/neutron_understack/api/definitions/understack_vni.py
skrobul: Cool cool - I misunderstood the requirements and thought we want to allow users to select the VNI, but after thinking about it, it would be bad idea. Thanks for testing

cardoe: dang @skrobul was right here. We need to do `ATTR_NOT_SPECIFIED` otherwise we got a VNI allocated for all routers, including those we don't want them to be allocated for.

PR #2136, opened 2026-07-10
feat(dnsmasq): add --conf-poll feature and upgrade to Debian forky

skrobul:
Bump the dnsmasq container to `debian:forky-20260623-slim` (dnsmasq 2.93-1), which is close enough to our dev tree that both vendored patches now apply cleanly, and rebase dhcp-allowed-srvids.patch onto that version.

Add conf-poll.patch, a new dnsmasq feature (`--conf-poll` / `--conf-poll-interval=<seconds>`, default 1s) written to solve a real problem in our Ironic setup: dnsmasq relies on `inotify` to notice changes to `--hostsdir`, `--dhcp-hostsdir` and `--dhcp-optsdir`. When those directories live on an NFS share and are modified or created by a different NFS client (i.e. not the machine running dnsmasq itself), the local kernel never gets an inotify event for it, so dnsmasq silently keeps serving stale DNS/DHCP data until it's sent a SIGHUP or restarted.

`--conf-poll` makes dnsmasq additionally `stat()` those directories on a timer (every `--conf-poll-interval` seconds) and reload anything that's new, changed, or removed, using the exact same reload logic as an inotify event or SIGHUP would trigger. It runs alongside inotify, not instead of it, so on non-NFS setups where inotify works fine it's just a harmless no-op backup check. It's opt-in and disabled by default, so existing deployments are unaffected unless the option is explicitly set. Deleting a file from `--dhcp-hostsdir/--dhcp-optsdir` intentionally still does not remove the DHCP config it added, since that already matches how those two options have always behaved in stock dnsmasq.

Related: https://rackspace.atlassian.net/browse/PUC-1729

### DIFF
--- a/components/ironic/dnsmasq-ss.yaml
+++ b/components/ironic/dnsmasq-ss.yaml
@@ -47,7 +47,8 @@ spec:
               add: ["NET_ADMIN", "NET_RAW"]
           args:
             - /usr/sbin/dnsmasq
-            - '--no-daemon'
+            - "--no-daemon"
+            - "--conf-poll"
           envFrom:
             - configMapRef:
                 name: ironic-dnsmasq
--- a/containers/dnsmasq/Dockerfile
+++ b/containers/dnsmasq/Dockerfile
@@ -1,11 +1,11 @@
 # syntax=docker/dockerfile:1
-FROM debian:bookworm-20240408-slim AS builder
+FROM debian:forky-20260623-slim AS builder
 
 ENV DEBFULLNAME="Marek Skrobacki"
 ENV DEBEMAIL="marek.skrobacki@rackspace.co.uk"
 ENV QUILT_PATCHES=debian/patches
 
-RUN echo "deb-src http://deb.debian.org/debian bookworm main" > /etc/apt/sources.list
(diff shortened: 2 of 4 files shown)


skrobul:
For future reference:

> [!NOTE]
> This patch is based on the code from dnsmasq fork stored in  https://github.com/skrobul/dnsmasq/tree/conf-poll-on-v2.93

Please note that this branch was based off the `v2.93` upstream tag rather than a `master`. This is to ensure that we use the exact same version as is shipped with underlying Debian forky package.

It's not a direct lift & shift from the [relevant commit](https://github.com/skrobul/dnsmasq/commit/9c95c0f9c3cfcff29e9272e584a1b249ebde5fea) either.

Given that this is a second patch (after [`dhcp-allowed-srvids.patch`](https://github.com/rackerlabs/understack/blob/dnsmasq-polling/containers/dnsmasq/dnsmasq/dhcp-allowed-srvids.patch)) that needs to be applied, it required adjustments to apply clearly. Specifically, the conflict is in `src/option.c` where  the convention is to use next-available option ID. In our case we used  392 for serverid patch and 393/394 for config polling.

If this patch ever needs to be ported to a newer version of dnsmasq, these numbers are likely to be consumed upstream so will need adjustment.

cardoe (approved):

PR #2137, opened 2026-07-10
fix(understack-workflows): ensure that Nautobot custom fields are strings

cardoe:
Nautobot custom fields must be strings so ensure that we string-ify
values for them to sync and compare correctly.


mfencik (approved):
LGTM

khackworth (approved):
ship it

PR #2156, opened 2026-07-20
feat: use primary galera host for MariaDB from OpenStack services

cardoe:
Per the discussion on the OpenStack mailing list, multiple people have
reported that the proper behavior is to have the services connect to the
galera primary node which is what this does.

https://lists.openstack.org/archives/list/openstack-discuss@lists.openstack.org/message/7D45OTQCEOVP7DQTDYNWLGCWOTFALWIF/

### DIFF
--- a/components/cinder/values.yaml
+++ b/components/cinder/values.yaml
@@ -33,6 +33,9 @@ conf:
 # values, but should include all endpoints
 # required by this chart
 endpoints:
+  oslo_db:
+    hosts:
+      default: mariadb-primary
   oslo_messaging:
     statefulset:
       replicas: 3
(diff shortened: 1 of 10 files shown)


nicholaskuechler:
If we put it in https://github.com/rackerlabs/understack/blob/main/components/images-openstack.yaml we wouldn't have to repeat it 50 times.

cardoe:
> If we put it in https://github.com/rackerlabs/understack/blob/main/components/images-openstack.yaml we wouldn't have to repeat it 50 times.

Which is documented for image references. We can redo all that to not be image references but it'll take a bit.

mfencik (approved):
LGTM

PR #2166, opened 2026-07-23
feat(ovn): add another HCG cleanup that was missed

cardoe:
There are still some networks and ports left dangling which prevented
some HCGs from being cleaned up. So compare against the live running
Neutron for the networks so that we can clean up those networks and
ports before cleaning up the orphaned HCGs.


cardoe:
https://review.opendev.org/c/openstack/neutron/+/998553 should include this case in the `neutron-ovn-db-sync-util` repair case.

haseebsyed12 (approved):

PR #2172, opened 2026-07-29
feat(ironic): wire up dynamic portgroup via TBN

cardoe:
Utilize Ironic's trait based networking to create a trait called
NETWORK_PORTGROUP which will create a portgroup of the two ports that
are in the category=network. It also sets up a default trait attachment
operation to the port with category=network, which mirrors the current
code hack we have in our Ironic repo allowing it to be dropped for us
using TBN.


stevekeay (approved):
This exercises new features and previously-unused OpenStack code paths.  So we should be ready to check for regressions as we roll this out.

PR #2180, opened 2026-07-31
fix(neutron-understack): release dynamic VLAN segment on tenant port delete

cardoe:
A baremetal port on a tenant network deleted directly (skipping the
update_port_postcommit bound->unbound transition) never released its
dynamic VLAN segment, since _tenant_network_port_cleanup only ran on
that transition. The leaked segment kept its VLAN unavailable in
ml2_vlan_allocations forever, and going stale there let a later
rebuild hand the same VLAN to an unrelated network.


haseebsyed12 (approved):

mfencik (approved):
LGTM

PR #2186, opened 2026-08-04
fix: Nautobot Secret Refs annotation as it is required before Creation

### LINKED ISSUE #2183: `NautobotSecretRef` is optional in CRD, reconciler fails when it is unset
`NautobotSpec.NautobotSecretRef` has no +kubebuilder:validation:Required or +kubebuilder:default marker in nautobot_types.go. This means the Kubernetes API server accepts a Nautobot CR with no nautobotSecretRef set at all.

However, Reconcile → getAuthTokenFromSecretRef assumes it’s always populated and hard-fails with:

`failed to get nautobot auth token {"error": "nautobotSecretRef \"\" is missing a namespace"}`

Expected behavior

• If nautobotSecretRef.Name is empty, the reconciler should skip gracefully: update .Status.Ready = false and .Status.Message with a clear “not configured” message, then requeue — instead of returning a raw error.

• If nautobotSecretRef.Namespace is empty but .Name is set, default to the CR’s own namespace rather than failing.

This issue first appeared in `make test`. Looks like it appears for the resource type "Prefix"

Raw logs

```
2026-08-03T16:21:31+05:30     INFO    resource needs sync     {"resource": "cluster", "reason": "data changed"}
  2026-08-03T16:21:31+05:30     INFO    resource needs sync     {"resource": "prefix", "reason": "data changed"}
  2026-08-03T16:21:31+05:30     ERROR   failed to get nautobot auth token       {"error": "nautobotSecretRef \"\" is missing a namespace"}
  github.com/rackerlabs/understack/go/nautobotop/internal/controller.(*NautobotReconciler).Reconcile
        /Users/geetika.batra/understack/go/nautobotop/internal/controller/nautobot_controller.go:183
  github.com/rackerlabs/understack/go/nautobotop/internal/controller.init.func1.1.2
        /Users/geetika.batra/understack/go/nautobotop/internal/controller/nautobot_controller_test.go:77
  github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3
        /Users/geetika.batra/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/node.go:475
  github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3
        /Users/geetika.batra/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/suite.go:894
  [FAILED] in [It] - /Users/geetika.batra/understack/go/nautobotop/internal/controller/nautobot_controller_test.go:80 @ 08/03/26 16:21:31.434
  << Timeline

  [FAILED] Unexpected error:
      <*errors.errorString | 0x498001636b40>: 
      nautobotSecretRef "" is missing a namespace
      {
          s: "nautobotSecretRef \"\" is missing a namespace",
      }
  occurred
  In [It] at: /Users/geetika.batra/understack/go/nautobotop/internal/controller/nautobot_controller_test.go:80 @ 08/03/26 16:21:31.434
------------------------------

Summarizing 1 Failure:
  [FAIL] Nautobot Controller When reconciling a resource [It] should successfully reconcile the resource
  /Users/geetika.batra/understack/go/nautobotop/internal/controller/nautobot_controller_test.go:80

Ran 1 of 1 Specs in 6.040 seconds
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestControllers (6.04s)

```

geetikabatra:
If nautobotSecretRef.Name is empty, the reconciler should skip gracefully: update .Status.Ready = false and .Status.Message with a clear “not configured” message, then requeue — instead of returning a raw error.

• If nautobotSecretRef.Namespace is empty but .Name is set, default to the CR’s own namespace rather than failing.

Fixes: [2183](https://github.com/rackerlabs/understack/issues/2183)

### DIFF
--- a/go/nautobotop/internal/controller/nautobot_controller.go
+++ b/go/nautobotop/internal/controller/nautobot_controller.go
@@ -172,10 +172,30 @@ func (r *NautobotReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{RequeueAfter: requeueAfter}, nil
 	}
 
-	// Create Nautobot client
+	// Validate nautobotSecretRef before attempting auth
+	if nautobotCR.Spec.NautobotSecretRef.Name == "" {
+		log.Info("nautobotSecretRef.Name is not configured, skipping sync")
+		nautobotCR.Status.Ready = false
+		nautobotCR.Status.Message = "nautobotSecretRef is not configured: secret name is required"
+		if err := r.Status().Update(ctx, &nautobotCR); err != nil {
+			log.Error(err, "failed to update status")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+	}
+
+	// Create Nautobot client. The not-configured case (empty secret name) is
+	// handled by the validation above; any error here is a genuine failure
+	// (RBAC, wrong namespace, API error, missing keys) and must be surfaced
+	// so backoff and the reconcile-error metric kick in.
 	username, token, err := r.getAuthTokenFromSecretRef(ctx, nautobotCR)
 	if err != nil {
 		log.Error(err, "failed to get nautobot auth token")
+		nautobotCR.Status.Ready = false
+		nautobotCR.Status.Message = fmt.Sprintf("failed to get authentication token: %v", err)
+		if statusErr := r.Status().Update(ctx, &nautobotCR); statusErr != nil {
+			log.Error(statusErr, "failed to update status after auth error")
+		}
 		return ctrl.Result{}, err
 	}
 	nautobotURL := fmt.Sprintf("http://%s.%s.svc.cluster.local/api", nautobotCR.Spec.NautobotServiceRef.Name, nautobotCR.Spec.NautobotServiceRef.Namespace)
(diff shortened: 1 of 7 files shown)


go/nautobotop/internal/controller/nautobot_controller.go
```diff
@@ -521,30 +538,47 @@ func (r *NautobotReconciler) syncDevice(ctx context.Context,
...
+	}
+
+	// Default namespace: use NautobotServiceRef.Namespace as a fallback
+	// (since the CRD is cluster-scoped and has no inherent namespace)
+	namespace := ""
+	if ref.Namespace != nil && *ref.Namespace != "" {
+		namespace = *ref.Namespace
+	} else if nautobotCR.Spec.NautobotServiceRef.Namespace != "" {
+		namespace = nautobotCR.Spec.NautobotServiceRef.Namespace
+	}
```
abhimanyu003, line 558: Can be done like this, in that way we don't need else if
```go
namespace := nautobotCR.Spec.NautobotServiceRef.Namespace
if ref.Namespace != nil && *ref.Namespace != "" {
	namespace = *ref.Namespace
}
```

geetikabatra: Thanks! Will change

cardoe:
### Code review

Found 3 issues:

**1. `required` on `NautobotSecretRef` doesn't enforce a name, and breaks the sample manifest.** The marker only requires the key's presence — nested `name` is still `omitempty` (unlike `configMapSelector`, which nests `required: [name]`), so `nautobotSecretRef: {}` still validates. And `config/samples/sync_v1alpha1_nautobot.yaml` has an empty `spec:`, so it'll now be rejected. Either update the sample or move `Required` onto `SecretKeySelector.Name`.

https://github.com/rackerlabs/understack/blob/8c23494b4c84210713a861f095fb684944dcf4b6/go/nautobotop/api/v1alpha1/nautobot_types.go#L35-L38

https://github.com/rackerlabs/understack/blob/8c23494b4c84210713a861f095fb684944dcf4b6/go/nautobotop/config/samples/sync_v1alpha1_nautobot.yaml#L6-L9

**2. Auth-fetch errors get swallowed as "not configured."** Any `getAuthTokenFromSecretRef` error — RBAC, wrong namespace, API hiccup, missing keys — becomes `"authentication not configured"` with a `nil` return, hiding it from backoff and the reconcile-error metric. Inconsistent with the rest of `Reconcile`, which returns hard errors. Keep the soft path only for the actual not-configured case.

https://github.com/rackerlabs/understack/blob/8c23494b4c84210713a861f095fb684944dcf4b6/go/nautobotop/internal/controller/nautobot_controller.go#L187-L193

**3. Namespace fallback contradicts the field's doc comment.** `SecretKeySelector.Namespace` says it "defaults to the namespace of the referent," but the code falls back to `NautobotServiceRef.Namespace`. Defensible for a cluster-scoped CR, but update the comment (and PR description).

https://github.com/rackerlabs/understack/blob/8c23494b4c84210713a861f095fb684944dcf4b6/go/nautobotop/api/v1alpha1/secret_type.go#L10-L16

geetikabatra:
This is ready to merge.

cardoe (approved):

PR #2190, opened 2026-08-04
docs: add documentation layout audit and proposal

cardoe:
Records the current docs layout, its problems, and a phased proposal to
restructure it around the project's three audiences: operators, contributors
and evaluators.

The headline finding is that contributors and evaluators have no front door:
there is no contributor section and no CONTRIBUTING.md, so developer docs live
in 58 unlinked markdown files outside docs/, and Overview is a grab-bag rather
than an evaluator's on-ramp. The proposal is phased so the nav rewrite that adds
the front doors needs no file moves, with moves deferred and governed by one
rule: move a file only when its audience changes.

This is a planning artifact, to be deleted once the phases land or are rejected.
It lives at the repository root because every file under docs/ is published.

### DIFF
--- a/docs-layout-audit.md
+++ b/docs-layout-audit.md
@@ -1,0 +1,349 @@
+# Documentation Layout Audit
+
+**Status:** proposal, for team review. This file is a planning artifact, not
+product documentation. Delete it once the phases below have landed or been
+rejected.
+
+It lives at the repository root rather than under `docs/` on purpose: every file
+under `docs/` has to be added to `nav:` in `properdocs.yml` and gets published to
+the site, and this is not something a reader of the docs site should ever find.
+
+## Why
+
+Does the current documentation layout serve the audiences the project has?
+
+- **system operators** deploying and troubleshooting the whole system,
+- **DC techs** working with individual machines,
+- **network operations** working with network configuration,
+- **users** consuming the cloud (tenants driving the OpenStack APIs and CLI),
+- **contributors** changing the code.
+
+Short answer: the layout has drifted in specific, fixable ways, and it fronts
+none of these five audiences well. Contributors have no entry point at all.
+Users have a thin `User Guide`, but its consumer readers are not who the
+surrounding operator docs address. And the three operator roles share a single
+undifferentiated `Operator Guide`: it does not distinguish deploying and
+troubleshooting the whole system from working on individual machines from
+configuring the network — it hands all three the same flat page list.
+
+## What is there now
+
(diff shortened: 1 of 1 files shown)


mfencik (approved):
LGTM

docs-layout-audit.md
```diff
@@ -0,0 +1,305 @@
...
+## Why
+
+Does the current documentation layout serve the three audiences the project has?
+
+- **operators** deploying and running UnderStack,
+- **contributors** changing the code,
+- **evaluators** deciding whether UnderStack fits.
+
+Short answer: the layout has drifted in specific, fixable ways, and two of the
+three audiences have no front door at all.
```
geetikabatra: which two, if I may ask?

docs-layout-audit.md
```diff
@@ -0,0 +1,305 @@
...
+
+It mixes three audiences: hardware **schema reference** (`device-types.md`,
+`hardware-traits.md`, `flavors.md`), an architecture overview, a generated oslo
+config dump, and `add-remove-app.md`, which is a contributor how-to (and carries
+the same staleness as `component-overview.md`). Its front door, `intro.md`, is
+three lines.
+
+### Duplication that costs readers
+
+- **Argo Workflows is explained three times**: `docs/component-argo-workflows.md`,
```
geetikabatra, line 87: I agree!

docs-layout-audit.md
```diff
@@ -0,0 +1,305 @@
...
+  `examples/*/README.md` — for zero content benefit.
+- `user-guide/` also keeps its path, nav-nested under Operations.
+- **`design-guide/` is the only directory that dies**, because its contents
+  genuinely split three ways.
+
+### Phasing
+
+Each phase is a separate reviewable PR.
+
+- **Phase 1 — nav rewrite plus the three front doors.** This is the phase that
```
geetikabatra: I think, we should also review what parts are no longer valid and need to be removed from the documentation.

geetikabatra:
Apart from a comment or two, it looks good to me otherwise.

docs-layout-audit.md
cardoe: Fair point.

docs-layout-audit.md
cardoe: Contributors. It was also suppose to be architectural or deep dive into areas. But Claude reworded me a bit. I'm going to fix that.

cardoe:
Ah merge when ready used Milan's review.

PR #2191, opened 2026-08-04
docs: add operator upgrade notes generated from release-note fragments

cardoe:
Adds operator-facing upgrade notes under `docs/release-notes/` — what a deployment has to *do* to move between versions: deploy repo changes, new or removed secrets, one-time manual steps, whether rollback is possible. Deliberately not a changelog; the merged PR list already lives on the releases page.

The notes are assembled by [scriv](https://scriv.readthedocs.io/) from fragments, so nobody edits a versioned page by hand and nobody types a version number.

## Writing a note

```bash
scriv create   # writes changelog.d/<date>_<you>_<branch>.md
```

Uncomment the section that fits, describe what the operator has to do, commit it in the same PR as the change, and label the PR `upgrade-impact`. Fragment content is copied into the page **verbatim**, so a note can be as long as the instructions need — numbered steps, before-and-after YAML, fenced command blocks.

`release-note-check.yaml` requires a fragment on PRs flagged upgrade-impacting (label, or `!` in the title) and short-circuits to success on everything else, so routine PRs see no new friction.

**Most changes need no fragment.** Tags are cut frequently and usually carry nothing an operator must act on. A release with no fragments gets no section and needs no docs change at all — empty sections teach operators these pages aren't worth opening.

## What happens at release time

`git tag && git push` is the whole procedure. On a `vX.Y.Z` tag, `release-notes.yaml`:

1. Collects the fragments into the series page and opens a `[Bot] docs:` PR against `main`.
2. Creates the GitHub release with auto-generated notes, linking the upgrade notes **only if there are any**.

Because `understack_ref` defaults to `HEAD` and the site publishes `main`, `make docs` also renders pending fragments into a generated `Unreleased` page, so deployments tracking `main` see a note as soon as its change lands.

## Two decisions worth reviewing

**Collection runs against the tagged commit, not the tip of `main`.** If it ran against `main`, a fragment merged in the minutes between tagging and the workflow finishing would be filed under a release that doesn't contain it. At ~1.5 merges/day that window is small, which is what makes it nasty — it would bite once, months later, and look inexplicable.

**Consequence: rendered notes land one commit *after* the tag,** so the tag doesn't contain its own notes. This is fine here: the site publishes from `main` and nothing builds docs from a tag checkout. The "Upgrading?" link in the release body 404s until the bot's PR merges, which is why that PR says to merge promptly.

## Monorepo filtering

GitHub's note generator can only filter by label and author — there is no path filtering. `labeler.yml` turns "this PR only touches something released on its own tag" into a `skip-changelog` label that `release.yml` excludes, so `understackctl` and friends are announced by their own release rather than twice.

The label is named for its effect rather than for one class of subtree, so it covers the CLI utilities and operators (`understackctl`, `nautobotop`, `dexop`, `ironic-hardware-exporter`, `kubectl-us-net`) alongside the images that carry their own tag namespace (`containers/ironic-ipxe/`, `ironic-images/`) — and can be applied by hand to keep any PR out of the notes. The rest of `containers/` is deliberately absent: those images are built on the platform `v*` tag, so they ship *with* a release and belong in its notes.

The labeler matches with `all-globs-to-all-files`, so a PR spanning a separately released subtree *and* the platform is still announced. That's deliberate: sampling `v0.4.20..v0.4.26`, 26 of 31 merged PRs touched several top-level directories, and this label removes a change from the notes entirely.

Artifact releases keep asset-only or hand-written bodies. Turning on `generate_release_notes` for them would be actively worse — the range `understackctl/v0.0.4..understackctl/v0.0.5` spans 205 merge commits, almost none touching `understackctl`. `RELEASING.md` records why, with a path-scoped `git log` if a body is ever wanted.

## Why scriv

towncrier's markdown template prepends `- ` regardless of `all_bullets`, which un-indents continuation lines and double-bullets a fragment that already starts with a list — fatal for notes carrying numbered steps and code blocks. scriv preserves fragment content byte-for-byte, and names fragments by date/author/branch, so nothing needs an issue number that doesn't exist when the note is written.

Pinned `scriv>=1.8.0`, the version that introduced the behaviour relied on here. The constraint lives only in `requirements-docs.txt`: `release-notes.yaml` installs from that file rather than doing its own `pip install scriv`, so the workflow and `make docs` can't drift, and the `cache-dependency-path` already pointing at it becomes accurate.

## Labels

`release.yml` references four labels, and neither renovate nor `actions/labeler` creates a label it applies. All four now exist on the repo:

- `upgrade-impact` — gated on by `release-note-check.yaml`. This gap predates this PR.
- `skip-changelog` — applied by the labeler, or by hand.
- `dependencies` — applied by renovate, categorised in `release.yml`.
- `breaking change` — already existed.

## Before this merges

Note the bot's notes PR will show **no status checks**, because PRs created with the default `GITHUB_TOKEN` don't trigger `pull_request` workflows. `main` requires a review and no required checks, so it stays mergeable. Avoiding this needs a GitHub App token with a protection bypass, which seemed the wrong trade for a machine-generated docs-only change.

## Verified

scriv end-to-end (create → collect into both the Unreleased and series pages), rich content preserved, un-edited fragments collecting to nothing, the zero-fragment path (scriv exits 2, handled in both the Makefile and the workflow), markdownlint on generated output, yamllint on all six yaml files, and every pre-commit hook. `properdocs build --strict` with the new generated nav entry is left to CI here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


requirements-docs.txt
```diff
@@ -8,3 +8,4 @@ mkdocs-multirepo-plugin>=0.8.3
 mkdocs-include-dir-to-nav
 mkdocs-callouts
 mkdocs-mermaid-zoom
+scriv
```
nidzrai: Can we pin this as scriv>=1.8.0 ?  coz 1.7.0 does not have any config to it .
and makefile uses --config option , which when runs on 1.7.0  would silently fail without any error . 
```
`understack on  enroll_fw [$?] 🐍 openstack
❯ scriv --version
scriv collect --help
scriv, version 1.7.0
Usage: scriv collect [OPTIONS]

  Collect and combine fragments into the changelog.

Options:
  --add / --no-add     'git add' the updated changelog file and removed
                       fragments.
  --edit / --no-edit   Open the changelog file in your text editor.
  --title TEXT         The title text to use for this entry.
  --keep               Keep the fragment files that are collected.
  --version TEXT       The version name to use for this entry.
  -v, --verbosity LVL  Either CRITICAL, ERROR, WARNING, INFO or DEBUG
  --help               Show this message and exit.
(openstack)`
```

nidzrai: https://github.com/rackerlabs/understack/pull/2191/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R72

nidzrai (approved):

requirements-docs.txt
cardoe: Done

PR #2192, opened 2026-08-05
fix(understackctl): make version stamping actually work

cardoe:
The Makefile has always built LDFLAGS with -X main.version and
-X main.commit, but neither variable existed and nothing exposed a
--version flag. build-all also expanded $() instead of $(LDFLAGS).
Declare both variables, wire them into `understackctl --version`.

### DIFF
--- a/go/understackctl/Makefile
+++ b/go/understackctl/Makefile
@@ -43,7 +43,7 @@ build-all:
 			OUTFILE=$$DIR/$(BINARY_NAME)$$EXT; \
 			mkdir -p $$DIR; \
 			echo "Building $$OUTFILE..."; \
-			CGO_ENABLED=0 GOOS=$$GOOS GOARCH=$$GOARCH go build $() -o $$OUTFILE . || echo "Failed to build $$GOOS/$$GOARCH"; \
+			CGO_ENABLED=0 GOOS=$$GOOS GOARCH=$$GOARCH go build $(LDFLAGS) -o $$OUTFILE . || echo "Failed to build $$GOOS/$$GOARCH"; \
 		done \
 	done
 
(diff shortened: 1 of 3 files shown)


geetikabatra:
LGTM by looking at the code. I haven't tried the ctl manually myself on my laptop. Is there a definitive guide which I could follow to test this myself? Would I require a dev setup of understack in my local to test this?

cardoe:
> LDTM by looking at the code. I haven't tried the ctl manually myself on my laptop. Is there a definitive guide which I could follow to test this myself? Would I require a dev setup of understack in my local to test this?

https://rackerlabs.github.io/understack/operator-guide/understackctl/

geetikabatra:
I thought we decided to sepearate version change and the change in the MAKEFILE separate.

cardoe:
> I thought we decided to sepearate version change and the change in the MAKEFILE separate.

One doesn't work without the other. There's no way to test them. You need to have LDFLAGS passed along to get the version info passed in.

geetikabatra:
I get it now, I had a gap in my knowledge here. It is clear to me now. I faintly remember that LDFLAGS were used in other projects, but you have defined it earlier in [MAKEFILE](https://github.com/rackerlabs/understack/pull/2192/changes#diff-6037e15d91839f86dbe66d3f448de64ad7b7c690c080878485f85bcb26376aecL22) which has the keyword version in it. Hence the confusion. Since above line was earlier added at some other epoch of time. I could not  relate to it.
A quick suggestion, if code for setting version was part of go build command earlier, the LDFLAGS variable shouldn't have incorporated it. It should have been in this PR. This helps reduce redundancy and miscomprehension of code.

cardoe:
I'm not sure what you're looking for me to change code wise. Will you just make the change?

geetikabatra:
It looks good to me. No change is required.

cardoe:
> It looks good to me. No change is required.

Okay. Will you approve the PR then?

geetikabatra (approved):

geetikabatra:
Done!

PR #2197, opened 2026-08-06
fix(ironic): correct the chassis/model trait

cardoe:
Correct the documentation and the code to match each other.


mfencik (approved):
LGTM

PR #2205, opened 2026-08-10
feat(openstack-sync-operator): Operator for syncing openstack data as kubernetes custom resources

haseebsyed12:
openstack-sync-operator is a [**shell-operator-based reconciliation system**](https://github.com/flant/shell-operator) that syncs Kubernetes-defined OpenStack configuration into live OpenStack clouds. It uses a **declarative, Helm-driven plugin model** where new resource types are defined via:

1. A **CRD** that defines the Kubernetes schema
2. A **Python hook script** in the container image 
3. **Helm values entries** that wire everything together at deployment time

```
┌─────────────────────────────────────────────────────────────────┐
│ Kubernetes Layer (Operator Deployment)                          │
│  - shell-operator base pod                                      │
│  - Helm-wired hook discovery                                    │
│  - RBAC auto-generated from CRD metadata                        │
└──────────────────────┬──────────────────────────────────────────┘
                       │ (watches K8s resources)
┌──────────────────────▼──────────────────────────────────────────┐
│ Hook/Plugin Layer (Python Scripts in /hooks)                    │
│  - Each hook: a standalone Python script                        │
│  - Contract: exports HOOK_CONFIG, handles --config flag         │
│  - Lifecycle: onStartup, onChange, periodicCron, etc            │
└──────────────────────┬──────────────────────────────────────────┘
                       │ (makes API calls)
┌──────────────────────▼──────────────────────────────────────────┐
│ OpenStack API Layer                                             │
│  - Uses OpenStack SDK via Python environment                    │
│  - Authenticated via /etc/openstack/clouds.yaml                 │
│  - Reconciles: create, update, delete, prune operations         │
└─────────────────────────────────────────────────────────────────┘
```

Full CRUD implementation of neutron router flavors are implemented in [feat(openstack-sync-plugins): neutron router flavor - #2217](https://github.com/rackerlabs/understack/pull/2217)

### DIFF
--- a/charts/argocd-understack/templates/application-openstack-sync-operator.yaml
+++ b/charts/argocd-understack/templates/application-openstack-sync-operator.yaml
@@ -1,0 +1,47 @@
+{{- if eq (include "understack.isEnabled" (list $.Values.site "openstack_sync_operator")) "true" }}
+{{- $app := $.Values.site.openstack_sync_operator }}
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: {{ printf "%s-%s" $.Release.Name "openstack-sync-operator" }}
+  finalizers:
+  - resources-finalizer.argocd.argoproj.io
+  annotations:
+    argocd.argoproj.io/compare-options: ServerSideDiff=true,IncludeMutationWebhook=true
+    {{- with $app.wave }}
+    argocd.argoproj.io/sync-wave: {{ quote . }}
+    {{- end }}
+{{- include "understack.appLabelsBlock" $ | nindent 2 }}
+spec:
+  destination:
+    namespace: {{ $.Values.site.openstack.namespace }}
+    server: {{ $.Values.cluster_server }}
+  project: understack
+  sources:
+  - path: components/openstack-sync-operator
+    ref: understack
+    repoURL: {{ include "understack.understack_url" $ }}
+    targetRevision: {{ include "understack.understack_ref" $ }}
+    helm:
+      ignoreMissingValueFiles: false
+      releaseName: openstack-sync-operator
+      valueFiles:
+      - $understack/components/openstack-sync-operator/values.yaml
+      - $deploy/{{ include "understack.deploy_path" $ }}/openstack-sync-operator/values.yaml
+  - ref: deploy
+    repoURL: {{ include "understack.deploy_url" $ }}
+    targetRevision: {{ include "understack.deploy_ref" $ }}
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    managedNamespaceMetadata:
+      annotations:
+        argocd.argoproj.io/sync-options: Delete=false
+    syncOptions:
+    - CreateNamespace=true
+    - ServerSideApply=true
+    - RespectIgnoreDifferences=true
+    - ApplyOutOfSyncOnly=true
+{{- end }}
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -13,6 +13,7 @@ on:
       - "containers/ironic-nautobot-client/**"
       - "containers/ironic-vnc-container/**"
       - "containers/shell-operator-ironic/**"
+      - "containers/openstack-sync-operator/**"
       - "containers/understack-tests/**"
       - "python/**"
       - ".github/workflows/containers.yaml"
@@ -44,6 +45,8 @@ jobs:
             prebuild_script: ./sync_from_upstream.sh
             prebuild_script_working_dir: containers/ironic-vnc-container/
           - name: shell-operator-ironic
+            target: prod
+          - name: openstack-sync-operator
             target: prod
           - name: nautobot
             target: prod
--- a/charts/argocd-understack/templates/application-openstack-sync-plugins.yaml
+++ b/charts/argocd-understack/templates/application-openstack-sync-plugins.yaml
@@ -1,0 +1,38 @@
+{{- if eq (include "understack.isEnabled" (list $.Values.site "openstack_sync_plugins")) "true" }}
+{{- $app := $.Values.site.openstack_sync_plugins }}
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: {{ printf "%s-%s" $.Release.Name "openstack-sync-plugins" }}
+  finalizers:
+  - resources-finalizer.argocd.argoproj.io
+  annotations:
+    argocd.argoproj.io/compare-options: ServerSideDiff=true,IncludeMutationWebhook=true
+    {{- with $app.wave }}
+    argocd.argoproj.io/sync-wave: {{ quote . }}
+    {{- end }}
+{{- include "understack.appLabelsBlock" $ | nindent 2 }}
+spec:
+  destination:
+    namespace: {{ $.Values.site.openstack.namespace }}
+    server: {{ $.Values.cluster_server }}
+  project: understack
+  sources:
+  - path: components/openstack-sync-plugins
+    ref: understack
+    repoURL: {{ include "understack.understack_url" $ }}
+    targetRevision: {{ include "understack.understack_ref" $ }}
+  - path: {{ include "understack.deploy_path" $ }}/openstack-sync-plugins
+    ref: deploy
+    repoURL: {{ include "understack.deploy_url" $ }}
+    targetRevision: {{ include "understack.deploy_ref" $ }}
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    syncOptions:
+    - ServerSideApply=true
+    - RespectIgnoreDifferences=true
+    - ApplyOutOfSyncOnly=true
+{{- end }}
--- a/charts/argocd-understack/values.yaml
+++ b/charts/argocd-understack/values.yaml
@@ -310,6 +310,18 @@ site:
     # -- Enable/disable deploying OpenStack sync shell-operator
     # @default -- false
     enabled: false
+    # -- Sync wave for deployment ordering
+    # @default -- 4
+    wave: 4
+
+  # -- OpenStack sync custom resources
+  openstack_sync_plugins:
+    # -- Enable/disable deploying OpenStack sync custom resources
+    # @default -- false
+    enabled: false
+    # -- Sync wave for deployment ordering
+    # @default -- 5
+    wave: 5
 
   # -- RabbitMQ Cluster Operator (site deployment)
   rabbitmq_system:
--- a/charts/argocd-understack/values.yaml
+++ b/charts/argocd-understack/values.yaml
@@ -305,6 +305,12 @@ site:
     # @default -- false
     enabled: false
 
+  # -- OpenStack sync shell-operator
+  openstack_sync_operator:
+    # -- Enable/disable deploying OpenStack sync shell-operator
+    # @default -- false
+    enabled: false
+
   # -- RabbitMQ Cluster Operator (site deployment)
   rabbitmq_system:
     # -- Enable/disable deploying RabbitMQ operator
--- a/components/openstack-sync-operator/Chart.yaml
+++ b/components/openstack-sync-operator/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v2
+name: openstack-sync-operator
+description: OpenStack configuration sync shell-operator hooks
+type: application
+version: 0.2.0
+appVersion: 1.0.0
--- a/components/openstack-sync-operator/crds/neutron.understack.rackspace.net_neutronrouterflavors.yaml
+++ b/components/openstack-sync-operator/crds/neutron.understack.rackspace.net_neutronrouterflavors.yaml
@@ -1,0 +1,159 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: neutronrouterflavors.neutron.understack.rackspace.net
+spec:
+  group: neutron.understack.rackspace.net
+  names:
+    kind: NeutronRouterFlavor
+    listKind: NeutronRouterFlavorList
+    plural: neutronrouterflavors
+    shortNames:
+    - nrf
+    singular: neutronrouterflavor
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    additionalPrinterColumns:
+    - name: Flavor
+      type: string
+      jsonPath: .spec.name
+    - name: ServiceType
+      type: string
+      jsonPath: .spec.service_type
+    - name: Driver
+      type: string
+      jsonPath: .spec.driver
+    - name: SyncStatus
+      type: string
+      jsonPath: .status.syncStatus
+    - name: Age
+      type: date
+      jsonPath: .metadata.creationTimestamp
+    schema:
+      openAPIV3Schema:
+        description: NeutronRouterFlavor defines one Neutron router flavor and its service profile.
+        type: object
+        required:
+        - spec
+        properties:
+          apiVersion:
+            type: string
+          kind:
+            type: string
+          metadata:
+            type: object
+          spec:
+            type: object
+            required:
+            - name
+            - driver
+            properties:
+              name:
+                description: Neutron router flavor name.
+                type: string
+                minLength: 1
+                maxLength: 255
+                pattern: ^[A-Za-z0-9._-]+$
+              service_type:
+                description: Neutron service type for the flavor.
+                type: string
+                enum:
+                - L3_ROUTER_NAT
+                minLength: 1
+                maxLength: 255
+                default: L3_ROUTER_NAT
+              service_provider:
+                description: Optional Neutron service provider name used when generating Neutron configuration.
+                type: string
+                minLength: 1
+                maxLength: 255
+                pattern: ^[A-Za-z0-9._-]+$
+              description:
+                description: Description stored on the Neutron router flavor.
+                type: string
+                maxLength: 1024
+              driver:
+                description: Service profile driver class.
+                type: string
+                minLength: 1
+                maxLength: 1024
+                pattern: ^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)+$
+              profile_description:
+                description: Description stored on the Neutron service profile.
+                type: string
+                maxLength: 1024
+              profile_id:
+                description: Existing Neutron service profile ID to attach instead of creating or discovering one.
+                type: string
+                format: uuid
+              meta_info:
+                description: Service profile metainfo payload.
+                type: object
+                properties:
+                  resource_class:
+                    description: Resource class consumed by a physical router provider.
+                    type: string
+                    minLength: 1
+                    maxLength: 255
+                    pattern: ^[A-Za-z0-9._:-]+$
+                  vni_alloc:
+                    description: VNI allocation mode for VRF router providers.
+                    type: string
+                    enum:
+                    - "off"
+                    - "on"
+                    - auto
+          status:
+            description: NeutronRouterFlavorStatus defines the observed sync state.
+            type: object
+            properties:
+              syncStatus:
+                description: SyncStatus indicates the synchronization state with Neutron.
+                type: string
+                enum:
+                - Synced
+                - Failed
+                - Unknown
+              lastSyncTime:
+                description: LastSyncTime is the last time the operator attempted to sync the flavor.
+                type: string
+                format: date-time
+              observedGeneration:
+                description: ObservedGeneration is the metadata generation last processed by the operator.
+                type: integer
+                format: int64
+              message:
+                description: Message provides details about the last sync attempt.
+                type: string
+                maxLength: 2048
+              conditions:
+                description: Conditions describe current observed state.
+                type: array
+                items:
+                  type: object
+                  required:
+                  - type
+                  - status
+                  properties:
+                    type:
+                      type: string
+                    status:
+                      type: string
+                      enum:
+                      - "True"
+                      - "False"
+                      - Unknown
+                    reason:
+                      type: string
+                    message:
+                      type: string
+                      maxLength: 2048
+                    lastTransitionTime:
+                      type: string
+                      format: date-time
+    subresources:
+      status: {}
--- a/components/openstack-sync-operator/templates/_crd.tpl
+++ b/components/openstack-sync-operator/templates/_crd.tpl
@@ -1,0 +1,38 @@
+{{/*
+Read hook CRD metadata used by RBAC and shell-operator environment wiring.
+*/}}
+{{- define "openstack-sync-operator.hookCrd" -}}
+{{- $root := index . 0 -}}
+{{- $hookName := index . 1 -}}
+{{- $hook := index . 2 -}}
+{{- $crdPath := required (printf "hooks.%s.crd is required when hook is enabled" $hookName) $hook.crd -}}
+{{- $crdYaml := required (printf "hooks.%s.crd file %s is empty or missing" $hookName $crdPath) ($root.Files.Get $crdPath) -}}
+{{- $crd := fromYaml $crdYaml -}}
+{{- if ne $crd.kind "CustomResourceDefinition" -}}
+{{- fail (printf "hooks.%s.crd must point to a CustomResourceDefinition" $hookName) -}}
+{{- end -}}
+{{- $group := required (printf "hooks.%s.crd spec.group is required" $hookName) $crd.spec.group -}}
+{{- $kind := required (printf "hooks.%s.crd spec.names.kind is required" $hookName) $crd.spec.names.kind -}}
+{{- $plural := required (printf "hooks.%s.crd spec.names.plural is required" $hookName) $crd.spec.names.plural -}}
+{{- $storageVersion := "" -}}
+{{- $hasStatus := false -}}
+{{- range $version := required (printf "hooks.%s.crd spec.versions is required" $hookName) $crd.spec.versions }}
+{{- if $version.storage -}}
+{{- $storageVersion = $version.name -}}
+{{- end -}}
+{{- if hasKey (default dict $version.subresources) "status" -}}
+{{- $hasStatus = true -}}
+{{- end -}}
+{{- end -}}
+{{- if not $storageVersion -}}
+{{- fail (printf "hooks.%s.crd must define a storage version" $hookName) -}}
+{{- end -}}
+{{- dict
+  "apiVersion" (printf "%s/%s" $group $storageVersion)
+  "group" $group
+  "hasStatus" $hasStatus
+  "kind" $kind
+  "plural" $plural
+  "resource" (printf "%s.%s" $plural $group)
+  | toYaml -}}
+{{- end }}
--- a/components/openstack-sync-operator/templates/_helpers.tpl
+++ b/components/openstack-sync-operator/templates/_helpers.tpl
@@ -1,0 +1,90 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "openstack-sync-operator.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+*/}}
+{{- define "openstack-sync-operator.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "openstack-sync-operator.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels.
+*/}}
+{{- define "openstack-sync-operator.labels" -}}
+helm.sh/chart: {{ include "openstack-sync-operator.chart" . }}
+{{ include "openstack-sync-operator.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels.
+*/}}
+{{- define "openstack-sync-operator.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "openstack-sync-operator.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use.
+*/}}
+{{- define "openstack-sync-operator.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "openstack-sync-operator.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
+
+{{/*
+Checksum for hook settings that affect shell-operator startup configuration.
+*/}}
+{{- define "openstack-sync-operator.hookConfigChecksum" -}}
+{{- include "openstack-sync-operator.configuredHooks" . | fromYaml | toJson | sha256sum -}}
+{{- end }}
+
+{{/*
+Normalize built-in plugin hooks and direct hook definitions.
+*/}}
+{{- define "openstack-sync-operator.configuredHooks" -}}
+{{- $hooks := dict -}}
+{{- $enabledPlugins := default dict .Values.plugins -}}
+{{- range $pluginName, $plugin := default dict .Values.pluginData -}}
+{{- $hook := default dict $plugin.hook -}}
+{{- if gt (len $hook) 0 -}}
+{{- $hookValues := dict -}}
+{{- range $key, $value := $hook -}}
+{{- $_ := set $hookValues $key $value -}}
+{{- end -}}
+{{- $_ := set $hookValues "enabled" (eq (get $enabledPlugins $pluginName) true) -}}
+{{- $_ := set $hooks $pluginName $hookValues -}}
+{{- end -}}
+{{- end -}}
+{{- range $hookName, $hook := default dict .Values.hooks -}}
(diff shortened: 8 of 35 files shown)


charts/argocd-understack/templates/application-openstack-sync-operator.yaml
```diff
@@ -0,0 +1,50 @@
...
+  project: understack
+  sources:
+  - path: components/openstack-sync-operator
+    ref: understack
+    repoURL: {{ include "understack.understack_url" $ }}
+    targetRevision: {{ include "understack.understack_ref" $ }}
+    helm:
+      fileParameters:
+      - name: routerFlavors.config
+        path: {{ $routerFlavorsConfig }}
```
cardoe: Remove this. Those should just be regular CRDs and we'll load those a different way.

components/openstack-sync-plugins/data/router_flavors.json
```diff
@@ -0,0 +1,43 @@
+[
```
cardoe: This data should be separate from the deploy of the operator. And should be a CRD.

components/openstack-sync-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,88 @@
...
+        env:
+        - name: OS_CLOUD
+          value: {{ .Values.openstack.cloud | quote }}
+        {{- if .Values.routerFlavors.enabled }}
+        - name: NEUTRON_ROUTER_FLAVORS_CONFIG
+          value: {{ .Values.routerFlavors.mountPath | quote }}
+        - name: NEUTRON_ROUTER_FLAVOR_SYNC_CRONTAB
+          value: {{ .Values.routerFlavors.crontab | quote }}
+        - name: NEUTRON_ROUTER_FLAVOR_PRUNE
+          value: {{ .Values.routerFlavors.prune | quote }}
```
cardoe: This is too specific for this case. It just needs to generically watch CRDs.

components/openstack-sync-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - name: openstack-clouds
+          mountPath: /etc/openstack/clouds.yaml
+          subPath: clouds.yaml
+          readOnly: true
+        {{- if .Values.routerFlavors.enabled }}
+        - name: router-flavors-config
+          mountPath: {{ .Values.routerFlavors.mountPath | quote }}
+          subPath: {{ .Values.routerFlavors.configMapKey | quote }}
+          readOnly: true
+        {{- end }}
```
cardoe: Yeah remove this

cardoe (changes requested):

cardoe:
### Code review

Nice chunk of work here. My goal for the initial commit is a generic shell-operator container with the OpenStack CLI, where watched resources are enrolled through separate charts and adding a hook is easy to follow. Most of the scaffolding is right — the Application shape, the `site.<component>.enabled` toggles, chart placement, and the docs wiring all match existing conventions. The notes below are about the plugin layer on top of that, plus two destructive bugs in the prune path.

**Tracked separately, not this PR:** the `pyproject.toml` ruff `extend` deviation is pre-existing and repo-wide (#2207); migrating `shell-operator-ironic` onto this operator is a follow-on (#2208); multi-cloud support with per-resource cloud selection and system vs project scope is a follow-on (#2210).

---

## Framework shape

**1. Building hooks into the image is right — the docs just contradict it**

Baking hooks in is what we want: hooks and their Python dependencies version together, CI tests the artifact that ships, and the image tag fully determines behavior. It also matches `containers/shell-operator-ironic`. A new container version per new resource type is expected and fine.

The problem is that the component doc claims the opposite:

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/docs/deploy-guide/components/openstack-sync-operator.md#L112-L115

while the README correctly documents adding a `COPY` line per plugin:

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/containers/openstack-sync-operator/README.md#L60-L67

Please make the component doc match the Dockerfile and state the versioning contract explicitly: adding a watched resource requires a new image, and the operator chart is bumped to match.

**2. No OpenStack CLI in the image**

The only dependency is `openstacksdk`, so there's no `openstack` binary. Compare `containers/shell-operator-ironic/requirements.txt`, which installs `python-openstackclient` and `python-ironicclient`.

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/containers/openstack-sync-operator/Dockerfile#L17-L21

**3. CRDs belong with the operator; only the CRs live separately**

A CRD is the API contract of the hook that reads it, so schema and code must version together. Right now the CRD is in the plugin chart at `components/openstack-sync-neutron-router-flavors/crds/`, which lets a site load a schema newer or older than the code consuming it. Move it to `components/openstack-sync-operator/crds/` using Helm's `crds/` convention, as [`go/dexop/helm/crds/clients.yaml`](https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/go/dexop/helm/crds/clients.yaml) and [`go/nautobotop/helm/crds/clients.yaml`](https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/go/nautobotop/helm/crds/clients.yaml) do, and as [`components/ironic/kustomization.yaml`](https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/components/ironic/kustomization.yaml#L13-L16) does by pulling `runbook-crd` and `runbook-operator` together.

The custom resources then live in their own chart with their own Application, following [`charts/nautobot-job-queues`](https://github.com/rackerlabs/understack/tree/582e2341088b2f332f9a74f956f73e185ad8a13e/charts/nautobot-job-queues) + [`application-nautobot-job-queues.yaml`](https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/charts/argocd-understack/templates/application-nautobot-job-queues.yaml). Note that pattern also adds a `$deploy/<site>/<name>` source so a site can add resources without changing understack.

Side benefit: with CRDs in the operator chart, `helm template` emits them and ArgoCD applies them with the operator, so the sync-wave `-1` plus `SkipDryRunOnMissingResource=true` juggling inside one Application is no longer needed.

**4. Drop the convention-based plugin discovery**

Nothing else in `charts/argocd-understack/templates/` discovers components by name prefix; every other Application is an explicit template.

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/charts/argocd-understack/templates/_helpers.tpl#L236-L257

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/charts/argocd-understack/templates/application-openstack-sync-operator.yaml#L31-L40

**5. Hook enablement and crontabs belong in the operator chart's values**

`HOOK_CONFIG` is built at module import, and shell-operator only invokes `--config` at startup:

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/hook.py#L58-L67

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/router_flavors_common.py#L27-L30

Those values arrive via `envFrom: configMapRef`, which Kubernetes does **not** propagate to a running container, and there's no checksum annotation to force a rollout — nor can there be, since the ConfigMap is rendered by a different Helm release:

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/components/openstack-sync-operator/templates/deployment.yaml#L40-L44

Failure case: set `routerFlavors.crontab: "*/15 * * * *"`, ArgoCD syncs green, the pod keeps syncing hourly, nothing reports a problem. Same for `prune` and `enabled`.

Since the image knows its own hooks, put "which hooks are enabled and on what crontab" in the operator chart's own values rather than a ConfigMap authored by a separate chart. That removes the cross-release coupling, makes a `checksum/config` pod annotation possible so a schedule change actually rolls out, and gives one uniform way to register a schedule instead of a per-plugin `<SERVICE>_<TARGET>_SYNC_CRONTAB` namespace that the operator has to know about.

**6. Fail loudly when a hook is enabled but not present in `/hooks`**

Prod pins the image to a hash via the deployment repo's values, so the pinned image and `understack_ref` move independently by design. A site pins an image, someone bumps `understack_ref` to pick up a newly added resource chart, the corresponding hook isn't in the pinned image — ArgoCD is green, the CRs are applied, nothing reconciles them, no error surfaces. The operator should verify at startup that every hook it's told to enable actually exists.

---

## Correctness

**7. Prune deletes router flavors the operator never created**

`is_prunable_flavor()` treats a flavor as prunable if its own description carries the marker **or** if any attached service profile is operator-managed:

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/delete_router_flavors.py#L53-L70

The second branch infers ownership from a *shared* object. The PR's own test documents it — `fl-remove` has no description marker at all, yet is asserted deleted purely because it shares `sp-shared` with a desired flavor:

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/python/openstack-sync/tests/test_router_flavors.py#L510-L540

Failure case: someone hand-creates a router flavor against the shared `neutron_understack.l3_router.vrf.Vrf` service profile, and it isn't attached to a live router yet. The next hourly sync deletes it. The description marker is the only real ownership signal; the profile branch should go.

**8. Disabling the plugin deletes every managed flavor**

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/components/openstack-sync-neutron-router-flavors/templates/hook-env-configmap.yaml#L10-L12

`NEUTRON_ROUTER_FLAVOR_ENABLED` is a literal `"true"` gated only on `.Values.hook.enabled` (default true), not on `.Values.routerFlavors.enabled`. Set `routerFlavors.enabled: false` and `templates/router-flavors.yaml` stops rendering CRs while the hook stays enabled. The binding context yields an empty snapshot, so `run()` calls `prune_removed_flavors(conn, [])` with an empty desired set — and with `prune: true` by default, every operator-managed `L3_ROUTER_NAT` flavor in Neutron is deleted. Combined with finding 7 above that reaches flavors the operator never created. "Disable this plugin" must not be destructive.

(`templates/rbac.yaml` is also gated on `routerFlavors.enabled`, so the same config drops RBAC out from under an enabled hook.)

**9. The operator chart ships no RBAC for its own ServiceAccount**

`components/openstack-sync-operator/templates/` contains only `serviceaccount.yaml` and `deployment.yaml`; every permission comes from whichever plugin chart happens to be enabled. Since `site.openstack_sync_operator.enabled` toggles independently, enabling the operator with no plugins yields a Pod whose SA can do nothing, with ArgoCD reporting Synced/Healthy and the only symptom in container logs. Compare `components/ironic/runbook-operator/role.yaml`, which ships its ClusterRole alongside the Deployment.

**10. A single bad flavor aborts the whole reconcile**

In `hook.py`, `run()` re-raises on the first `sync_flavor` failure, so every later flavor is left unsynced with no status written.

Minor: no `livenessProbe`/`readinessProbe`, though shell-operator serves `/healthz`. `wait_for_openstack_network()` retries 30×10s before raising, so a hung hook sits Running with no kubelet restart.

---

## Housekeeping

**11. Chart templates should be `.yaml.tpl`**

Every existing Helm chart under `components/` uses that extension ([`etcdbackup`](https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/components/etcdbackup/templates/job-etcd-backup.yaml.tpl), `components/openstack/templates/`, `envoy-configs`, `understack-cdn`). Helm renders them the same, and the extension is why they don't trip the YAML linters. Using plain `.yaml` here required broadening a global exclude, which disables `check-yaml` for all of `components/*/templates/` repo-wide and for any file added there in future:

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/.pre-commit-config.yaml#L14-L16

Renaming lets both this and the `.yamllint.yaml` change be reverted.

**12. Drop the `pr-2205` image tag**

https://github.com/rackerlabs/understack/blob/582e2341088b2f332f9a74f956f73e185ad8a13e/components/openstack-sync-operator/values.yaml#L3-L6

Leave `tag` unset so it falls back to `appVersion`, with the deployment repo pinning the hash — the convention in `charts/undersync/values.yaml`. No other `values.yaml` under `components/` or `charts/` pins a `pr-` tag.

---

## Suggested split

Land the framework alone — container, generic chart with the CRDs, plain Application — and move the Neutron router flavors work (CRs, `router_flavors.json`, hook, Python package) to a follow-up as its first consumer. That proves the "add a hook without touching the operator chart" property instead of asserting it, and keeps findings 7 and 8 off the critical path for the framework commit.

The reconciliation logic is otherwise in reasonable shape, and the service-profile marker scheme in `router_flavors_common.py` is a sound idea — it just needs to stop being used to infer *flavor* ownership.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cardoe:
### Code review — pass 2 (`536301ad`, "openstack-sync-plugins iter-1")

Re-reviewed after your push. The chart layer got a solid pass. `python/openstack-sync/` and `containers/openstack-sync-operator/Dockerfile` are byte-identical to the previous revision, so everything I raised about the hooks, the image, and the prune logic is still open — spelled out in full below so you don't have to cross-reference.

## Resolved

**Convention-based plugin discovery is gone.** `understack.openstackSyncPlugins` deleted from `_helpers.tpl`, and the Application now keys off an explicit `site.openstack_sync_plugins` toggle. Collapsing the per-service charts into one `components/openstack-sync-plugins` chart works for me.

**Disabling a plugin is no longer destructive.** `<PREFIX>_ENABLED` now derives from `$plugin.enabled` rather than being a hardcoded `"true"`, and `hook-env-configmap.yaml`, `rbac.yaml`, and `resources.yaml` all gate on that same flag, so the RBAC-vs-hook skew went away with it.

**`templates/_crd.tpl` is more than I asked for, and good.** Deriving `apiVersion`, `kind`, `resource`, `group`, `plural`, and `hasStatus` from the CRD file means the env vars and RBAC rules can't drift from the schema. The `values.schema.json` is a welcome addition too.

## New in this push

**`config: []` silently falls back to the bundled data, and enabled-with-no-config still prunes everything**

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/components/openstack-sync-plugins/templates/resources.yaml#L6-L13

`{{- if not $config }}` treats an empty list as falsy, so `plugins.neutronRouterFlavors.config: []` doesn't mean "no resources" — it silently reverts to the chart's bundled `data/router_flavors.json`. There is no way to express an empty set.

Worse, a plugin enabled with neither `config` nor `configPath` renders zero CRs while `<PREFIX>_ENABLED` stays `"true"`. The hook then runs against an empty snapshot and calls `prune_removed_flavors(conn, [])` with an empty desired set, deleting every operator-managed `L3_ROUTER_NAT` flavor. Same destination as the bug this push fixed, reached by a different route. The durable fix belongs in Python: treat an empty desired set as "nothing to do" rather than "delete everything", and require an explicit opt-in for a full purge.

**Stale yamllint entry**

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/.yamllint.yaml#L6-L8

`components/openstack-sync-neutron-router-flavors/templates/` no longer exists.

**Squash before merge**

The branch is two commits now, the second being `openstack-sync-plugins iter-1`. Please squash to a single standalone commit — the history shouldn't narrate the design iterations.

## Still open

**Prune infers flavor ownership from a shared service profile.** This is the one I would gate a real cloud on.

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/delete_router_flavors.py#L46-L71

`is_prunable_flavor()` treats a flavor as prunable if its own description carries the marker **or** if any attached service profile is operator-managed. The second branch infers ownership from a shared object. The test at [`test_router_flavors.py#L510-L540`](https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/python/openstack-sync/tests/test_router_flavors.py#L510-L540) documents it: `fl-remove` has no description marker at all, yet is asserted deleted purely because it shares `sp-shared` with a desired flavor. So a hand-created flavor using the shared `neutron_understack.l3_router.vrf.Vrf` profile, not yet attached to a live router, gets deleted on the next hourly sync. The description marker is the only real ownership signal; the profile branch should go.

**The docs still contradict the Dockerfile.**

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/docs/deploy-guide/components/openstack-sync-operator.md#L116-L120

This says no operator image changes are needed to add a plugin, while [`README.md#L60-L67`](https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/containers/openstack-sync-operator/README.md#L60-L67) correctly documents adding a `COPY` line. Building hooks into the image is the right call — hooks and their Python dependencies version together, CI tests the artifact that ships, and it matches `containers/shell-operator-ironic`. So fix the doc, and state the contract: adding a watched resource requires a new image, and the operator chart is bumped to match.

**No OpenStack CLI in the image.**

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/containers/openstack-sync-operator/Dockerfile#L17-L21

Only `openstacksdk` is installed, so there is no `openstack` binary. Compare `containers/shell-operator-ironic/requirements.txt`, which installs `python-openstackclient` and `python-ironicclient`.

**CRDs belong with the operator; only the CRs belong in the plugins chart.** The CRD is the API contract of the hook that reads it, so schema and code must version together. It currently sits in `components/openstack-sync-plugins/crds/`, which lets a site load a schema newer or older than the code consuming it. Move it to `components/openstack-sync-operator/crds/`, as [`go/dexop/helm/crds/clients.yaml`](https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/go/dexop/helm/crds/clients.yaml) and [`go/nautobotop/helm/crds/clients.yaml`](https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/go/nautobotop/helm/crds/clients.yaml) do, and as [`components/ironic/kustomization.yaml`](https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/components/ironic/kustomization.yaml#L13-L16) does by pulling `runbook-crd` and `runbook-operator` in together. That also drops the need for the sync-wave `-1` plus `SkipDryRunOnMissingResource=true` juggling.

**The plugins chart should be its own Application, not an extra source on the operator's.** Follow [`charts/nautobot-job-queues`](https://github.com/rackerlabs/understack/tree/536301ad47badeee49dba5b008544079946a4e4b/charts/nautobot-job-queues) plus [`application-nautobot-job-queues.yaml`](https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/charts/argocd-understack/templates/application-nautobot-job-queues.yaml). That pattern also adds a `$deploy/<site>/<name>` source so a site can add resources without changing understack.

**Changing a crontab never reaches a running pod.**

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/hook.py#L58-L67

`HOOK_CONFIG` is built at module import and shell-operator only invokes `--config` at startup. The values arrive via `extraEnvFrom` referencing a ConfigMap rendered by a *different* Helm release, and Kubernetes does not propagate ConfigMap changes into a running container — so there is no checksum annotation that could force a rollout. Set `SYNC_CRONTAB: "*/15 * * * *"`, ArgoCD syncs green, the pod keeps syncing hourly, nothing reports a problem. Since the image knows its own hooks, put "which hooks are enabled and on what crontab" in the operator chart's own values so a `checksum/config` pod annotation works and there is one uniform way to register a schedule.

**A hook that is enabled but absent from `/hooks` should fail loudly.** Prod pins the image via the deployment repo's values, so the pinned image and `understack_ref` move independently by design. Bump `understack_ref` to pick up a new plugin against an older pinned image and the hook simply isn't there: ArgoCD green, CRs applied, nothing reconciles them, no error.

**The operator chart ships no RBAC for its own ServiceAccount.** `components/openstack-sync-operator/templates/` contains only `_helpers.tpl`, `deployment.yaml`, and `serviceaccount.yaml`. Every permission comes from the plugins chart. Since `site.openstack_sync_operator.enabled` toggles independently, enabling the operator alone yields a Pod whose SA can do nothing, with ArgoCD reporting Synced/Healthy and the only symptom in container logs. Compare `components/ironic/runbook-operator/role.yaml`, which ships its ClusterRole next to the Deployment.

**One bad flavor aborts the whole reconcile.** `run()` re-raises on the first `sync_flavor` failure, so every later flavor is left unsynced with no status written. Also no `livenessProbe`/`readinessProbe`, though shell-operator serves `/healthz`, and `wait_for_openstack_network()` retries 30×10s before raising — so a hung hook sits Running with no kubelet restart.

**Chart templates should be `.yaml.tpl`.** Every existing Helm chart under `components/` uses that extension ([`etcdbackup`](https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/components/etcdbackup/templates/job-etcd-backup.yaml.tpl), `components/openstack/templates/`, `envoy-configs`, `understack-cdn`). Helm renders them identically, and the extension is why they don't trip the YAML linters. Using plain `.yaml` required broadening a global exclude, which disables `check-yaml` for all of `components/*/templates/` repo-wide and for anything added there in future:

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/.pre-commit-config.yaml#L14-L16

Renaming lets both that and the `.yamllint.yaml` additions be reverted.

**Drop the `pr-2205` image tag.**

https://github.com/rackerlabs/understack/blob/536301ad47badeee49dba5b008544079946a4e4b/components/openstack-sync-operator/values.yaml#L3-L6

Leave `tag` unset so it falls back to `appVersion`, with the deployment repo pinning the hash — the convention in `charts/undersync/values.yaml`.

---

Still worth splitting the Neutron router flavors work into a follow-up so the framework lands on its own. That moves the prune-ownership bug, the abort-on-first-failure, and the empty-config purge off this PR's critical path.

Tracked separately, not this PR: the `pyproject.toml` ruff `extend` deviation (#2207), migrating `shell-operator-ironic` onto this operator (#2208), and multi-cloud support with per-resource cloud selection and system vs project scope (#2210).

🤖 Generated with [Claude Code](https://claude.ai/code)

cardoe:
### Code review — pass 3 (`893a4c30`)

Big improvement. The framework now looks the way I was asking for: CRDs ship with the operator and are read back through `_crd.tpl` so RBAC and hook env can't drift from the schema, hook enablement and crontabs live in the operator chart's own values behind a `checksum/` annotation that actually rolls the pod, a `verify-hooks` initContainer fails loudly on a missing hook, RBAC ships with the Deployment, probes are wired to `9115`, templates are back to `.yaml.tpl` so the `.pre-commit-config.yaml`/`.yamllint.yaml` broadening is reverted, the plugins CRs are their own Application with a `$deploy/<site>/` source, and the `pr-2205` tag is gone. The prune-ownership and abort-on-first-failure bugs went away with the split.

Found 4 issues:

1. `extraEnv` and `extraEnvFrom` are declared in values but never rendered, so setting them silently does nothing

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/components/openstack-sync-operator/values.yaml#L24-L27

`git grep extraEnv components/openstack-sync-operator/` matches only those two lines — no template reads them. The container's `env:` is built solely from `OS_CLOUD`, `POD_NAMESPACE`, and the derived hook vars, and there is no `envFrom:` at all:

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/components/openstack-sync-operator/templates/deployment.yaml.tpl#L96-L108

Every other pod-level value in this chart is consumed (`podAnnotations`, `podLabels`, `resources`, `nodeSelector`, `tolerations`, `affinity`, `rbac.rules`), so these two are the outliers. `values.schema.json` is `additionalProperties: true` at the top level, so it won't flag them either. Either render them or drop them — a site that sets `extraEnv` to inject a proxy or an extra credential gets no error and no effect.

2. The image needs `openstacksdk`, `python-openstackclient` and `python-ironicclient`

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/python/openstack-sync/pyproject.toml#L4-L13

`[project]` has no `dependencies` key at all. The previous revision at least had `openstacksdk`, so this went backwards. The Dockerfile installs `python3` plus the package itself and nothing else:

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/containers/openstack-sync-operator/Dockerfile#L12-L19

Meanwhile the chart already mounts `clouds.yaml` and sets `OS_CLOUD`, which only means something if a client is present:

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/components/openstack-sync-operator/templates/deployment.yaml.tpl#L97-L112

Please add all three — `openstacksdk` for hooks that talk to the APIs directly, and `python-openstackclient` plus `python-ironicclient` so a hook can shell out to `openstack` / `baremetal`. That matches [`containers/shell-operator-ironic/requirements.txt`](https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/containers/shell-operator-ironic/requirements.txt), and it is what #2208 will need when the Ironic runbook operator moves onto this image. The point of this container is to be the shared plugin host, so its dependency set should be stable across hooks rather than growing with each one.

3. `neutronRouterFlavors` is wired up but `router_flavors.py` doesn't exist

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/components/openstack-sync-operator/values.yaml#L32-L41

`pluginData.neutronRouterFlavors.hook.path` points at `/hooks/router_flavors.py`, which is nowhere in the branch — the Dockerfile only copies `placeholder.py`:

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/containers/openstack-sync-operator/Dockerfile#L18-L19

So the commit titled "Adding first consumer Neutron router flavors to sync operator" adds the CRD, four CRs and the values scaffolding but no consumer, and flipping the one boolean the docs call the enablement step puts the Deployment into `Init:Error` via `verify-hooks`. Both toggles default to `false` and the docs do warn to enable it "only after the image contains the hook", so nothing breaks as shipped — but a plugin entry that can only ever fail is a trap for the next person. Either drop the `neutronRouterFlavors` values and CRs so the framework lands genuinely on its own, or file a tracking issue for the hook the way #2208 and #2210 are tracked and reference it from the commit message.

4. The two new component docs skip the doc-macro convention

https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/docs/deploy-guide/components/openstack-sync-operator.md#L1-L4

All 54 existing pages under `docs/deploy-guide/components/` open with frontmatter (`charts:`/`kustomize_paths:` plus `deploy_overrides:`) and call `{{ component_argocd_builds() }}` and `{{ secrets_disclaimer }}` from [`docs/macros.py`](https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/docs/macros.py). Both new pages have no frontmatter and hand-write that prose instead — see the "How ArgoCD Builds It" section listing the value files literally. Commit `f4bd7159` converted all 52 pages at the time specifically to stop hand-copying this, and `scripts/check-component-docs.py` only checks that the file exists, so CI won't catch the regression. Compare [`undersync.md`](https://github.com/rackerlabs/understack/blob/893a4c30002f76151bb178a3efef9138487a4a59/docs/deploy-guide/components/undersync.md#L1-L9).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cardoe:
### Code review — pass 4 (`0b0ef7aa`)

Pass 3's four issues are all addressed: `extraEnv`/`extraEnvFrom` are gone, `openstacksdk` + `python-openstackclient` + `python-ironicclient` are declared in `pyproject.toml`, `router_flavors.py` exists and is `COPY`'d into the image, and both docs pages now carry the frontmatter and `{{ component_argocd_builds() }}`.

Found 2 issues:

1. The operator doc states reconciliation behavior that `router_flavors.py` does not implement. `main()` only handles `--config`; on a CR event or a scheduled run shell-operator invokes the hook with no arguments and it returns 0 without reading `$BINDING_CONTEXT_PATH`, importing `openstacksdk`, or patching status — nothing in `python/openstack-sync/` does. The commit subject says "(interface only)" and the PR body points at #2217, but neither reaches the published page, which asserts the behavior as present fact and only caveats the *disabled* state. Follow the enablement step at [L113-L119](https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/docs/deploy-guide/components/openstack-sync-operator.md#L113-L119) and you get a pod that passes `verify-hooks`, generated RBAC, green probes, applied CRs, and no Neutron changes, with no error logged.

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/docs/deploy-guide/components/openstack-sync-operator.md#L161-L166

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/python/openstack-sync/openstack_sync/hooks/router_flavors.py#L70-L75

2. A `plugins` key with no matching `pluginData` entry is silently ignored. `configuredHooks` ranges over `.Values.pluginData` and looks each name up in `.Values.plugins`, never the reverse, and both `deployment.yaml.tpl` and `rbac.yaml.tpl` consume only its output. So a misspelled or renamed key in `$deploy/<site>/openstack-sync-operator/values.yaml` — e.g. `plugins.neutronRouterFlavor: true` — yields no env var, no RBAC rule, no `verify-hooks` entry and no error; `helm template` succeeds and ArgoCD reports Synced/Healthy. `values.schema.json` accepts any boolean-valued key, so it can't catch it either. This is the residual case of the gap `verify-hooks` was added to close, and `deployment.yaml.tpl` already uses `fail` for the duplicate-env-var case, so the same treatment fits.

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/components/openstack-sync-operator/templates/_helpers.tpl#L76-L85

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/components/openstack-sync-operator/values.schema.json#L12-L19

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cardoe:
### Pass 4 disposition — merging, follow-ups to come

Four items to carry into a follow-on. None block the merge: the framework shape is what I was after, and both `site.openstack_sync_operator.enabled` and `site.openstack_sync_plugins.enabled` default to `false`, so none of this is live on merge.

**1. The operator doc describes reconciliation that `router_flavors.py` doesn't implement.** `main()` only handles `--config`; on a CR event or a scheduled run shell-operator invokes the hook with no arguments and it returns 0 without reading `$BINDING_CONTEXT_PATH`, importing `openstacksdk`, or patching status. The commit subject says "(interface only)" and the PR body points at #2217, but neither reaches the published page, which states the behavior as present fact and only caveats the *disabled* state. Following the enablement step at [L113-L119](https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/docs/deploy-guide/components/openstack-sync-operator.md#L113-L119) gives a pod that passes `verify-hooks`, generated RBAC, green probes, applied CRs, and no Neutron changes, with no error logged. Either tone the doc down to match "interface only" or land it alongside #2217.

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/docs/deploy-guide/components/openstack-sync-operator.md#L161-L166

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/python/openstack-sync/openstack_sync/hooks/router_flavors.py#L70-L75

**2. A `plugins` key with no matching `pluginData` entry is silently ignored.** `configuredHooks` ranges over `.Values.pluginData` and looks each name up in `.Values.plugins`, never the reverse, and both `deployment.yaml.tpl` and `rbac.yaml.tpl` consume only its output. A misspelled or renamed key in `$deploy/<site>/openstack-sync-operator/values.yaml` yields no env var, no RBAC rule, no `verify-hooks` entry and no error — `helm template` succeeds and ArgoCD reports Synced/Healthy. `values.schema.json` takes any boolean-valued key, so it won't catch it either. This is the residual case of the gap `verify-hooks` was added to close, and `deployment.yaml.tpl` already uses `fail` for the duplicate-env-var case, so the same treatment fits.

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/components/openstack-sync-operator/templates/_helpers.tpl#L76-L85

**3. Both sources of `application-openstack-sync-plugins.yaml` carry a `ref:` that is never dereferenced.** There's no `helm.valueFiles` and no `$understack/`/`$deploy/` usage anywhere in the Application, so neither ref does anything. To be clear this isn't introduced here — `application-global-workflows.yaml` and `application-argo-events-workflows.yaml` have the identical two-source pattern and survived the cleanups in 92ee95eb and 8f9e2d73, which only touched single-source apps. Since those refs are what tripped argoproj/argo-cd#25460, I'd rather clean up all three together than let a fourth copy accumulate.

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/charts/argocd-understack/templates/application-openstack-sync-plugins.yaml#L22-L28

**4. `rbac.yaml.tpl` calls `hookCrd` unguarded, so an enabled hook without a CRD can't render.** `_crd.tpl` does `required "...crd is required when hook is enabled"`, so `helm template --set hooks.myhook.enabled=true --set hooks.myhook.path=/hooks/x.py --set hooks.myhook.envPrefix=MYHOOK` fails outright at `rbac.yaml.tpl:10`. The chart contradicts itself here: `deployment.yaml.tpl` guards the same call with `if $envPrefix` / `if or $hookEnabled $crdPath`, and `values.schema.json` declares `crd` optional. It fails loudly rather than silently and every hook we have planned has a CRD, so it's low priority — but a cron-only hook is a shape the framework should support.

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/components/openstack-sync-operator/templates/rbac.yaml.tpl#L8-L12

https://github.com/rackerlabs/understack/blob/0b0ef7aa9f9efdcceea3c07b9b24b05b5a0d939f/components/openstack-sync-operator/templates/_crd.tpl#L7-L10

Merging as-is and opening a follow-on for these.

cardoe (approved):

PR #2206, opened 2026-08-11
fix(workflows): quote string literals in argo when expressions

nicholaskuechler:
Argo Workflows uses the `expr` expression engine to evaluate `when` conditions. Bare (unquoted) string values containing hyphens like `wanted-but-not-project` are parsed as arithmetic subtraction (`wanted - but - not - project`), which fails because these identifiers aren't numbers. The fix wraps comparison values in single quotes so they're treated as string literals.


khackworth (approved):
lgtm

PR #2211, opened 2026-08-12
feat(images): Ubuntu image build updates and fixes, and adds ubuntu 26.04 builder

nicholaskuechler:
* Updates the disk image builder packages to latest versions
* Adds missing shim for Ubuntu 24.04 so that it now builds and works correctly. The missing shim was causing it to fail to boot after IPA deployed the image to the node.
* Adds image builder config to build Ubuntu 26.04

### DIFF
--- a/.github/workflows/build-ironic-images.yaml
+++ b/.github/workflows/build-ironic-images.yaml
@@ -112,7 +112,7 @@ jobs:
           mkdir -p upload
           find . -maxdepth 1 -type f \( -name '*.qcow2' -o -name '*.kernel' -o -name '*.initramfs' \) -exec cp {} upload/ \;
         env:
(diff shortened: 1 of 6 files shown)


cardoe:
Just wondering if any of these fixes exist upstream or have been submitted upstream?

nicholaskuechler:
> Just wondering if any of these fixes exist upstream or have been submitted upstream?

Well seeing as we're using the latest diskimage-builder, I think it's safe to say they don't exist upstream. I don't even know whether they should.

I was unable to trace the original changes which broke our ubuntu 24.04 build... 2 years ago it worked fine, but sometime over the past 2 years - with thousands of Understack and OpenStack and diskimage-builder changes over that time period, it broke.

The ubuntu 26.04 does exist in diskimage-builder but it seems like other folks are using it fine, without any changes. Or perhaps no one is using at all so no one knows it doesn't work? :) The closest issue I could find is https://bugs.launchpad.net/ironic/+bug/2150502 which seems related.

cardoe (approved):

PR #2217, opened 2026-08-14
feat(openstack-sync): full CRUD plugin for Neutron router flavors

haseebsyed12:
## Layout

```
openstack_sync/
  utils.py                      Kubernetes Secret access + memoised connections
  hooks/
    common.py                   binding-context I/O, CR status patching
    framework.py                HookConfig, SyncPlugin, run_sync(), run_hook()
    placeholder.py              connectivity probe (no CRs)
    router_flavors.py           NeutronRouterFlavor hook
  plugins/
    common.py                   OpenStack helpers shared by all plugins
    neutron/router_flavors/
      config.py                 plugin constants
      markers.py                ownership markers
      reconcile.py              converge one CR
      prune.py                  delete resources whose CR was removed
```


cardoe:
### Code review

Found 6 issues. The theme: a lot of machinery is built and wired into the chart, but never called.

**1. Four functions have no non-test callers, so three chart env knobs are inert.**
`git grep` finds only definitions for `patch_flavor_status`, `prune_removed_flavors`, `load_router_flavor_resources`, and `wait_for_openstack_network`. `reconcile_router_flavor` calls `sync_flavor` and returns. Consequences: `STATUS_ENABLED` does nothing and `.status.syncStatus` / the `SyncStatus` printer column stay empty on success *and* failure; `NEUTRON_ROUTER_FLAVOR_PRUNE` does nothing and all 238 lines of `delete.py` are unreachable, so deleting a CR orphans the Neutron flavor and profile forever (the "D" of the commit title); `READY_RETRIES`/`READY_DELAY` do nothing, and `wait_for_openstack_network` is duplicated in `plugins/common.py:276`.

https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/hooks/router_flavors.py#L218-L224

**2. Disabling the plugin no longer disables the hook.**
The gate moved from `NEUTRON_ROUTER_FLAVOR_ENABLED` to "is `SYNC_CRONTAB` non-empty", but `configuredHooks` injects env for every *configured* hook regardless of `enabled`, and `SYNC_CRONTAB` defaults to `"0 * * * *"`. `helm template` with stock values yields `ENABLED="false"` alongside `SYNC_CRONTAB="0 * * * *"`, so with `plugins.neutronRouterFlavors: false` the hook still registers its watch and schedule — while `rbac.yaml.tpl` gates the `neutronrouterflavors` get/list/watch rules on `$hook.enabled`, so that watch 403s. This undoes the `<PREFIX>_ENABLED`-derives-from-`$plugin.enabled` fix from #2205. `placeholder.py` still uses `OPENSTACK_PLACEHOLDER_ENABLED`.

https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/hooks/router_flavors.py#L78-L87

**3. Abort-on-first-failure is back.**
`return 1` inside the loop abandons every remaining object in a Synchronization/Schedule batch, so one malformed CR leaves the rest unsynced — with no status written, per issue 1. This was flagged on #2205 and went away with the split. The new test only covers a single-bad-item batch.

https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/hooks/common.py#L218-L222

**4. The hourly sync never dispatches.**
The filter compares `context["binding"]` against `"neutron-router-flavors"`, but the schedule is registered as `"name": "hourly sync"` and shell-operator sets `binding` to the schedule's own name, not the k8s binding referenced via `includeSnapshotsFrom`. The `else:` branch added here for Schedule snapshots is unreachable; only Add/Modify events reconcile. Worth confirming against a real pod's `BINDING_CONTEXT_PATH`. Note `snapshot_items()` scans all contexts for the snapshot key, which is the correct pattern — but it's only reachable from the dead `load_router_flavor_resources`.

https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/hooks/router_flavors.py#L102-L108

**5. Replacement service profiles are created without ownership markers.**
`meta_info if configured_profile_id else managed_meta_info(meta_info)` — when `profile_id` is set but that profile is gone, the newly created one gets no markers, so `is_managed_service_profile` never recognizes it and prune can never reclaim it. Relatedly, a stale `profile_id` silently degrades to discover-or-create, contradicting the CRD's "to attach instead of creating or discovering one".

https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/create.py#L106-L118

**6. `comparable_meta_info` is a no-op that claims otherwise.**
Docstring says it strips operator-managed keys; the body is `{k: v for k, v in normalized.items()}` with no filter. Unused today (callers import the real one from `router_flavors_common`), so it's a trap for the next plugin author rather than a live bug.

https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/plugins/common.py#L160-L172

Checked and clear: the empty-desired-set prune guard from #2205 is present; `DEFAULT_SECRET`/`DEFAULT_CLOUD` just wire up values added in 497d6cd7; framework conventions (hooks baked into the image, CRDs with the operator, no prefix-based discovery) are all respected. Minor: the PR body is empty, and the commit headline isn't Conventional Commit style.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cardoe:
### Follow-up: use stdlib `logging` instead of `log_fn`

Design note, not a defect. This package should adopt `logging` with output to stderr, dropping the `log_fn` plumbing.

Today there are three ways to emit a line: bare `print(..., file=sys.stderr)` (pre-existing in `placeholder.py`), a [`log()` wrapper](https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/router_flavors_common.py#L97) hardcoding a `[router_flavors]` prefix, and [`log_fn` parameters](https://github.com/rackerlabs/understack/blob/fa81b55b1216f240ffbcea209695edf2a7287daf/python/openstack-sync/openstack_sync/hooks/common.py#L115-L119) defaulting to `lambda msg: print(msg, file=sys.stderr)` threaded through signatures.

`log_fn` is hand-rolled dependency injection for what `logging` already provides. `logger = logging.getLogger(__name__)` handles per-module routing so nothing gets passed down; `__name__` plus a formatter replaces the per-plugin prefix wrapper; and `caplog` covers the test capture that `log_fn` is currently carrying (`test_hook_common.py` passes `logs.append`). Levels also stop being faked as `"WARNING: "` string prefixes (`hooks/common.py:139`, `create.py:88`) and become filterable.

This is also the rest of the repo's convention — 64 files under `python/` use `getLogger(__name__)`; `openstack-sync` is the only package with no `logging` at all.

Two constraints when doing it:

- Configure via `logging.basicConfig(stream=sys.stderr, ...)` **inside each hook's `main()`**, not at import time. shell-operator captures the hook's stderr, so stderr keeps current behavior; import-time setup would compound the read-env-at-import staleness trap from #2205.
- Nothing may log to stdout — `build_hook_config` output is parsed as JSON from stdout.

🤖 Generated with [Claude Code](https://claude.ai/code)

python/openstack-sync/openstack_sync/hooks/router_flavors.py
```diff
@@ -4,113 +4,421 @@
...
+    if failed_resources:
+        LOG.error(
+            "Skipping router flavor prune because %s flavor(s) failed to reconcile",
+            len(failed_resources),
+        )
+        return 1
+
+    for credentials, credential_resources in grouped_resources.items():
+        conn = connections[credentials]
+        prune_removed_flavors(
```
cardoe: Prune computes its desired set only from this pod's `POD_NAMESPACE` snapshot but `prune_removed_flavors` operates on the shared Neutron cloud. With `PRUNE` on and two namespaces managing CRs against the same `(secretName, cloudName)`, this deletes flavors owned by the other namespace's CRs — and the authoritative-empty path at line 408 (`authoritative_empty_desired=True`) wipes *all* managed `L3_ROUTER_NAT` flavors in the cloud. We only run in the `openstack` cloud for now so it's not a live risk, but noting it: we need to come back and scope prune to marker + namespace ownership, not just the marker.

python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/router_flavors_common.py
```diff
@@ -0,0 +1,177 @@
...
+from openstack_sync.plugins.common import managed_meta_info as managed_meta_info_with
+from openstack_sync.plugins.common import meta_info_matches_without
+from openstack_sync.plugins.common import normalize_meta_info
+from openstack_sync.plugins.common import wait_for_openstack_network as wait_for_network
+
+# ---------------------------------------------------------------------------
+# Router-flavor CRD identity
+# ---------------------------------------------------------------------------
+# The chart injects these from the rendered CRD when the hook has an envPrefix.
+CRD_API_VERSION = os.environ["NEUTRON_ROUTER_FLAVOR_CRD_API_VERSION"]
```
cardoe: These module-level env reads (the `os.environ[...]` hard lookups here plus the `env_bool`/`int`/`float` parsing further down) run at import, and `hooks/router_flavors.py` imports this module at load. A missing or malformed var therefore raises *before* the hook can emit its config — including the `--config` path shell-operator needs just to register the hook. A bad value should degrade, not stop the hook from loading. Defer these reads into the config-building/run path.

python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/update.py
```diff
@@ -0,0 +1,86 @@
...
+    flavor_description_has_marker,
+)
+from openstack_sync.plugins.neutron.router_flavors.router_flavors_common import (
+    managed_flavor_description,
+)
+
+LOG = logging.getLogger(__name__)
+
+
+def ensure_flavor(conn: Any, name: str, service_type: str, description: str) -> Any:
```
cardoe: `ensure_flavor` only reconciles description drift. `find_flavor` matches on name alone, so a pre-existing flavor with a different `service_type` or `is_enabled=False` is returned unchanged, then gets stamped with the operator marker + profile and reported `Synced` (and becomes prune-eligible) while still mismatched/disabled. Flagging for a follow-up — should reconcile `service_type`/`is_enabled`, or refuse to adopt a mismatched flavor.

python/openstack-sync/openstack_sync/hooks/router_flavors.py
```diff
@@ -4,113 +4,421 @@
...
+            [resource.flavor for resource in credential_resources],
+        )
+
+    if PRUNE_REMOVED_FLAVORS:
+        deleted_only_credentials = set(deleted_resources_by_credentials) - set(
+            grouped_resources
+        )
+        for credentials in sorted(deleted_only_credentials):
+            secret_name, cloud_name = credentials
+            conn = get_openstack_connection(secret_name, cloud_name)
```
cardoe: The reconcile loop at line 332 isolates connection failures per credential group, but this deleted-only prune calls `get_openstack_connection`/`wait_for_openstack_network` bare. A transient Keystone/Neutron hiccup here raises *after* the current resources were already patched `Synced`, propagates to `main()`'s broad except → `return 1` → shell-operator re-runs the whole hook. Wrap it like the reconcile loop.

python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/delete.py
```diff
@@ -0,0 +1,253 @@
...
+            "Router flavor %s is still used by %s router(s); skipping deletion",
+            flavor_name,
+            len(routers),
+        )
+        return True
+
+    return False
+
+
+def service_profile_attached_to_any_flavor(conn: Any, profile_id: str) -> bool:
```
cardoe: This prune is O(profiles × flavors) in Neutron calls: `service_profile_attached_to_any_flavor` re-lists every L3 flavor per candidate profile, and `prune_orphaned_service_profiles` scans all profiles every reconcile. On a busy cloud the hourly sync hammers the API. This needs an immediate follow-up to this PR — build the flavor→profile map once per prune pass.

python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/delete.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def is_prunable_service_profile(profile: Any) -> bool:
+    driver = service_profile_driver(profile)
+    return bool(PRUNE_DRIVER_PREFIXES) and any(
+        driver.startswith(prefix) for prefix in PRUNE_DRIVER_PREFIXES
+    )
+
+
+def is_prunable_flavor(conn: Any, flavor: Any) -> bool:
```
cardoe: Minor: `is_prunable_flavor` takes `conn` but never uses it, so callers thread a connection through for nothing and it misleadingly implies API calls. Not worried about it — drop the param when convenient.

python/openstack-sync/openstack_sync/hooks/router_flavors.py
cardoe: Well my skill went sideways here. This is merely a comment so that we don't forget about this if we start using multiple namespaces. So it'd be good to capture this as an issue and not something to fix before merging.

python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/router_flavors_common.py
cardoe: This comment was supposed to be a question if this is a problem with the design of the shell-operator or not.

python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/update.py
cardoe: This is one to identify as an issue and not a blocker to merge.

python/openstack-sync/openstack_sync/hooks/router_flavors.py
cardoe: This seems like it should be fixed. I'm also okay as a fast follow up.

python/openstack-sync/openstack_sync/plugins/neutron/router_flavors/delete.py
cardoe: This is a follow up fix up to me.

cardoe:
I did a review pass and there are a handful of things worth tightening up — mainly around prune scoping when multiple credentials map to the same Neutron project (potential for deleting flavors owned by another credential group), the prune sweep running across all credential groups on a single-CR event, and a couple of smaller robustness items (orphaned profile cleanup when prune is disabled, driver re-verification in `find_matching_profile`, and a `KeyError` in `_split_events` that can abort a whole batch).

None of these are blocking and this PR has been open long enough. Let's get it merged and I'll track these as follow-on work.

cardoe (approved):

PR #2220, opened 2026-08-17
feat(oso): dynamically loads credentials from named secret at runtime

haseebsyed12:
## Upgrade impact

### DIFF
--- a/python/openstack-sync/openstack_sync/utils.py
+++ b/python/openstack-sync/openstack_sync/utils.py
@@ -1,0 +1,93 @@
+"""Shared utilities for all openstack-sync hooks.
+
+Provides Kubernetes secret access and OpenStack connection management so
+every hook can use the same building blocks without duplication.
+"""
+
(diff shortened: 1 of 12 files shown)


cardoe:
I'd honestly rather let the operator have the ability to read the data it needs and pass in a `secretRef` with the resource so that it can utilize the correct one.

python/openstack-sync/openstack_sync/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Returns:
+        An authenticated ``openstack.connection.Connection``.
+    """
+    cache_key = (secret_name, cloud_name)
+    if cache_key in _connection_cache:
+        return _connection_cache[cache_key]
+
+    import openstack  # type: ignore[import]
+    import yaml
```
cardoe, line 85: Any reason not to import these at the top?

cardoe (approved):

PR #2221, opened 2026-08-17
feat(undersync-switch): authenticate to Undersync with Keystone

cardoe:
The undersync-switch workflow was the last consumer of the static
undersync-token Secret. It now mounts the baremetal-manage clouds.yaml
and posts to Undersync over a keystoneauth1 session, matching what
neutron-understack already does. The unused nautobot-token mount and
NAUTOBOT_URL env are dropped along with it.


cardoe:
`undersync` needs a token from `service:service`

PR #2227, opened 2026-08-21
feat: track rack and position in Ironic extra fields

cardoe:
Refactor the code path of syncing devices from Ironic into Nautobot to utilize the `extra.rack` and `extra.position` data that would come from Ironic. To do this I first changed how we determine the Nautobot Location, which should always be equal to the OpenStack region and has not been respected before. So this updates the code to use the OpenStack region and perform a lookup of the Nautobot Location based on that value first and thread that through. There are more objects and item in Nautobot that we've treated globally unique but should be scoped to this Location but I've left that to a future update. Lastly (or first commit really) document the `extra` values that are used by scripts and code on an Ironic node.

### DIFF
--- a/python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py
+++ b/python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py
@@ -21,6 +21,7 @@ from ironicclient.common.apiclient import exceptions as ironic_exceptions
 from openstack.connection import Connection
 from pynautobot import RequestError
 from pynautobot.core.api import Api as Nautobot
+from pynautobot.core.response import Record
 
 from understack_workflows.ironic.client import IronicClient
 from understack_workflows.ironic.provision_state_mapper import ProvisionStateMapper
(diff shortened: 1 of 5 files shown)


python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py
```diff
@@ -180,6 +221,97 @@ def _generate_device_name(device_info: DeviceInfo) -> None:
...
+            position,
+            device_info.uuid,
+        )
+        return False
+
+    try:
+        # Scope by location: rack names are only unique within a location.
+        try:
+            rack = nautobot_client.dcim.racks.get(
+                name=rack_name, location_id=location.id
```
haseebsyed12: change from location_id=location.id
to location=location.id
```GET /api/dcim/racks/?name=F20-1&location_id=3dfa4bf7-…
  -> HTTP 400 Bad Request
  -> {'location_id': ['Unknown filter field']}

GET /api/dcim/racks/?name=F20-1&location=3dfa4bf7-…
  -> HTTP 200 OK  count=1
```

python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py
```diff
@@ -180,6 +221,97 @@ def _generate_device_name(device_info: DeviceInfo) -> None:
...
+                "single Nautobot rack, falling back to switch-based location",
+                rack_name,
+                location.name,
+                device_info.uuid,
+            )
+            return False
+
+        device_info.rack_id = rack.id
+        device_info.location_id = _get_record_value(rack.location, "id")
+        device_info.position = rack_position
```
nidzrai, line 308: why have we not set face = 'front' here  ?   as per https://github.com/rackerlabs/understack/pull/2227/changes#diff-aee644a2cc3ac88705de2643667f5dc5f85e317cf6e77b292878a3243403fbddR25

https://github.com/rackerlabs/understack/blob/6570a687bd344e14b308019e5a6f1b613c00f298/python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py#L600 . oh I see , do we want old face to be carried forward ?

python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py
haseebsyed12: keep whatever Nautobot already has rather than defaulting to `front`
```python
# Face is required when position is set
 target_face = device_info.face or "front"
```
I think we need to capture value from nautobot first as we did with re-enrollment flow on UUId mismatch.

_set_location_from_extra fills in rack, location, position. It never touches face, so device_info.face stays None

 fix is just filliping the lines 525 and 526 
```python
 # Face is required when position is set. Fall back to whatever the
 # device already has so a rear-mounted device isn't flipped to front.
current_face = _get_record_value(nautobot_device.face, "value")
target_face = device_info.face or current_face or "front"

```

python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py
cardoe: Yeah I wasn't messing with that. Just carrying it forward. Hopefully it's cleaner now.

python/understack-workflows/understack_workflows/oslo_event/nautobot_device_sync.py
cardoe: Hopefully this if fixed now.

haseebsyed12 (commented):
LGTM

haseebsyed12 (approved):

nidzrai (approved):

PR #2229, opened 2026-08-24
feat(openstack-sync-operator): migrate Ironic runbook shell-operator onto the generic openstack-sync-operator

haseebsyed12:
https://github.com/rackerlabs/understack/pull/2314

## Upgrade impact


.pre-commit-config.yaml
```diff
@@ -111,7 +111,7 @@ repos:
       - id: component-docs-check
         name: Component Docs Check
         description: Ensure every ArgoCD Application template has a component docs page.
-        entry: python scripts/check-component-docs.py
+        entry: python3 scripts/check-component-docs.py
```
cardoe: Was this failing?

haseebsyed12: Yes In my local. Not allowing me to commit.

cardoe:
Comparing this against the old `shell-operator-ironic` script, I think we've reintroduced a bug I already fixed once in `319dabc0`.

`patch_runbook()` and `set_traits()` both build the request path from the runbook *name*:

https://github.com/rackerlabs/understack/blob/2906664cbd77d767eb22ec8547bba246b5ad6847/python/openstack-sync/openstack_sync/plugins/ironic/runbooks/client.py#L134-L151

and `reconcile.py` calls them with `spec["runbookName"]` instead of the runbook's `uuid`, even though the runbook object it already has in hand carries one:

https://github.com/rackerlabs/understack/blob/2906664cbd77d767eb22ec8547bba246b5ad6847/python/openstack-sync/openstack_sync/plugins/ironic/runbooks/reconcile.py#L157-L187

https://github.com/rackerlabs/understack/blob/2906664cbd77d767eb22ec8547bba246b5ad6847/python/openstack-sync/openstack_sync/plugins/ironic/runbooks/reconcile.py#L195-L212

That's the exact pattern `319dabc0` moved away from - the traits PUT and the runbook PATCH both needed the UUID against a real Ironic on the same microversion (1.112) this plugin targets. `FakeBaremetal` in the tests keys everything by name, so it won't catch this - this'll break the first time a runbook gets updated or its traits change against a real cluster.

We should be passing `existing["uuid"]` / `runbook["uuid"]` into `patch_runbook` and `set_traits`, not the name.

PR #2231, opened 2026-08-24
Proposal for sync operator

geetikabatra:
## Upgrade impact

### DIFF
--- a/docs/proposals/ipsync.md
+++ b/docs/proposals/ipsync.md
@@ -1,0 +1,632 @@
+---
+authors:
+    - "@geetikabatra"
+reviewers:
+    - "@cardoe"
+    - "@mfencik"
+    - "@abhimanyu003"
+    - "@haseebsyed12"
+creation-date: 2026-07-20
+last-updated: 2026-08-17
+status: provisional
+---
+
+# Nautobot Resource Sync (ipsync)
+
+## Table of Contents
+
+- [Glossary](#glossary)
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+    - [Architecture](#architecture)
+    - [Data Ownership: Static vs Dynamic](#data-ownership-static-vs-dynamic)
+    - [User Stories](#user-stories)
+    - [Requirements](#requirements)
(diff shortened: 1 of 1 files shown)


docs/proposals/ipsync.md
```diff
@@ -0,0 +1,632 @@
...
+
+## Open Questions
+
+1. **Do tenant allocations need to be authored in the deploy repo at all?** This
+   proposal says the deploy repo authors the *ranges*, and the individual tenant
+   VLAN, VNI and subnet allocations inside those ranges are runtime values that
+   Nautobot merely records. The stricter reading — every VLAN declared in git
+   before a tenant can use it — would mean giving up tenant self-service network
+   creation, since a tenant API call cannot wait on a merge. This needs agreement
+   before VLANs are implemented, because it decides whether the event path stays.
```
cardoe, line 486: No. Tenant allocations are going to be dynamic which is why we're just providing the VLAN range, which is provided by the VLAN group.

docs/proposals/ipsync.md
```diff
@@ -0,0 +1,632 @@
...
+2. **What should happen to a hand edit made directly in Nautobot** to a static
+   type? As specified it is reverted on the next reconcile, silently. Reverting is
+   right, but it may be worth reporting rather than doing quietly, so the person
+   who made the edit finds out.
```
cardoe, line 490: Things should be reverted. Ultimately hand edits is what got us here in the first place but folks keep getting around the removal of permissions.

docs/proposals/ipsync.md
```diff
@@ -0,0 +1,632 @@
...
+3. **Which Nautobot mechanism marks ownership** — tag, custom field, or
+   relationship — and how do we backfill objects nautobotop already created?
+2. **Can the global cluster reach site Neutron/Ironic APIs?** Sites reaching
+   global is confirmed; this direction is not, and it decides whether we need a
+   per-site agent.
```
cardoe, line 495: It can but we'll have to define that mechanism for it to do so.

cardoe (commented):
So I think this would be good to bring into the ADR repo since it'll be rendered there appropriately but this is great. Thank you for working on it.

geetikabatra:
This proposal is no longer valid. So closing this MR.

PR #2238, opened 2026-08-24
fix(cinder): simplify NetApp NVMe driver to minimal override

khackworth:
Removing most of the custom driver, leaving the nqn -> initiator mapping


python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
```diff
@@ -416,8 +497,10 @@ def _volume_to_library(self, volume) -> Generator[NetAppMinimalLibrary]:
             )
 
         volume["host"] = original_host.replace(f"{svm_name}{_SVM_NAME_DELIM}", "")
-        yield lib
-        volume["host"] = original_host
+        try:
+            yield lib
+        finally:
+            volume["host"] = original_host
```
cardoe: Just wondering what this was hitting?

python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
```diff
@@ -456,10 +539,10 @@ def extend_volume(self, volume, new_size):
 
     def initialize_connection(self, volume, connector):
         """Initialize connection to volume."""
-        # TODO: the nova ironic driver sends the field 'initiator' but the NetApp
-        # cinder driver expects the field to be 'nqn' so copy the field over
+        # Translate Ironic 'initiator' field to NetApp 'nqn' field
```
cardoe: This comment isn't right. It's literally the value that nova is sending not Ironic.

python/cinder-understack/cinder_understack/dynamic_netapp_driver.py
```diff
@@ -134,6 +134,87 @@ def do_setup(self, ctxt):
         # Performance monitoring library.
         self.perf_library = perf_cmode.PerformanceCmodeLibrary(self.client)
 
+    def _map_namespace(self, name, host_nqn):
+        """Maps namespace to SVM+ostype subsystem and returns IDs.
+
+        This overrides the upstream implementation to use our custom
+        subsystem naming: {svm}-{ostype} instead of upstream's
+        random openstack-{uuid} naming.
```
cardoe: Any reason for the SVM name being mixed in here? There's a limit to namespace attachments I believe which is why a random UUID was used if there was room.

cardoe (commented):
If the purpose is to just override the NVMe subsystem of the stock driver then we should do that instead of hacking this to not update the original host. This will fail in other wise

nidzrai:
do we really need dynamic_netapp_driver  for the shared SVM design ? I think Cinder can use normal configured backends. We only need a small wrapper around the default NetApp NVMe driver unless we fix Novaor ironic to send nqn directly
```python
 if "initiator" in connector and "nqn" not in connector:
    connector["nqn"] = connector["initiator"]
```

python/cinder-understack/cinder_understack/netapp_nvme.py
```diff
@@ -0,0 +1,30 @@
...
+
+Thin wrapper around the native NetApp NVMe driver that translates
+the 'initiator' field from Nova/Ironic to the 'nqn' field expected
+by the NetApp driver.
+"""
+
+from cinder.volume.drivers.netapp.dataontap.nvme_library import NetAppNVMeStorageLibrary
+
+
+class NetAppNVMeLibrary(NetAppNVMeStorageLibrary):
```
nidzrai: we had used NetAppNVMeStorageLibrary https://github.com/openstack/cinder/blob/33b782f532d1cd199aac421ef98735ea6585971b/cinder/volume/drivers/netapp/dataontap/nvme_library.py#L62  
 Cinder loads NetAppCmodeNVMeDriver and That driver internally owns NetAppNVMeStorageLibrary. we had to inherit this coz we wanted to initialise multiple SVMS and one SVM per project.
normally it was one backend to one netapp_server but then we had to do one backend , discover many os-* SVMs n route per project. we wrote our own outer Cinder driver, NetappCinderDynamicDriver, and inside it we created multiple NetApp libraries. 

something like , 
NetappCinderDynamicDriver
  libraries["proj-a"] -> NetAppNVMeStorageLibrary
  libraries["proj-b"] -> NetAppNVMeStorageLibrary
  libraries["c"] -> NetAppNVMeStorageLibrary 

but right now we have one one NetAppNVMeStorageLibrary . we don't need NetappCinderDynamicDriver structure. and I think we should let NetAppCmodeNVMeDriver own its NetAppNVMeStorageLibrary.

khackworth: ahh, good call out, i see, thank you!

python/cinder-understack/cinder_understack/netapp_nvme.py
```diff
@@ -0,0 +1,30 @@
...
+    All other functionality is provided by the native upstream driver.
+    """
+
+    def initialize_connection(self, volume, connector):
+        """Initialize connection with connector field translation.
+
+        Nova/Ironic send 'initiator' but NetApp driver expects 'nqn'.
+        Translate if needed, then call upstream.
+        """
+        if "initiator" in connector and "nqn" not in connector:
```
nidzrai, line 27: on detach cider must tell netapp to Remove host NQN from namespace and it would look for  initiator instead of nqn.

khackworth: Unless I'm missing something, I think timing on this one may be based on previous/following commit.

nidzrai (approved):

PR #2244, opened 2026-08-25
feat(understack-workflows): Management_switch_port and Management_switch required in enroll-fw PUC-2006

nidzrai:
## What does this change do?
makes Management_switch_port and Management_switch required in enroll-fw


khackworth (approved):
looks good

PR #2298, opened 2026-09-03
chore(neutron): switch back to chart managed bin scripts

cardoe:
We had to switch to a fork of the bin scripts due to incompatible changes but now we've made the necessary fixes upstream so we can drop these overrides.

## Upgrade impact


cardoe:
@nidzrai I think we will need this for your bump to work. But I'm not sure if there's any other impact of this.

haseebsyed12 (approved):

PR #2321, opened 2026-09-09
fix(openstack-sync-operator): clear orphaned profiles when prune is off and few improvements

haseebsyed12:
## Upgrade impact

### DIFF
--- a/components/openstack-sync-operator/DESIGN-NOTES.md
+++ b/components/openstack-sync-operator/DESIGN-NOTES.md
@@ -30,10 +30,22 @@ is turned on in the operator values and its script is in the image.
 
 ### Runtime model
 
-There's no always-running loop or work queue. shell-operator watches the CRDs and
+There's no always-running loop of our own. shell-operator watches the CRDs and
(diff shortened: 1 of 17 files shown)


cardoe (changes requested):
There is still a lifecycle gap when PRUNE is false.

The orphan sweep is called from RouterFlavorPlugin.prune(), but the framework does not obtain a connection for a deletion-only run when plugin.config.prune is false. If a reconcile creates a service profile and then fails before binding it, deleting the final CR gives us one last binding context containing the credentials. However, there are no resources to reconcile, so no connection is cached, and _run_prune skips the group before RouterFlavorPlugin.prune() can call prune_orphaned_profiles(). With no CRs left, later scheduled runs also have no credentials and the orphan remains indefinitely.

I think the framework or plugin contract needs to distinguish flavor pruning from safe orphan collection so this plugin can request a connection even when PRUNE is false. Please add an integration test through run_sync() with a deletion-only input, PRUNE=false, and an orphaned managed profile; calling RouterFlavorPlugin.prune() directly does not exercise the connection gate that causes the problem.

PR #2337, opened 2026-09-14
feat(openstack-sync-operator): retain CR via finalizer until prune succeeds

haseebsyed12:
Now kubernetes finalizers make deletion two-phase: Kubernetes sets metadata.deletionTimestamp, keeps the object around, and waits until the controller removes its finalizer.

## Upgrade impact


python/openstack-sync/openstack_sync/hooks/framework.py
```diff
@@ -626,14 +707,88 @@ def run_sync(plugin: SyncPlugin, inputs: HookInputs) -> int:
...
 def _fail_group(plugin: SyncPlugin, group: list[SyncResource], message: str) -> None:
     for resource in group:
         _patch_status(plugin, resource, "Failed", message)
 
 
+def _ensure_finalizers(
+    plugin: SyncPlugin, resources: list[SyncResource]
+) -> set[tuple[str | None, str | None]]:
+    """Add finalizers to live CRs before reconciling external resources."""
+    if not plugin.uses_finalizer():
```
cardoe: When `PRUNE` is disabled, we need to remove any existing framework finalizers from live CRs. Otherwise, CRs finalized while pruning was enabled retain the finalizer indefinitely and can become stuck in `Terminating` if the hook or operator is later disabled or removed. This also makes the implementation inconsistent with the README’s statement that the finalizer is left off when `PRUNE=false`.

python/openstack-sync/openstack_sync/hooks/framework.py
```diff
@@ -626,23 +748,126 @@ def run_sync(plugin: SyncPlugin, inputs: HookInputs) -> int:
         )
         return 1
```
cardoe, line 749: When `PRUNE` is disabled, a reconciliation failure or unreadable CR returns here before `_release_deleted_finalizers()` runs. That means a deleting CR with a stale framework finalizer remains stuck in `Terminating` for as long as any unrelated CR keeps failing. When `uses_finalizer()` is false, we need to release stale deleted-resource finalizers on this early-return path too.

python/openstack-sync/openstack_sync/hooks/framework.py
```diff
@@ -626,23 +748,126 @@ def run_sync(plugin: SyncPlugin, inputs: HookInputs) -> int:
...
+    # uses_finalizer() is false there is no such cleanup to wait for: any
+    # finalizer still on a deleted CR is stale (left from when the plugin did
+    # use one), and it must be released even if a best-effort prune could not
+    # connect -- otherwise the CR is wedged in Terminating for a step it does
+    # not depend on. When uses_finalizer() is true a failed prune keeps the
+    # finalizer, because the cleanup it guards did not complete.
+    if plugin.uses_finalizer() and prune_code != 0:
+        return prune_code
+
+    release_code = _release_deleted_finalizers(plugin, inputs.deleted_resources)
```
cardoe, line 764: We should not release the finalizer solely because `plugin.prune()` returned normally. Both current prune implementations treat some incomplete deletions as success—for example, router flavors still attached to routers and Ironic runbook delete conflicts. The CR then disappears while the OpenStack resource remains. If this was the last CR for those credentials, later runs have no credentials with which to retry the cleanup. These incomplete-deletion paths need to propagate failure while the finalizer is active.

PR #2338, opened 2026-09-14
feat(ironic): panos driver base - PUC-2049

khackworth:
Adding in the scaffolding for the PAN-OS driver. 

- panos_hardware.py - Hardware type registration with standard Ironic interfaces
- panos_inspect.py - Inspection interface with 3-password credential fallback (standard → preconfig → factory)
- panos_management.py - Management interface (power, boot device operations)

Added in tests to accommodate what was intended. Secrets are set to soft fail if undercloud_deploy isn't released for that env.

### DIFF
--- a/python/ironic-understack/ironic_understack/drivers/panos_hardware.py
+++ b/python/ironic-understack/ironic_understack/drivers/panos_hardware.py
@@ -1,0 +1,94 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+"""PAN-OS Hardware Type for Palo Alto firewalls."""
+
+from ironic.drivers import generic
+from ironic.drivers.modules import noop
+from ironic.drivers.modules.network import neutron
+from ironic.drivers.modules.storage import noop as noop_storage
+
+from ironic_understack.drivers import panos_inspect
+from ironic_understack.drivers import panos_management
+
+
+class PanosHardware(generic.ManualManagementHardware):
+    """Hardware type for Palo Alto (PAN-OS) firewalls.
+
+    Intended for nodes that represent Palo Alto firewall appliances
+    (PA-1410, PA-5410, etc.). These devices:
+    - Have no deploy ramdisk
+    - Are managed via PAN-OS XML API
+    - Use Neutron for network port binding
+    - Require custom inspect and management interfaces
+
+    All other interfaces (boot, power, deploy, raid, etc.) use no-op
+    implementations since they don't apply to firewall appliances.
+
+    Boot and power are inherited from ManualManagementHardware, which
+    uses FakePower (power state controlled by manual operator) since
+    there's no automated power management for these devices.
+    """
+
+    @property
+    def supported_bios_interfaces(self):
+        """No BIOS on firewall appliances."""
+        return [noop.NoBIOS]
+
+    @property
+    def supported_console_interfaces(self):
+        """Console access not supported via Ironic for PAN-OS."""
+        return [noop.NoConsole]
+
+    @property
+    def supported_deploy_interfaces(self):
+        """No image deployment for firewall appliances."""
+        return [noop.NoDeploy]
+
+    @property
+    def supported_firmware_interfaces(self):
+        """Firmware updates handled separately."""
+        return [noop.NoFirmware]
+
+    @property
+    def supported_inspect_interfaces(self):
+        """Use custom PAN-OS inspect interface for hardware discovery."""
+        return [panos_inspect.PanosInspect]
+
+    @property
+    def supported_management_interfaces(self):
+        """Use custom PAN-OS management interface for lifecycle operations."""
+        return [panos_management.PanosManagement]
+
+    @property
+    def supported_network_interfaces(self):
+        """Use Neutron for network port binding."""
+        return [neutron.NeutronNetwork]
+
+    @property
+    def supported_raid_interfaces(self):
+        """No RAID on firewall appliances."""
+        return [noop.NoRAID]
+
+    @property
+    def supported_rescue_interfaces(self):
+        """Rescue mode not applicable to firewall appliances."""
+        return [noop.NoRescue]
+
+    @property
+    def supported_storage_interfaces(self):
+        """No storage management for firewall appliances."""
+        return [noop_storage.NoopStorage]
+
+    @property
+    def supported_vendor_interfaces(self):
+        """No vendor-specific passthrough needed."""
+        return [noop.NoVendor]
--- a/python/ironic-understack/ironic_understack/drivers/panos_inspect.py
+++ b/python/ironic-understack/ironic_understack/drivers/panos_inspect.py
@@ -1,0 +1,335 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+"""PAN-OS Inspect Interface for Palo Alto firewalls."""
+
+from typing import ClassVar
+
+from ironic.common import exception
+from ironic.common import states
+from ironic.drivers import base
+from oslo_log import log
+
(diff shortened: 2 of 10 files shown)


cardoe (commented):
As written this driver cannot load or run. Comments inline, but the two that reshape the PR are the hardware type name and where the credentials come from.

python/ironic-understack/ironic_understack/drivers/panos_hardware.py
```diff
@@ -0,0 +1,94 @@
...
+from ironic.drivers import generic
+from ironic.drivers.modules import noop
+from ironic.drivers.modules.network import neutron
+from ironic.drivers.modules.storage import noop as noop_storage
+
+from ironic_understack.drivers import panos_inspect
+from ironic_understack.drivers import panos_management
+
+
+class PanosHardware(generic.ManualManagementHardware):
```
cardoe: `bf0df36f` added a `paloalto` hardware type for this same device class, after this branch was cut, so merging this as-is leaves two non-interchangeable driver names for the same appliances. Rather than carry both, we will rename `paloalto` over to `panos` and this PR should build on the renamed type.

That also removes most of this file. `PaloAltoHardware` is a thin `NetdevHardware` subclass, and `PanosHardware` here is a near-verbatim copy of `NetdevHardware` that differs only in the inspect and management interfaces. Please subclass `NetdevHardware` and override `supported_inspect_interfaces` and `supported_management_interfaces` only; every other property in this file is already inherited.

The rename needs a note in `docs/design-guide/ironic.md`, which currently documents `netdev` only, and we should confirm whether any node already carries `driver=paloalto` before the rename lands.

python/ironic-understack/pyproject.toml
```diff
@@ -28,10 +28,15 @@ chassis_model = "ironic_understack.hooks.inspect_hook_chassis_model:InspectHookC
 
 [project.entry-points."ironic.hardware.types"]
 netdev = "ironic_understack.drivers.netdev_hardware:NetdevHardware"
+panos = "ironic_understack.drivers.panos_hardware:PanosHardware"
```
cardoe: These entry points register `panos` as a hardware type and as `inspect`/`management` interfaces, but nothing adds it to `enabled_hardware_types`, `enabled_inspect_interfaces`, or `enabled_management_interfaces` in the `values.yaml` this PR already edits. The conductor will not load the hardware type or either interface, so none of this code is reachable.

`docs/contributing/writing-ironic-drivers.md` covers this: add it to `enabled_hardware_types`, and make sure every interface it lists is present in the corresponding `enabled_<interface>_interfaces` option. `enabled_inspect_interfaces` and `enabled_management_interfaces` currently have no `panos` entry at all. With the rename above, the existing `paloalto` entry in `enabled_hardware_types` becomes `panos` rather than a second addition.

python/ironic-understack/ironic_understack/drivers/panos_inspect.py
```diff
@@ -0,0 +1,421 @@
...
+
+        LOG.info("[node:%s] Starting PAN-OS inspection for %s", node.uuid, address)
+
+        try:
+            # Get credentials from Kubernetes secret
+            # (mounted at /etc/panos-credentials/)
+            # Try in order: standard -> preconfig -> factory
+            # If credentials don't exist yet, fail gracefully
+            try:
+                from understack_workflows.helpers import credential
```
cardoe: This raises `ModuleNotFoundError` at runtime. `ironic-understack` declares only `ironic` and `pyyaml`, and `containers/ironic/Dockerfile` installs only `/src/ironic`, `/src/sushy`, `/src/understack/ironic-understack`, and `proliantutils` — `understack-workflows` is never installed into the ironic image.

The `except FileNotFoundError` below does not catch that, so the "secret may not be deployed yet" path this block exists to provide never runs. Nothing tests `inspect_hardware()`, which is why CI does not catch it either. Moving the credentials to `oslo.config` removes this import entirely; please also add a test that covers this path.

components/ironic/values.yaml
```diff
@@ -265,6 +265,9 @@ pod:
           - name: ironic-inspection-rules
             mountPath: /etc/ironic/inspection-rules/
             readOnly: true
+          - name: panos-credentials
```
cardoe: Credentials should not arrive as files mounted from a Secret. `ironic_understack/conf/__init__.py` already registers an `ironic_understack` option group via `setup_conf()`, and these belong there as `cfg.StrOpt` with `secret=True` so they stay out of logs and config dumps.

That drops this mount and volume, the `/etc/panos-credentials` path handling in `panos_inspect.validate()`, and the cross-package `credential()` call. It also closes a gap this PR leaves open: nothing in the tree creates the `panos-credentials` Secret, while every other external secret here is provisioned through an `ExternalSecret`.

python/ironic-understack/ironic_understack/drivers/panos_inspect.py
```diff
@@ -0,0 +1,421 @@
...
+    # in the traditional server sense
+    ESSENTIAL_PROPERTIES: ClassVar[set] = set()
+
+    def get_properties(self):
+        """Return the properties of the interface.
+
+        :returns: dictionary of <property name>:<property description> entries.
+        """
+        return {
+            "panos_address": "Management IP address or hostname for API access. "
```
cardoe: `get_properties()` advertises `panos_address`, `panos_username`, and `panos_password` as required `driver_info` keys, plus `panos_verify_ssl` and `panos_api_port`, but the code reads `driver_info["management_ip"]` and hardcodes `username = "admin"` and `verify_ssl = False`. `panos_api_port` is never referenced anywhere. This is what `openstack baremetal driver property list` surfaces, so it documents key names an operator cannot use.

The two interfaces of the same hardware type also disagree: `panos_management.py` intends to read `panos_address` while this one requires `management_ip`. Once the credentials move to `oslo.config`, this should collapse to config options plus whatever genuinely belongs per-node, and both interfaces need to agree on the key carrying the management address.

python/ironic-understack/ironic_understack/drivers/panos_management.py
```diff
@@ -0,0 +1,286 @@
...
+            "Optional, only used during initial setup.",
+            "panos_management_gateway": "Management gateway. "
+            "Optional, only used during initial setup.",
+        }
+
+    def validate(self, task):
+        """Validate that required credentials are present.
+
+        :param task: a TaskManager instance.
+        :raises: MissingParameterValue if required parameters are missing.
```
cardoe: This documents `:raises: MissingParameterValue` but the body is a TODO comment. Unlike the helper stubs, Ironic calls `validate()` before every task, so it silently passes no matter how the node is configured. Either implement it or drop the `:raises:` line so the contract is not misleading.

python/ironic-understack/ironic_understack/drivers/panos_hardware.py
```diff
@@ -0,0 +1,94 @@
...
+    """Hardware type for Palo Alto (PAN-OS) firewalls.
+
+    Intended for nodes that represent Palo Alto firewall appliances
+    (PA-1410, PA-5410, etc.). These devices:
+    - Have no deploy ramdisk
+    - Are managed via PAN-OS XML API
+    - Use Neutron for network port binding
+    - Require custom inspect and management interfaces
+
+    All other interfaces (boot, power, deploy, raid, etc.) use no-op
```
cardoe: This says boot, power, deploy and raid "use no-op implementations", then says two lines down that boot and power are inherited and use `FakePower`. The second statement is the correct one — Ironic has no `NoBoot` or `NoPower`, and the inherited boot interfaces are iPXE/PXE, which also contradicts the "Have no deploy ramdisk" bullet above. `netdev_hardware.py` words this accurately and is worth copying verbatim.

python/ironic-understack/ironic_understack/drivers/panos_inspect.py
```diff
@@ -0,0 +1,421 @@
...
+        # /etc/panos-credentials/, not from driver_info
+        if not driver_info.get("management_ip"):
+            raise exception.MissingParameterValue(
+                f"Node {task.node.uuid} missing required driver_info parameter: "
+                "management_ip"
+            )
+
+        # Check if credentials exist (but don't fail - just log warning)
+        # This allows understack to deploy before undercloud-deploy
+        import os
```
cardoe: `import os` belongs at the top of the file, as does the `credential` import further down. Please do not import inside function bodies.

python/ironic-understack/ironic_understack/tests/fixtures/system_info.xml
```diff
@@ -0,0 +1,28 @@
+<response status="success"><result><system><hostname>PA-1410</hostname><ip-address>10.15.149.107</ip-address><public-ip-address>unknown</public-ip-address><netmask>255.255.255.0</netmask><default-gateway>10.15.149.1</default-gateway><is-dhcp>no</is-dhcp><ipv6-address>unknown</ipv6-address><ipv6-link-local-address>fe80::8e36:7aff:fe23:3a3a/64</ipv6-link-local-address><mac-address>8c:36:7a:23:3a:3a</mac-address><time>Fri Sep 11 12:22:58 2026
```
cardoe: These look like captured values from a real unit rather than authored test data — serial `026701009879`, MACs `8c:36:7a:23:3a:3a` and `60:15:2b:61:1a:00`, management IP `10.15.149.107`, and realistic uptimes, all consistent across this fixture, `interface_management.xml`, and the sample block in `_collect_system_info()`. If this is not a lab appliance, please replace them with placeholders.

PR #2341, opened 2026-09-14
feat(neutron): Updates Neutron scaling and performance tuning settings

nicholaskuechler:
Updates Neutron scaling and performance tuning settings


cardoe (changes requested):
Some details behind this would be good. The slow or time out was related to CFS not scheduling the threads on there. I'm not saying we don't need more, we definitely will but some kind of metric or what we can analyze rather than bumping the numbers up.

The rpc_workers don't do much so I'm curious what work is needed there.

haseebsyed12 (approved):

cardoe (approved):

PR #2344, opened 2026-09-14
feat(kubectl-us): Adds new one-shot backup command for kubectl-us

nicholaskuechler:
## What does this change do?

Adjustments for the `kubectl-us-net` package. Firstly it gets renamed to simply `kubectl-us` which gives us the base command `kubectl us`.

Adds a new command `kubectl us backup local` to perform a database dump and local download of mariadb, ovn nb db, and ovn sb db.

Renames the existing commands to be under `kubectl us net <command>`.

Demo:

``` bash
 kubectl us

 Usage: kubectl-us [OPTIONS] COMMAND [ARGS]...

 Operate and troubleshoot UnderStack.

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --context            TEXT  kubectl context to use (default: current-context)                                                                                                                                 │
│ --namespace  -n      TEXT  Namespace hosting the OVN NB/SB pods [default: openstack]                                                                                                                         │
│ --nb-pod             TEXT  Northbound OVSDB pod name [default: ovn-ovsdb-nb-0]                                                                                                                               │
│ --sb-pod             TEXT  Southbound OVSDB pod name [default: ovn-ovsdb-sb-0]                                                                                                                               │
│ --os-cloud           TEXT  OpenStack cloud name (default: OS_CLOUD env / clouds.yaml default)                                                                                                                │
│ --help                     Show this message and exit.                                                                                                                                                       │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ net     Troubleshoot UnderStack's Neutron/OVN data plane.                                                                                                                                                    │
│ backup  Download UnderStack backups (MariaDB, OVN NB/SB) locally.                                                                                                                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```


python/kubectl-us/us_cli/commands/backup.py
```diff
@@ -0,0 +1,202 @@
...
+    context_slug: str,
+    ts: int,
+) -> BackupResult:
+    password = kube.get_secret_value(conn_ctx, MARIADB_SECRET, MARIADB_SECRET_KEY)
+    dest = _dest(directory, "mariadb", context_slug, "sql", ts)
+    # Pass the password via MYSQL_PWD (through `env`) instead of `-p<pw>` so
+    # it never appears in mariadb-dump's argv/process list inside the pod.
+    argv = [
+        "env",
+        f"MYSQL_PWD={password}",
```
cardoe: The root password is still exposed here. Although it is not present in `mariadb-dump`'s argv, it is embedded in the long-running local `kubectl` argv and in the remote exec request.

MariaDB Operator 25.8.4 already injects this secret into the `mariadb` container as `MARIADB_ROOT_PASSWORD`, so we should not retrieve or decode it locally. Please exec into the explicit `mariadb` container and expand it only inside the pod, for example:

```python
argv = [
    "sh",
    "-c",
    'MYSQL_PWD="$MARIADB_ROOT_PASSWORD" exec mariadb-dump '
    "-u root --single-transaction --routines --triggers --all-databases",
]
kube.exec_to_file(conn_ctx, MARIADB_POD, "mariadb", argv, str(dest))
```

This lets us remove `get_secret_value()`, the local base64 decoding, and their tests while keeping the password out of the local process arguments and exec request.

python/kubectl-us/us_cli/kube.py
```diff
@@ -36,6 +37,70 @@ def stream_exec_in_pod(
...
+    ovsdb-client/mariadb-dump byte stream, so it's never allocated.
+
+    stderr is captured and included in the raised error on failure so the
+    caller can surface why kubectl/the in-pod command failed.
+    """
+    cmd = ctx.kubectl_base() + ["exec", "-i", "-n", ctx.namespace, pod]
+    if container:
+        cmd += ["-c", container]
+    cmd += ["--", *argv]
+    with open(dest_path, "wb") as dest:
```
cardoe: If `kubectl exec` exits nonzero, this leaves a zero-length or truncated file under the final backup filename. It can then be mistaken for a completed backup. Opening with `wb` can also destroy an existing same-name backup before the replacement has succeeded.

Please stream into a uniquely named temporary file in the destination directory, atomically move it into place with `os.replace()` only after a successful exit, and unlink the temporary file on every failure path. Please also add a test that a failed command creates no final backup file and does not modify an existing destination.

python/kubectl-us/us_cli/commands/backup.py
nicholaskuechler: Updated

python/kubectl-us/us_cli/kube.py
nicholaskuechler: Updated

cardoe (approved):

PR #2349, opened 2026-09-15
feat(openstack-sync): redesign openstack sync framework

haseebsyed12:
Split the one big framework module into smaller focused ones so each part is
easier to understand, change, and test.

Also adds proper delete handling: finalizers keep a CR around long enough to
clean up its OpenStack resource before it's removed. Pruning stays safe, if we
can't be sure what should exist, nothing gets deleted.

## Upgrade impact


geetikabatra:
should we add a pause mechanism? Today the only way to stop the framework from managing a CR is to delete it, but delete can prune and destroy the real OpenStack resource, so "stop managing" and "destroy" are the same action. What happens during an incident when someone is manually fixing a resource, a debugging session, a migration/maintenance window, or a bad reconcile causing damage every run? For all of these I think we'd want a pause, a "do not touch" marker (e.g. a paused annotation, following the Cluster API convention) that skips reconcile, skips prune (treat it as still desired), keeps the CR, and shows the paused state in status. Is pause intentionally out of scope here, or a planned follow-up?

haseebsyed12:
> should we add a pause mechanism? Today the only way to stop the framework from managing a CR is to delete it, but delete can prune and destroy the real OpenStack resource, so "stop managing" and "destroy" are the same action. What happens during an incident when someone is manually fixing a resource, a debugging session, a migration/maintenance window, or a bad reconcile causing damage every run? For all of these I think we'd want a pause, a "do not touch" marker (e.g. a paused annotation, following the Cluster API convention) that skips reconcile, skips prune (treat it as still desired), keeps the CR, and shows the paused state in status. Is pause intentionally out of scope here, or a planned follow-up?



> should we add a pause mechanism? Today the only way to stop the framework from managing a CR is to delete it, but delete can prune and destroy the real OpenStack resource, so "stop managing" and "destroy" are the same action. What happens during an incident when someone is manually fixing a resource, a debugging session, a migration/maintenance window, or a bad reconcile causing damage every run? For all of these I think we'd want a pause, a "do not touch" marker (e.g. a paused annotation, following the Cluster API convention) that skips reconcile, skips prune (treat it as still desired), keeps the CR, and shows the paused state in status. Is pause intentionally out of scope here, or a planned follow-up?

If we remove marker by editing at Openstack end it will become unmanaged then we can delete CR

geetikabatra:
LGTM!

geetikabatra (approved):

PR #2350, opened 2026-09-15
fix(neutron): run the neutron RPC server so the ironic agent registers

cardoe:
### The problem

`neutron-server` runs under uWSGI, which serves the API and nothing else: the WSGI
entry point loads the paste app and returns, so no process in our deployment ever
calls `start_rpc_workers()`. `rpc_workers = 2` in our `neutron.conf` is inert, and
the RPC half of Neutron does not exist. Upstream is explicit that a WSGI deployment
needs three more processes alongside the API — `neutron-rpc-server`,
`neutron-periodic-workers` and `neutron-ovn-maintenance-worker`
([config-wsgi](https://docs.openstack.org/neutron/latest/admin/config-wsgi.html)).
We run one of the three. The comment that kept the rpc-server turned off — "this
sends requests to the various agents" — had the direction backwards:
`neutron-rpc-server` is what *receives* from agents.

Everything below lives in `Ml2Plugin.start_rpc_listeners()`, which only runs in the
rpc-server process:

- `q-plugin` — the ML2 `RpcCallbacks`: `get_device_details`, `update_device_up` /
  `update_device_down`, security group and network info.
- `q-reports-plugin` — `AgentExtRpcCallback`: agent `report_state`, plus the
  `get_agents` and `delete_agent` endpoints.
- `q-server-resource-versions` — the fanout agents use to report resource versions.
- `mechanism_manager.start_driver_rpc_listeners()` — each ML2 mechanism driver's own
  RPC listeners.
- every service plugin's RPC listeners, because `_get_rpc_workers()` hands the whole
  plugin directory to `RpcWorker` — trunk and L3 included.

### What it is causing

1. **`neutron-ironic-agent` never registers.** `report_state` is a cast, so the
   agent logs a successful report every interval while the message lands in an
   exchange with no bound queue. There is no `q-reports-plugin` queue in the
   `neutron` vhost on any cluster; dev's 25 `Baremetal Node` agents have been
   `Alive = False` since 2025-11-19, and `rax-prod-dfw3` and
   `rax-prod-iad3-rxdb-infra` are in the same state.
2. **The `baremetal` mechanism driver is permanently inert.**
   `BaremetalMechanismDriver` is a `SimpleAgentMechanismDriverBase`: `bind_port`
   resolves `context.host_agents('baremetal')` and logs "Refusing to bind port ...
   to dead agent" for anything not alive. With no live agent it never binds, so
   every baremetal port is bound by `understack` (VXLAN) or `undersync` (VLAN). We
   configure `ovn,understack,baremetal,undersync` and effectively run three of them.
3. **Stale agent rows can never be pruned.** `cleanup_stale_agents()` fetches down
   agents via `get_agents` over RPC, which is a call, not a cast. The agent only
   catches `NoSuchMethod`, so with no server it blocks until `MessagingTimeout` on
   every start.
4. **It stayed invisible because the agents we can see are not RPC agents.** The OVN
   Controller Gateway agents in `openstack network agent list` are synthesized from
   OVN SB `Chassis_Private` rows by the OVN mechanism driver, so the list never
   looked empty.
5. **Anything else that assumes an RPC path fails silently rather than loudly**,
   because a cast to an exchange with no queue always succeeds. That is the part
   worth fixing ahead of need: today the failure mode is a clean log line.

### The changes

- Enable the `neutron-rpc-server` deployment, and drop the claim that it only sends
  requests to agents, which is what kept it turned off.
- Add the missing `pod.etcSources` for `neutron_rpc_server` and
  `neutron_periodic_worker`. This is a second bug in the same area: the chart
  appends `neutron-ks-etc` itself only when `manifests.secret_ks_etc` is true, and
  we set it false, so every component has to list its `etcSources` explicitly.
  Without it both pods get an empty `/etc/neutron/neutron.conf.d` instead of the
  keystone, ironic, nova and placement service user credentials — so flipping the
  manifest flag on its own would have produced a running but broken rpc-server.
- Run two rpc-server replicas so state reports survive a rollout or a node drain.
- Pin `networking-baremetal` to `stable/2026.1` in `containers/neutron/Dockerfile`.
  The airship base image installs it from PyPI unpinned and currently lands 8.1.0 —
  the post-2026.1 series, whose CI only ever ran against `constraints/upper/master`.
  `stable/2026.1` is the only branch whose `tox.ini` pins
  `constraints/upper/2026.1`, the library set the image actually ships.
  `requirements.txt` is identical between the two.

### Behaviour change: the baremetal driver starts binding

This is not only a heartbeat fix. Once the agents report live they come with
`bridge_mappings` built from the `physical_network` of each Ironic port, which makes
`baremetal` a candidate binder for flat and VLAN segments — and it sits ahead of
`undersync` in our driver order. The guard that keeps it from taking the
hierarchical VXLAN→VLAN binding away from `undersync` is in
`try_to_bind_segment_for_agent`, which returns `False` when the first segment on the
network is `vxlan` or `geneve`. That guard, and the two fixes next to it, are why
`networking-baremetal` is pinned in this change and why it is not ready to merge:
the backports below are exactly what makes a live baremetal agent safe in our chain.

### The periodic worker

`neutron-periodic-workers` is the process that runs plugin `get_workers()`
periodics — quota cleanup, L3 garbage collection, agent health checks — and it
explicitly skips the ML2/OVN `MaintenanceWorker`. That is why 9edf31fc only needed
`deployment_ovn_maintenance_worker` for the netdev janitor: our reconcile rides
upstream's `ovn_maintenance_periodics` hook inside the maintenance worker. The
generic periodic worker stays disabled here; what this change fixes is its
credentials, so that when the router driver's reconciliation grows from an
OVN-maintenance rider into a plugin worker, enabling it is a one-line change rather
than a repeat of the `etcSources` trap above.

## Do not merge yet

Blocked on five upstream changes.

**networking-baremetal backports to `stable/2026.1`.** Once these land, bump
`NETWORKING_BAREMETAL_GIT_REF` in `containers/neutron/Dockerfile` to pick them up:

- [ ] [1005793](https://review.opendev.org/c/openstack/networking-baremetal/+/1005793) — Remove `binding:host_id` setting for L2VNI subports
- [ ] [1005794](https://review.opendev.org/c/openstack/networking-baremetal/+/1005794) — Guard rabbit config access in `_get_notification_transport_url`
- [ ] [1005795](https://review.opendev.org/c/openstack/networking-baremetal/+/1005795) — Remove invalid requested-chassis opt for localnet ports

**openstack-helm.** Once these land, bump `chartVersion` for neutron in
`charts/argocd-understack/values.yaml` from `2026.1.41+7c81bea7c` to a chart release that
contains them:

- [ ] [1005787](https://review.opendev.org/c/openstack/openstack-helm/+/1005787) — thread `ovn.ini` into the ironic agent
- [ ] [1005789](https://review.opendev.org/c/openstack/openstack-helm/+/1005789) — add a readiness probe to the ironic agent

The two openstack-helm changes are not strictly required for the rpc-server to work, but
they fix the other two defects on the same pod (OVN event-driven reconciliation disabled,
and no probe of any kind), so they should ship together.

## Verification so far

- `docker build --platform linux/amd64 --target build` succeeds; `uv` installs
  `networking-baremetal==7.2.1.dev8` from the pinned ref, replacing 8.1.0 from the base image.
- `helm template` of the OSH neutron chart against these values: all five components get
  `projected: neutron-ks-etc` at `/etc/neutron/neutron.conf.d/` (previously `emptyDir` for
  `rpc_server` and `periodic_worker`); rpc-server renders 2 replicas with init containers
  `[init, ovn-neutron-init]`.
- Not yet deployed. The acceptance test is a `q-reports-plugin` queue appearing in
  the `neutron` vhost and dev's 25 stale `Baremetal Node` agents flipping to
  `Alive = True`.

## Upgrade impact

Enabling the rpc-server is picked up by a normal resync and needs no operator action,
matching how 9edf31fc enabled the OVN maintenance worker.


mfencik:
Would be good to explain what is the current problem, what is it causing?

cardoe:
`neutron-server` runs under uWSGI, which serves the API only, so nothing starts neutron's RPC listeners and `rpc_workers` is inert. `neutron-rpc-server` is what *receives* from agents, not what sends to them — the comment saying otherwise is what kept it off.

With no consumer on `q-reports-plugin`, `neutron-ironic-agent`'s `report_state` casts are dropped silently and it never registers: dev's 25 `Baremetal Node` agents have been `Alive = False` since 2025-11-19, both prod clusters the same. That leaves the `baremetal` mechanism driver inert too, since it is agent-based and won't bind to a dead agent — four configured, three running. The same process hosts the ML2 callbacks (`q-plugin`) and service-plugin RPC (trunk, L3), and `neutron-periodic-workers` is where the router driver's periodics will run.

The `etcSources` half is a second bug: `secret_ks_etc` is false, so the chart mounts `neutron-ks-etc` only where it's listed and the rpc-server would have come up with no service credentials. The `networking-baremetal` pin carries the guard that stops the now-live baremetal driver taking the VLAN binding from `undersync`, which is why it's in the same PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

haseebsyed12 (approved):

haseebsyed12:
need to change the image pr-2350

nidzrai (approved):

components/images-openstack.yaml
```diff
@@ -41,11 +41,14 @@ images:
     neutron_l2gw: "ghcr.io/rackerlabs/understack/neutron:2026.1"
     neutron_linuxbridge_agent: "ghcr.io/rackerlabs/understack/neutron:2026.1"
     neutron_metadata: "ghcr.io/rackerlabs/understack/neutron:2026.1"
-    neutron_ovn_metadata: "ghcr.io/rackerlabs/understack/neutron:2026.1"
```
nidzrai, line 44: this change ?

nidzrai: cz I don't see any chart template reading this .

PR #2352, opened 2026-09-16
feat(openstack-sync): add NeutronSegmentRange CRD and sync plugin

geetikabatra:
Add a NeutronSegmentRange CRD to the openstack-sync-operator and a segment_ranges hook plus plugin package that reconciles Neutron network segment ranges from custom resources, built on the new SyncPlugin framework.

- CRD: neutron.understack.rackspace.net/v1alpha1 NeutronSegmentRange
- plugin: reconcile/create/adopt by owner-prefixed name, prune on removal
- immutable network_type/physical_network mismatches fail loudly
- wired into operator values.yaml (plugins.neutronSegmentRanges, disabled)
- prune driven by framework CleanupPolicy (no manual config.prune gate)

## Upgrade impact


components/openstack-sync-operator/crds/neutron.understack.rackspace.net_neutronsegmentranges.yaml
```diff
@@ -0,0 +1,214 @@
...
+                    minLength: 1
+                    maxLength: 256
+              name:
+                description: >-
+                  Neutron network segment range name. This is the identity the
+                  operator uses to find, adopt, create, and prune the range, so
+                  it must be unique per cloud.
+                type: string
+                minLength: 1
+                maxLength: 255
```
haseebsyed12: We are prepending NAME_PREFIX = "understack-sr:" which  is 14 chars, 
so the CRD’s safe spec.name max should be 241.

components/openstack-sync-operator/crds/neutron.understack.rackspace.net_neutronsegmentranges.yaml
```diff
@@ -0,0 +1,214 @@
...
+                maxLength: 32
+              physical_network:
+                description: >-
+                  The physical network the segment range is bound to. Required
+                  for ``vlan`` ranges and must be omitted for the tunnelled
+                  types (``vxlan``, ``gre``, ``geneve``); the operator enforces
+                  this at reconcile time.
+                type: string
+                minLength: 1
+                maxLength: 255
```
haseebsyed12: physical_network should be capped at 64
```
physical_network = sa.Column(sa.String(64), nullable=False,
                                 server_default='') 
```

python/openstack-sync/openstack_sync/plugins/neutron/segment_ranges/reconcile.py
```diff
@@ -0,0 +1,205 @@
...
+    checks = (
+        ("network_type", str(get_value(segment_range, "network_type", default=""))),
+        (
+            "physical_network",
+            get_value(segment_range, "physical_network", default=None),
+        ),
+    )
+    want = {
+        "network_type": spec["network_type"],
+        "physical_network": spec.get("physical_network"),
```
haseebsyed12: This makes tunnel segment ranges non-idempotent. 
For vxlan/gre/geneve, the spec can omit `physical_network`, but Neutron normalizes non-VLAN segment ranges to `physical_network=""`. 
```
is_vlan = self.network_type == constants.TYPE_VLAN
self.physical_network = kwargs['physical_network'] if is_vlan else ''
```

cardoe:
CRD schema and the name-prefix ownership marker both look right. Two blockers though.

**Needs a rebase first — main moved under it.** `9c8535a2` landed after this branch's last commit and converted `SyncPlugin.reconcile` to return a `ReconcileResult`, migrating the other three plugins with it. `SegmentRangePlugin.reconcile()` still returns the bare `list[str]` from `sync_segment_range`. `runner.py` reads `result.notes` and `result.extra_status` *outside* the `try/except` around the reconcile call, so a plain list raises an uncaught `AttributeError` on the first CR that syncs cleanly and takes the rest of the credential group down with it. Return `ReconcileResult(notes=...)`, or wrap it in the hook like `router_flavors.py` does.

CI is green because `test_segment_ranges_reconcile.py` only calls `_validate_spec` and `_immutable_drift` directly. The siblings each ship a `test_*_hook.py` that drives the plugin through `run_sync` — that's what would have caught this, and it's worth adding here.

**Prune treats an in-use range as a successful delete.** `_delete_range` catches `ConflictException` and logs "skipping delete", but `prune_removed_ranges` never records it and never raises. Same thing I flagged on #2337: we can't release the finalizer just because `prune()` returned normally. Under `FINALIZED_PRUNE` the CR disappears, the range stays in Neutron, and nothing is left to retry the cleanup. `subnet_pools/prune.py` has the identical shape but appends to `incomplete` and raises at the end — this looks like that code minus the last three lines. Test precedent in `test_prune.py::test_prune_fails_when_removed_flavor_delete_conflicts`.

Smaller:

- `_mutable_updates` sends `shared` and `project_id`, but both are `allow_put: False` in `neutron_lib/api/definitions/network_segment_range.py` — only `name`, `minimum`, `maximum` accept a PUT. A CR that flips `shared` gets a raw 400 instead of the `ConfigError` we raise for the other immutable fields. Move them into `_immutable_drift` and stop advertising them as mutable in the docstring.

- The CRD description and `markers.py` both say we adopt existing ranges, but `load_managed_ranges` only caches names already carrying the `understack-sr:` prefix. Re-adopting our own range works; one created out-of-band doesn't — we take the create branch and Neutron's unique constraint on `(default, network_type, physical_network, minimum, maximum)` rejects it. Narrow the wording or implement the lookup.

- `new_cache()`'s comment claims the listing is reused by prune. Neither `SyncPlugin.prune` nor `PruneRequest` carries the cache, and `prune_removed_ranges` re-lists from scratch.

`values.yaml`, the Dockerfile `COPY`, the 241 cap, and the tunnel-idempotency fix in `b1b883f9` all look fine.

